PR #154, opened 2023-06-30
feat: adding pytensor.config and documentation for optional dependency

AisOmar:


src/hssm/hssm.py
```diff
@@ -46,9 +47,9 @@
 
     import arviz as az
     import pandas as pd
-    import pytensor
 
 LogLikeFunc = Callable[..., ArrayLike]
+pytensor.config.floatX = "float32"
```
digicosmos86, line 52: I am slightly wary about changing a global option under the hood that affects more than HSSM, so I will defer to @AlexanderFengler for decision on this. We should at least add a warning that this option is changed just so that users know we changed this and to play well with other packages

digicosmos86 (approved):
Looks good! I would also add setting `pytensor.config.floatX = "float32"` to `index.md` and `readme.md` for the moment until the PyMC team comes up with a fix to the type mismatch issue

README.md
```diff
@@ -29,7 +29,36 @@ You can also install the bleeding edge version of `hssm` directly from this repo
...
+
+To utilize the nuts_numpyro sampler, please follow these steps:
+
+1. Install numpyro by executing the following command:
+```
+pip install numpyro
+```
+2. Import the necessary modules and configure the required settings:
+```
+import numpyro
```
AlexanderFengler, line 56: let's add something here also that points towards the jax documentation for installing gpu version of jax.

AlexanderFengler (approved):

src/hssm/hssm.py
AlexanderFengler: I am not sure if this is still relevant here (says outdated and during my review I didn't encounter this in the diff anymore. 

But in general, yeah we could at least a warning or print statement that let's users know that this setting was changed. 

We could instead have this code explicit in all of our examples, but that wouldn't look as clean, so slightly biased against it.

digicosmos86: @AlexanderFengler check out #161. Was thinking of adding a convenience function in `hssm` so float types can be changed explicitly (and also `pytensor` and `jax`) at the same time. Can be a good compromise

PR #177, opened 2023-07-06
Added a convenience function to simulate data

digicosmos86:


AlexanderFengler (approved):
looks fine. 
I have some suggestions for additional convenience, but let's just put this in for now.

AisOmar (approved):
Looks good, we should also add this to documentation in mkdocs

digicosmos86:
> Looks good, we should also add this to documentation in mkdocs

@AisOmar Yep. The documentation will be automatically updated in incorporate the new functions 😸

PR #179, opened 2023-07-06
Added a convenience function to change float types for pytensor and jax

digicosmos86:


AisOmar (approved):
thanks for doing that, I think updating this in Getting started will be useful too.

digicosmos86:
> thanks for doing that, I think updating this in Getting started will be useful too.

@AisOmar That's a great point. I have updated the README for now. Since there are multiple updates at the same time, I'll leave updating the notebooks till the end to incorporate all of them. Thanks for catching it

AlexanderFengler (approved):
Looks good.  should be reasonably transparent to users now :).

PR #180, opened 2023-07-06
Fixed sample() compatibility with `JAX` based samplers

digicosmos86:


AisOmar (approved):
looks good, just make sure it runs successful sampling. I will work on tests.

AlexanderFengler (approved):
ok

PR #185, opened 2023-07-07
137 add sampling tests

AisOmar:


AlexanderFengler (approved):
looks good. Don't see anything wrong with the tests :)

digicosmos86 (approved):
Looks good! To make the tests go faster, we can probably skip initialization and set a small number of samples (like 10)

AisOmar:
> Looks good! To make the tests go faster, we can probably skip initialization and set a small number of samples (like 10)

good point

Issue #186, opened 2023-07-08
Bug in `param.py`


digicosmos86:
The `Param` class is not parsing `bounds` specified in parameters correctly. More accurately, `HSSM` was not correctly passing the correct parameters to `Param` class.

Issue #157, opened 2023-07-05
DOCS: Header


digicosmos86:
Per Michael's suggestion, we can add a note about the hamburger menu when screen width is less than 1280px

Add a header for link to discussion, etc

PR #187, opened 2023-07-08
Fix `Param` creation and update docs for `v0.1.1` release

digicosmos86:
The PR addresses the following issues:

#186
Parameters specified in `include` with `bounds` was not parsing correctly. Fixed this problem in `hssm.py` and `param.py`.

#157 
Added some animation that points the users to the collapsed navigation bar on smaller screens.
image

Also updated the documentation.

@AlexanderFengler I moved your 006 notebook to `docs/notebooks` together with the images. However, the images are not rendered correctly. Can you take a look at `docs/notebooks/main_tutorial.ipynb`?


AlexanderFengler:
hm that's a weird one. 
it renders fine on local :/.

AlexanderFengler:
ok got it, the fix is to put the images in another folder (not sure why but mkdocs seems to do it this way)
so where the `main_tutorial.ipynb` notebook is, you add a `main_tutorial` folder, then in there put `images/...`.

mkdocs seems to automatically extend the `images/...png` into `main_tutorial/images/...png`. 

pushed this now (including some unnecessary inflation of images which we can erase again.

digicosmos86:
Pulled your commit and it works great on my end now :)

AisOmar (approved):
looks good

AlexanderFengler (approved):
ok

PR #192, opened 2023-07-10
feat: fixing small issues

AisOmar:
- fixing issues in the notebooks
- adding notebooks folder in the git ignore
- making sure notebooks rerun everytime


mkdocs.yml
```diff
@@ -23,7 +23,7 @@ nav:
 plugins:
   - search
   - mkdocs-jupyter:
-      execute: False
+      execute: True
```
digicosmos86, line 26: The getting started notebook (not in this repo but in another PR) is slightly tricky. It will need `graphviz` to render correctly which might not exist on github runners. I wonder if you could add an exclusion rule to that notebook (`docs/notebooks/getting_started.ipynb`)? I think `mkdocs-jupyter` supports that

digicosmos86 (approved):
Looks great! Thanks for fixing these issues! If you could exclude the `get_started` notebook from being rendered every time that'd be great!

mkdocs.yml
AlexanderFengler: the `graphviz` issue might come up in other notebooks too? 
If this applies also to `main_tutorial.ipnb` e.g., `graphviz` is used in there for sure.

AlexanderFengler (approved):

PR #193, opened 2023-07-11
feat: fixing small issues

AisOmar:


AlexanderFengler (approved):
looks fine, but is the `main_tutorial.ipynb` back in the notebook folder?

PR #195, opened 2023-07-12
Fixed naming inconsistencies in model graphs

digicosmos86:
This issue is now fixed:

image


AisOmar (approved):

tests/test_utils.py
```diff
@@ -233,7 +233,7 @@ def test_get_alias_dict():
 
     assert alias_regression_a["c(rt, response)"]["c(rt, response)"] == "rt,response"
     assert alias_regression_a["c(rt, response)"]["Intercept"] == "v"
```
AlexanderFengler, line 235: Is this intended as `==` 'v' ?
Looks like you were expecting `==` 'a' ?

AlexanderFengler (commented):
Just one commnt, potentially a misunderstanding.

tests/test_utils.py
digicosmos86: This is actually correct. Since `v` is the `parent`, the `Intercept` for `rt, response` should be `v`, even though `a` is the parameter that has a regression, because the main formula for the model is `c(rt, response) ~ 1`

AlexanderFengler: ok makes sense thanks

PR #213, opened 2023-07-17
Added prior predictive sampling

digicosmos86:
After using a hack mentioned [here](https://discourse.pymc.io/t/pm-sample-prior-predictive-fails-with-incorrect-size-info-passed-to-the-rng-fn-function-of-the-randomvariable/12513/5), the problem is successfully resolved.

We can now use `model.sample_prior_predictive()` to perform prior predictive sampling

### DIFF
--- a/src/hssm/distribution_utils/dist.py
+++ b/src/hssm/distribution_utils/dist.py
@@ -117,12 +117,15 @@ def make_ssm_rv(model_name: str, list_params: list[str]) -> Type[RandomVariable]
 
         # NOTE: This is wrong at the moment, but necessary
         # to get around the support checking in PyMC that would result in error
-        ndim_supp: int = 0
+        ndim_supp: int = 1
 
         ndims_params: list[int] = [0 for _ in list_params]
         dtype: str = "floatX"
         _print_name: tuple[str, str] = ("SSM", "\\operatorname{SSM}")
         _list_params = list_params
+
+        def _supp_shape_from_params(*args, **kwargs):
+            return (2,)
 
         # pylint: disable=arguments-renamed,bad-option-value,W0221
         # NOTE: `rng` now is a np.random.Generator instead of RandomState
@@ -159,6 +162,11 @@ def make_ssm_rv(model_name: str, list_params: list[str]) -> Type[RandomVariable]
             else:
                 size = args[-1]
                 args = args[:-1]
+
+            # Although we got around the ndims_supp issue, the size parameter passed
+            # here is still an array with one element. We need to take it out.
+            if not np.isscalar(size):
+                size = np.squeeze(size)
 
             arg_arrays = [np.asarray(arg) for arg in args]
 
--- a/src/hssm/distribution_utils/dist.py
+++ b/src/hssm/distribution_utils/dist.py
@@ -114,8 +114,6 @@ def make_ssm_rv(model_name: str, list_params: list[str]) -> Type[RandomVariable]
         """WFPT random variable."""
 
         name: str = "SSM_RV"
-
-        # NOTE: This is wrong at the moment, but necessary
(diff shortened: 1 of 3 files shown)


src/hssm/distribution_utils/dist.py
```diff
@@ -117,13 +117,16 @@ class SSMRandomVariable(RandomVariable):
...
         # to get around the support checking in PyMC that would result in error
-        ndim_supp: int = 0
+        ndim_supp: int = 1
 
         ndims_params: list[int] = [0 for _ in list_params]
         dtype: str = "floatX"
         _print_name: tuple[str, str] = ("SSM", "\\operatorname{SSM}")
         _list_params = list_params
 
+        def _supp_shape_from_params(*args, **kwargs):
```
AlexanderFengler, line 127: can you add a comment above this to explain (for future reference) what this does, and link to issue?

src/hssm/hssm.py
```diff
@@ -516,6 +517,37 @@ def sample_posterior_predictive(
...
+        ----------
+        draws
+            Number of draws to sample from the prior predictive distribution. Defaults
+            to 500.
+        var_names
+            A list of names of variables for which to compute the prior predictive
+            distribution. Defaults to ``None`` which means both observed and unobserved
+            RVs.
+        omit_offsets
+            Whether to omit offset terms in the plot. Defaults to ``True``.
```
AlexanderFengler, line 539: plot?

tests/test_hssm.py
```diff
@@ -217,3 +217,18 @@ def test_model_with_approx_differentiable_likelihood_type(data_angle):
     assert model.loglik == download_hf(loglik)
     trace = model.sample()
     assert isinstance(trace, az.InferenceData)
+
+
+def test_sample_prior_predictive(data):
```
AlexanderFengler, line 222: Prob wise to add a test that involves multiple regressions, and one that involves hierarchical terms. 
just to make sure nothing funny happens. 

Ok if you disagree, but if so, please add a comment in the code that explains why these tests cover abstractly speaking all cases of interest.

src/hssm/distribution_utils/dist.py
```diff
@@ -160,6 +163,11 @@ def rng_fn(
                 size = args[-1]
                 args = args[:-1]
 
+            # Although we got around the ndims_supp issue, the size parameter passed
+            # here is still an array with one element. We need to take it out.
+            if not np.isscalar(size):
+                size = np.squeeze(size)
```
AlexanderFengler, line 169: Reading this line I went down a bit and also looked at line 185 ++.
Do we report somewhere the complete logic on how we deal with sizes? 
Just for future reference, I think we should record this as comments.

AlexanderFengler (changes requested):
Looks good, left a few comments.

src/hssm/distribution_utils/dist.py
digicosmos86: Added this as Note in the docstring

src/hssm/hssm.py
digicosmos86: This is directly copied from bambi documentation lol. I don't know what this refers to

tests/test_hssm.py
digicosmos86: Added tests

src/hssm/hssm.py
AlexanderFengler: prob just take 'plot' out? 
Can add back in if it turns out to make sense :)

AlexanderFengler (approved):
Left single comment but ready for merging.

src/hssm/hssm.py
digicosmos86: Will do

PR #215, opened 2023-07-18
Added lapse probability distribution

digicosmos86:
Now `HSSM` class supports `p_outlier` and `lapse` parameters by default.


src/hssm/distribution_utils/dist.py
```diff
@@ -276,8 +332,24 @@ def dist(cls, **kwargs):  # pylint: disable=arguments-renamed
             return super().dist(dist_params, **other_kwargs)
 
         def logp(data, *dist_params):  # pylint: disable=E0213
+            if list_params[-1] == "p_outlier":
+                p_outlier = dist_params[-1]
+                dist_params = dist_params[:-1]
+
             logp = loglik(data, *dist_params)
 
+            if list_params[-1] == "p_outlier":
```
AlexanderFengler, line 341: just flagging, this should be bypassed if `p_outlier` was `None` but also if `p_outlier` was set to 0

src/hssm/distribution_utils/dist.py
```diff
@@ -276,8 +332,24 @@ def dist(cls, **kwargs):  # pylint: disable=arguments-renamed
...
+                logp = pt.log(
+                    (1.0 - p_outlier) * pt.exp(logp)
+                    + p_outlier
+                    * pt.exp(
+                        pm.logp(
+                            get_distribution_from_prior(lapse).dist(**lapse.args),
```
AlexanderFengler, line 347: no efficiency concerns about this call got get_distribution_from_prior?

is this essentially a constructor and once the graph is built, this will not be "called" as such?

src/hssm/hssm.py
```diff
@@ -230,15 +242,29 @@ def __init__(
...
+                self.list_params = default_params[model][:]
 
         self._parent = self.list_params[0]
+        if p_outlier is not None and lapse is None:
+            raise ValueError(
+                "You have specified `p_outlier`. Please also specify `lapse`."
+            )
+        if lapse is not None and p_outlier is None:
+            raise ValueError(
+                "You have specified `lapse`. Please also specify `p_outlier`."
```
AlexanderFengler, line 260: Set to warning

You have specified the `lapse` argument, to include a `lapse` distribution. The `p_outlier` is set to either 0 or `None` however. Your lapse distribution will be ignored. Was this intended?

src/hssm/hssm.py
```diff
@@ -230,15 +242,29 @@ def __init__(
...
+            )
+        if "p_outlier" in self.list_params:
```
AlexanderFengler, line 262: btw. just flagging for future versions. 

Rn is it possible to specify the `p_outlier` via a regression model and make it hierarchical?

AlexanderFengler (commented):
Minor comment.

src/hssm/distribution_utils/dist.py
digicosmos86: That is correct. You can't use `pm.Distribution` directly without having a model in the context. That's why `bambi` has its own `bmb.Prior`. This basically uses `bmb.Prior` to construct the likelihood

src/hssm/hssm.py
digicosmos86: Sounds good!

src/hssm/hssm.py
digicosmos86: Not right now but that's why I wanted this to be reviewed last. I thought of a few things to do this more consistently

digicosmos86:
@AlexanderFengler Okay I improved the way `p_outlier` is treated. Now there should be more consistency. It now _theoretically_ supports `p_outliers` with regressions. However, we still need decisions on how to print the information if it's a regression, and whether it should also be converted to hierarchical when `self.hierarchical == True`. We should fix this in a future update.

src/hssm/distribution_utils/dist.py
```diff
@@ -96,19 +101,26 @@ def make_ssm_rv(model_name: str, list_params: list[str]) -> Type[RandomVariable]
...
     """
     if model_name not in ssms_model_config:
-        logging.warning(
-            f"You supplied a model '{model_name}', which is currently not supported in "
+        _logger.warning(
+            "You supplied a model '%s', which is currently not supported in "
             + "the ssm_simulators package. An error will be thrown when sampling from "
-            + "the random variable or when using any posterior sampling methods."
+            + "the random variable or when using any "
+            + "posterior posterior sampling methods.",
```
AlexanderFengler, line 117: posterior predictive or prior predictive sampling methods

src/hssm/distribution_utils/dist.py
```diff
@@ -96,19 +101,26 @@ def make_ssm_rv(model_name: str, list_params: list[str]) -> Type[RandomVariable]
...
+            model_name,
         )
 
+    if lapse is not None and list_params[-1] != "p_outlier":
+        list_params.append("p_outlier")
+
     # pylint: disable=W0511, R0903
     class SSMRandomVariable(RandomVariable):
         """WFPT random variable."""
```
AlexanderFengler, line 126: SSM random variable

src/hssm/distribution_utils/dist.py
```diff
@@ -243,12 +267,32 @@ def rng_fn(
...
 
             if not is_all_scalar:
-                output = output.reshape((*max_shape[:-1], max_shape[-1] * n_samples, 2))
+                sims_out = sims_out.reshape(
+                    (*max_shape[:-1], max_shape[-1] * n_samples, 2)
+                )
 
-            return output
+            if p_outlier is not None:
+                assert cls._lapse is not None
```
AlexanderFengler, line 278: potentially add informative error message here?

src/hssm/distribution_utils/dist.py
```diff
@@ -243,12 +267,32 @@ def rng_fn(
...
+                out_shape = sims_out.shape[:-1]
+                replace = rng.binomial(n=1, p=p_outlier, size=out_shape).astype(bool)
+                replace = np.stack([replace, replace], axis=-1)
+                n_draws = np.prod(out_shape)
```
AlexanderFengler, line 282: np.squeeze ?

src/hssm/distribution_utils/dist.py
```diff
@@ -243,12 +267,32 @@ def rng_fn(
...
+                lapse_rt = pm.draw(
```
AlexanderFengler, line 283: did you speed test this?
I know usually pm.draw() compile the RV first, just making sure that in this case it doesn't redo that all the time.

AlexanderFengler (changes requested):
sorry few more comments, but I think this is the last round :)

src/hssm/distribution_utils/dist.py
digicosmos86: Not sure what you mean here? np.prod returns a scalar after multiplying all array elements. We need to know how many draws we need from a bernulli distribution

src/hssm/distribution_utils/dist.py
AlexanderFengler: Ok as per your answer from the previous comment, this should take all draws over all posterior (prior) predictive samples at the same time?
In this case previous point isn't relevant.

digicosmos86: Just tested with 10 runs in a loop. Seems like each time it takes about the same amount of time. This might be a function of PyMC recompiling the entire computational graph for prior predictive sampling

src/hssm/distribution_utils/dist.py
digicosmos86: This is resolved through the `HSSM` class (when `p_outlier` is set to `None` or `0`, `lapse` passed to it will be `None`)

src/hssm/distribution_utils/dist.py
AlexanderFengler: ok missed that sims_out.shape[:-1] still includes the 'draws' dimension for e.g. posterior predictive.

AlexanderFengler: so mistakenly assumed it's a scalar (in a np.ndarray) anyways.

AlexanderFengler (approved):
approved

PR #220, opened 2023-07-21
Fix analytical likelihoods to increase numerical stability

digicosmos86:
Sampling is more stable now. We no longer have `a` going crazy. Still some issues with `logp_ddm_sdv` producing some divergence, but it tends to go away by increasing `target_accept`.


AlexanderFengler:
We could actually add target accept as a parameter that has a model_specific default. 
I actually think we may be able to improve and finally fix `ddm_sdv`, but in the meanwhile we can just set target accept high for that model.

The user can still override, but prob makes sense to have that default.

src/hssm/hssm.py
```diff
@@ -203,15 +215,15 @@ def __init__(
...
+            default_config = deepcopy(default_model_config[model][loglik_kind])
 
             self.model_config = (
                 default_config
                 if model_config is None
                 else merge_dicts(default_config, model_config)
             )
             self.loglik = self.model_config["loglik"] if loglik is None else loglik
-            self.list_params = default_params[model]
+            self.list_params = default_params[model][:]
```
AlexanderFengler, line 226: what's the effect of this change?

AlexanderFengler: A priori seems equivalent

src/hssm/likelihoods/analytical.py
```diff
@@ -322,45 +258,35 @@ def logp_ddm_sdv(
...
-        - pt.log(sv**2 * rt + 1 + epsilon) / 2
-        - 2 * pt.log(a + epsilon)
-    )
-    logp = pt.where(rt <= 0, OUT_OF_BOUNDS_VAL, logp)
-    checked_logp = check_parameters(
-        logp,
-        sv >= 0,
-        msg="sv >= 0",
+
+    p = pt.maximum(ftt01w(rt, a, z_flipped, err, k_terms), pt.exp(OUT_OF_BOUNDS_VAL))
```
AlexanderFengler, line 266: let's rename out_of_bounds_val to out_of_bounds_logp

src/hssm/likelihoods/analytical.py
```diff
@@ -322,45 +258,35 @@ def logp_ddm_sdv(
...
 
-def logp_ddm(
+def logp_ddm_sdv(
     data: np.ndarray,
     v: float,
     a: float,
     z: float,
     t: float,
+    sv: float,
     err: float = 1e-7,
```
AlexanderFengler, line 290: default k_terms to 20 here
if this makes it very slow we can rethink, but it could definitely help make it more rebust.

AlexanderFengler, line 290: set err to 1e-15 default

AlexanderFengler (commented):
Left minor comments. 
Good move on simplifying the decision_func() part :)

src/hssm/config.py
```diff
@@ -41,6 +41,7 @@
             "default_priors": {
                 "v": {"name": "Uniform", "lower": -10.0, "upper": 10.0},
                 "a": {"name": "HalfNormal", "sigma": 2.0},
+                "z": None,
```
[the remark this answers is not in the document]
digicosmos86: Because `z` has bounds, a uniform prior will be automatically created. Specifying default priors here will actually truncate the default priors. This will be fixed when we support `inf` as bounds.

src/hssm/hssm.py
digicosmos86: This is actually important. We need to create a copy, because `list_params` are lists. Without creating a copy any changes to this object will affect all models pointing to the same object. It was quite a sneaky bug to uncover.

src/hssm/likelihoods/analytical.py
digicosmos86: Isn't this a lower bound? Should we make this more specific?

src/hssm/config.py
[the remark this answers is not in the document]
AlexanderFengler: ok sounds good.

src/hssm/hssm.py
AlexanderFengler: got it thanks

src/hssm/likelihoods/analytical.py
AlexanderFengler: yes just to make explicit that this is the lower bound on logp value.

src/hssm/distribution_utils/dist.py
```diff
@@ -104,11 +109,15 @@ def make_ssm_rv(model_name: str, list_params: list[str]) -> Type[RandomVariable]
     """
     if model_name not in ssms_model_config:
         logging.warning(
-            f"You supplied a model '{model_name}', which is currently not supported in "
+            "You supplied a model '%s', which is currently not supported in "
             + "the ssm_simulators package. An error will be thrown when sampling from "
-            + "the random variable or when using any posterior sampling methods."
+            + "the random variable or when using any posterior sampling methods.",
+            model_name,
```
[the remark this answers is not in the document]
digicosmos86: Yes, you can specify with the `rv` field in `model_config`

src/hssm/hssm.py
```diff
@@ -124,6 +126,14 @@ class HSSM:
             will be `analytical`. For other models supported, it will be
             `approx_differentiable`. If the model is a custom one, a ValueError
             will be raised.
+    p_outlier : optional
+        The fixed lapse probability or the prior distribution of the lapse probability.
+        Defaults to a fixed value of 0.05. When `None`, the lapse probability will not
```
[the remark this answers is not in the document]
digicosmos86: Will fix this specifically in the lapse distribution PR

AlexanderFengler (approved):
approved

PR #223, opened 2023-07-21
Add default model settings that turn every parameter hierarchical

digicosmos86:
Now the constructor of `hssm.HSSM` has a `hierarchical` parameter that defaults to `True`.  When this parameter is `True`, and `data` dataframe has column names `participant_id`, all unspecified parameter will be turned into a regression of `x ~ 1 + (1 | subject)". All regression parameters will be using default provided by `bambi`.

Right now this is not perfect: The priors used by `bambi` cannot be printed out using the model output, and the random effects of the parent parameter do not have the parameter name as prefix.


AlexanderFengler:
Could you make the reason for why they can't be printed explicit? Just for reference and understanding if this is a fundamental limitation (which I assume it is not).

I would say as a compromise, let's push this, but set default to `false`. Otherwise we would end up in a situation upon announcement where the simplest example doesn't yield the correct output, which I think we should avoid.

src/hssm/distribution_utils/dist.py
```diff
@@ -114,16 +114,22 @@ class SSMRandomVariable(RandomVariable):
         """WFPT random variable."""
 
         name: str = "SSM_RV"
-
-        # NOTE: This is wrong at the moment, but necessary
         # to get around the support checking in PyMC that would result in error
-        ndim_supp: int = 0
+        ndim_supp: int = 1
```
AlexanderFengler, line 118: can you motivate this change?

The comment above didn't change but this setting did now :). 

Also not sure why this change is related to this pr.

src/hssm/distribution_utils/dist.py
```diff
@@ -114,16 +114,22 @@ class SSMRandomVariable(RandomVariable):
...
         dtype: str = "floatX"
         _print_name: tuple[str, str] = ("SSM", "\\operatorname{SSM}")
         _list_params = list_params
 
+        # PyTensor, as of version 2.12, enforces a check to ensure that
+        # at least one parameter has the same ndims as the support.
+        # This overrides that check and ensures that the dimension checks are correct.
+        # For more information, see this issue
+        # https://github.com/lnccbrown/HSSM/issues/36
+        def _supp_shape_from_params(*args, **kwargs):
```
AlexanderFengler, line 130: I guess this is just the diff to main, and includes multiple pr's ?.
Because this change doesn't seem related to the current pr either.

src/hssm/hssm.py
```diff
@@ -164,10 +169,12 @@ def __init__(
         model_config: Config | None = None,
         loglik: str | PathLike | LogLikeFunc | pytensor.graph.Op | None = None,
         loglik_kind: LoglikKind | None = None,
+        hierarchical: bool = True,
```
AlexanderFengler, line 172: as metioned in high level comment, let's prob set this to false if print statements don't work. (However didn't understand why this breaks print statements)

tests/test_hssm.py
```diff
@@ -217,3 +218,66 @@ def test_model_with_approx_differentiable_likelihood_type(data_angle):
...
+    data = data.iloc[:10, :].copy()
+    data["participant_id"] = np.arange(10)
+
+    model = HSSM(data=data)
+    assert all(param.is_regression for param in model.params)
+
+    model = HSSM(data=data, v=bmb.Prior("Uniform", lower=-10.0, upper=10.0))
+    assert all(param.is_regression for param in model.params if param.name != "v")
+
+    model = HSSM(data=data, a=bmb.Prior("Uniform", lower=-10.0, upper=10.0))
```
AlexanderFengler, line 273: a should be positive here in terms of prior

AlexanderFengler (approved):
generally looks good, left few comments :)
I will approve for now even though ofc we should fix the print issue (not yet clear to me why this print doesn't work tbh, it implies print on the equivalent hierarchical model never worked?)

src/hssm/distribution_utils/dist.py
digicosmos86: This is a change from a previous PR that's already merged. The number of PRs were getting out of hand lol

digicosmos86:
> Could you make the reason for why they can't be printed explicit? Just for reference and understanding if this is a fundamental limitation (which I assume it is not).
> 
> I would say as a compromise, let's push this, but set default to `false`. Otherwise we would end up in a situation upon announcement where the simplest example doesn't yield the correct output, which I think we should avoid.

Right now everything that's printed comes from `Model.params`, which only stores information about parameters that are specified. For unspecified parameters using defaults provided by Bambi, it is possible to dig into the bambi `model` object for more information. I'll eventually be able to do that, but it will just take a while

src/hssm/distribution_utils/dist.py
digicosmos86: This is from a different PR for sure

digicosmos86:
Okay @AlexanderFengler I have actually fixed the problems mentioned in this PR. The group specific terms for the parent parameter are now properly aliased (it was actually a bug), and now the model looks into `bambi.Model` for correct prior outputs.

```
Hierarchical Sequential Sampling Model
Model: angle

Response variable: rt,response
Likelihood: approx_differentiable
Observations: 1000

Parameters:

v:
    Formula: v ~ 1 + (1|participant_id)
    Priors:
        v_Intercept ~ Normal(mu: 0.0, sigma: 2.5)
        v_1|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 2.5))
    Link: identity
    Explicit bounds: (-3.0, 3.0)
a:
    Formula: a ~ 1 + (1|participant_id)
    Priors:
        a_Intercept ~ Uniform(lower: 0.0, upper: 0.5)
        a_1|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 1.0))
    Link: identity
    Explicit bounds: (0.3, 3.0)
z:
    Formula: z ~ 1 + (1|participant_id)
    Priors:
        z_Intercept ~ Normal(mu: 0.0, sigma: 1.0)
        z_1|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 1.0))
    Link: identity
    Explicit bounds: (0.1, 0.9)
t:
    Formula: t ~ 1 + (1|participant_id)
    Priors:
        t_Intercept ~ Normal(mu: 0.0, sigma: 1.0)
        t_1|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 1.0))
    Link: identity
    Explicit bounds: (0.001, 2.0)
theta:
    Formula: theta ~ 1 + (1|participant_id)
    Priors:
        theta_Intercept ~ Normal(mu: 0.0, sigma: 1.0)
        theta_1|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 1.0))
    Link: identity
    Explicit bounds: (-0.1, 1.3)
```

AlexanderFengler:
ok great :). Let's merge it

PR #227, opened 2023-07-25
Fix `nuts_numpyro` sampling error with regressions

digicosmos86:


AlexanderFengler (commented):
looks good, not sure what / why tests are failing...

AlexanderFengler:
I see the failed test concern an equality test on some tensors. 
Likely the test where you check if jax / pytensor likelihoods produce the same outputs?

digicosmos86:
> I see the failed test concern an equality test on some tensors. Likely the test where you check if jax / pytensor likelihoods produce the same outputs?

This is fixed. By making all tests consistent (testing equal to 4 decimals) all tests passed.

AlexanderFengler:
So you put this down to cumulative rounding errors with float32?

AlexanderFengler:
> > So you put this down to cumulative rounding errors with float32?
> 
> I believe so. Also it was passing on my laptop, so platform differences could also be a reason

argh. Ok not too satisfying (slightly scary) but ok or now :)

AlexanderFengler (approved):
approved

PR #231, opened 2023-07-25
Updated examples to use normal instead of uniform distributions

digicosmos86:


README.md
```diff
@@ -41,45 +40,12 @@ You can also install the bleeding edge version of `hssm` directly from this repo
...
-Alternatively, you have the option to install the graphviz binaries manually and then install the Python bindings using pip with the following command:
-
-```
-pip install graphviz
-```
-
-**Dependency for sampler="nuts_numpyro"**
 
-To utilize the nuts_numpyro sampler, please follow these steps:
+You will need optional dependencies to use JAX-based samplers and graph the models.
```
AlexanderFengler, line 44: here is says optional, but I think the other pr said it's now a strict dependency?

AlexanderFengler (changes requested):

README.md
digicosmos86: Yeah at the point of this PR numpyro was still optional. If we merge this first and the next one next, all will be updated just fine ;)

AlexanderFengler (approved):
approved

PR #232, opened 2023-07-25
Better defaults for model.sample()

digicosmos86:
We now install numpyro by default and use `nuts_numpyro` sampler for `approx_differentiable` likelihoods and `jax` backend.


AlexanderFengler (approved):
Just a note here concerning installation.
Jax won't be installed as gpu version by default I think, even if you are on gpu hardware. 
But in a fresh environment it will be installed as a dependency automatically.

In this case, on gpu hardware one should first install jax gpu then the rest.

We should prob document this somewhere, especially now that numpyro is strict dependency (which in turn has jax dependency).

PR #234, opened 2023-07-25
Updated the way model configs are handled

digicosmos86:
The following are updated:

1. `inf` bounds are supported. If there is no specified default priors, here's how default priors are produced:

(-inf, upper): TruncatedNormal(mu=upper, upper=upper, sigma=2.0)
(lower, inf): TruncatedNormal(mu=lower, lower=lower, sigma=2.0)
(0, inf): HalfNormal(sigma=2.0)
(lower, upper): Uniform(lower, upper)

Specified priors will be truncated accordingly as well.

2. Added a `hssm.show_defaults()` convenient method to pretty print the defaults.
3. Updated model defaults to remove unnecessary defaults, also removed "backend" field because it can be determined programmatically. This avoids repetition and accidental mistakes.


AlexanderFengler (approved):
looks good :)

PR #235, opened 2023-07-25
added Dockerfile

zenkavi:
Not the most elegant way of doing it but so far I haven't run into any other issues with the image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,78 @@
+# Copyright (c) Jupyter Development Team.
+# Distributed under the terms of the Modified BSD License.
+
+# HSSM image found at zenkavi/hssm
+# docker pull zenkavi/hssm
+
+ARG BASE_CONTAINER=jupyter/minimal-notebook:python-3.9
+FROM $BASE_CONTAINER
+
+LABEL maintainer="Zeynep Enkavi <zenkavi@caltech.edu>"
+
+USER root
+
+# ffmpeg for matplotlib anim & dvipng for latex labels
+RUN apt-get update && \
+    apt-get upgrade -y && \
+    apt-get install -y build-essential && \
+    apt-get install -y --no-install-recommends apt-utils && \
+    apt-get install -y --no-install-recommends ffmpeg dvipng && \
+    apt install -y graphviz &&\
+    rm -rf /var/lib/apt/lists/*
+    
+USER $NB_UID
+
+## TODO: Move these to a requirements.txt and install in a virtual env instead of as root
+RUN pip install --upgrade pip && \
+    pip install --no-cache-dir "scipy==1.10.1" && \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+    apt-get install -y build-essential && \
+    apt-get install -y --no-install-recommends apt-utils && \
+    apt-get install -y --no-install-recommends ffmpeg dvipng && \
+    apt install -y graphviz &&\
+    rm -rf /var/lib/apt/lists/*
+    
+USER $NB_UID
+
+## TODO: Move these to a requirements.txt and install in a virtual env instead of as root
+RUN pip install --upgrade pip && \
```
AlexanderFengler, line 26: could you try a minimal version here?
After installing the jupyterlab related stuff, pip install hssm (in the best case alone, but potentially with very few extra installs), should suffice. 
I looks like you are manually installing a ton of dependencies that would be installed upon hssm install.

AlexanderFengler: (Admittedly I am no docker-expert, but this is what it looks like :))

AlexanderFengler (changes requested):
Thank you for the pull request. 
Just left a small note.

Dockerfile
zenkavi: Makes sense. I'm not an expert either but I cleaned it up and made a new image and it seemed to work fine. I actually found another bug so it was good to review it again.

Before I send an updated PR though I had another question: how are the tests in the repo intended to be run? I used them to quickly check if the new image was working as expected. Do you use these internally ad hoc? Are they automated somehow with each build?

The only testing framework I've used in Python is unittest (https://docs.python.org/3/library/unittest.html#organizing-test-code). I can translate your tests and run them all when building an image. So if the tests are exhaustive then this would make sure that there wouldn't be an image with a broken version of hssm. The con for this is that each script in the tests directory would change slightly. But if you accept the format and adhere to it with future tests then they can always be run automatically with the same single command (`python -m unittest`).

Alternatively tests can stay as they are and I can in effect loop through everything in the directory. So if you have another pipeline these are already seamlessly going through that would not be messed up. I don't have a preference either way.

AlexanderFengler: Thank you :). 

As per Paul's answer below, I don't think you have to worry about running tests in the docker image. As long as you install from pip that is automatically guaranteed (but in fact since we run tests on every PR, the same is true if you install from the main branch on git, even though I would recommend pip).

zenkavi:
Sorry for the delay and thanks for the comments. I think I made the changes you suggested (including updating the tutorial directories). I'm happy to do any additional changes or feel free to merge and do more on your end. I also don't mind if you prefer rejecting this PR and writing a better/more comprehensive version yourself. This was just my attempt to share what works well for me so I thought I'd share in case there are others who might find it useful.

PR #240, opened 2023-07-26
Documentation reorganization and various bug fixes

digicosmos86:
This PR updates documentation. The files were not organized and the URLs of the various documentation pages reflects that. After the updates, the documentation URLs now correctly reflects the structure of the site. Also fixed the Google Colab links that were wrong in many Jupyter notebooks. We will generate these links automatically in the future.

Also includes the following bug fixes:

1. Information about priors are not correctly passed to bambi in some cases, resulting in failures in model creation.
2. The new defaults for `jax` backends were not properly applied due to the removal of `jax` field from model configs.
3. Truncated priors are now displayed correctly with the original prior and the bounds.


AlexanderFengler (approved):
this looks good, don't see any issues.

PR #245, opened 2023-07-31
Added a `hssm.Prior` class to better handle bounds

digicosmos86:
This is the first attempt to refactor the code.

Now we have a `hssm.Prior` class as a replacement to `bmb.Prior` that internally handles truncation when bounds are provided. In the future it will also allow users to use `hssm.Prior` instead of `bmb.Prior` to reduce unnecessary imports.

### DIFF
--- a/src/hssm/param.py
+++ b/src/hssm/param.py
@@ -2,12 +2,13 @@
 
 from __future__ import annotations
 
-from typing import Any, Callable, Union, cast
+import logging
+from typing import Any, Union, cast
 
 import bambi as bmb
 import numpy as np
-import pymc as pm
-from bambi.backend.utils import get_distribution
+
+from .prior import Prior
 
 # PEP604 union operator "|" not supported by pylint
 # Fall back to old syntax
@@ -15,7 +16,8 @@ from bambi.backend.utils import get_distribution
 # Explicitly define types so they are more expressive
 # and reusable
 ParamSpec = Union[float, dict[str, Any], bmb.Prior]
-BoundsSpec = tuple[float, float]
+
+_logger = logging.getLogger("hssm")
 
 
 class Param:
@@ -61,7 +63,7 @@ class Param:
         prior: ParamSpec | dict[str, ParamSpec] | None = None,
         formula: str | None = None,
         link: str | bmb.Link | None = None,
-        bounds: BoundsSpec | None = None,
+        bounds: tuple[float, float] | None = None,
         is_parent: bool = False,
     ):
         self.name = name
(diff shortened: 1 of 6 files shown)


src/hssm/param.py
```diff
@@ -71,7 +73,7 @@ def __init__(
         self._is_truncated = False
 
         if self.bounds is not None:
-            self.bounds = cast(BoundsSpec, self.bounds)
+            self.bounds = cast(tuple[float, float], self.bounds)
             if any(not np.isscalar(bound) for bound in self.bounds):
                 raise ValueError(f"The bounds of {self.name} should both be scalar.")
```
AlexanderFengler, line 78: np.inf is scalar I assume

src/hssm/param.py
```diff
@@ -419,76 +417,43 @@ def make_bounded_prior(prior: ParamSpec, bounds: BoundsSpec) -> float | bmb.Prio
...
     Returns
     -------
-    float | bmb.Prior
-        A float if `prior` is a float, otherwise a bmb.Prior object.
+    float | Prior
+        A float if `prior` is a float, otherwise a hssm.Prior object.
     """
     lower, upper = bounds
 
     if isinstance(prior, float):
```
AlexanderFengler, line 425: might make sense to outsource all these type-checks into a separate function for readability? Something like _prior_type_checks()

AlexanderFengler: ah well looking at it, this function is not doing much beyond that anyways so makes no sense :)

src/hssm/prior.py
```diff
@@ -0,0 +1,135 @@
...
+                + "`dist` and `bounds` arguments cannot both be supplied."
+            )
+            lower, upper = self.bounds
+            if np.isinf(lower) and np.isinf(upper):
+                return
+
+            self.is_truncated = True
+            self.dist = _make_truncated_dist(self.name, lower, upper, **self.args)
+            self._args = self.args.copy()
+            self.args: dict = {}
```
AlexanderFengler, line 59: overwriting the original .args ?

src/hssm/param.py
digicosmos86: Yep. It is just another `float`

src/hssm/prior.py
AlexanderFengler: what kind of issues are we talking about?
Just asking because .args is something that is expected to be carried by a Bambi prior right?

AlexanderFengler: To make sure there aren't any unintended side-effects introduced by skipping this attribute (making it empty)

AlexanderFengler: I see ok

AlexanderFengler (approved):
approved

PR #251, opened 2023-08-03
Refactor: the Param class

digicosmos86:
The Param class was meant to be an internal record-keeping mechanism to store user-supplied parameter specifications and provide an interface between HSSM and `bambi`. After this PR, the users will have the option to directly specify parameters using `hssm.Param` class. The advantage of this is that the fields and types are more explicit and some IDEs provide hints for fields and types.

To do this, the Param class was slightly modified. The information passed to it will not be immediately processed until `.convert()` is called, so that it can still be modified later in HSSM code (such as filling with defaults and setting parameters as parents, etc).

In this process, another bug was discovered. When filling in default_priors to user-specified `Param`s, if bounds was not already provided and there are default bounds, we should not truncate the default priors. We have now added a flag `Param.do_not_truncate()` to indicate this. Unnecessarily truncating priors leads to weird errors.

### DIFF
--- a/src/hssm/param.py
+++ b/src/hssm/param.py
@@ -52,28 +52,57 @@ class Param:
         If provided, the prior will be created with boundary checks. If this
         parameter is specified as a regression, boundary checks will be skipped at
         this point.
-    is_parent
-        Determines if this parameter is a "parent" parameter. If so, the response
-        term for the formula will be "c(rt, response)". Default is False.
     """
 
     def __init__(
         self,
-        name: str,
+        name: str | None = None,
         prior: ParamSpec | dict[str, ParamSpec] | None = None,
         formula: str | None = None,
         link: str | bmb.Link | None = None,
         bounds: tuple[float, float] | None = None,
-        is_parent: bool = False,
     ):
         self.name = name
+        self.prior = prior
         self.formula = formula
-        self._parent = is_parent
-        self.bounds = tuple(float(x) for x in bounds) if bounds is not None else None
+        self.link = link
+        self.bounds = bounds
         self._is_truncated = False
+        self._is_parent = False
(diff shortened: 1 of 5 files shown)


src/hssm/param.py
```diff
@@ -52,28 +52,63 @@ class Param:
...
+        # The initializer does not do anything immediately after the object is initiated
+        # because things could still change.
+
+    def update(self, **kwargs):
+        """Update the initial information stored in the class."""
+        if self._is_converted:
+            raise ValueError("Cannot update the object. It has already been processed.")
+        for attr, value in kwargs.items():
+            if not hasattr(attr):
+                raise ValueError(f"{attr} does not exist.")
```
AlexanderFengler, line 84: 'f"{attr} is not a valid attribute of a Param' ?

src/hssm/param.py
```diff
@@ -52,28 +52,63 @@ class Param:
...
+        self._is_parent = True
+
+    def do_not_truncate(self):
+        """Flag that prior should not be truncated.
+
+        This is most likely because both default prior and default bounds are supplied.
+        """
+        self._do_not_truncate = True
+
+    def convert(self):
```
AlexanderFengler, line 98: can you choose a slightly more expressive name here? 
convert to what?

src/hssm/param.py
```diff
@@ -52,28 +52,63 @@ class Param:
...
+        """Process the information passed to the class."""
```
AlexanderFengler, line 99: Same here concerning this doc string.

AlexanderFengler (approved):
Left some minor comments, please read before merging. 
In general though it is ok to merge, the comments refer to nice-to-haves.

PR #258, opened 2023-08-04
238 refactor organize tests

digicosmos86:
Mainly just cleaned up the tests a bit. Using one file for dataset simulation across the board.


tests/test_distribution_utils.py
```diff
@@ -68,3 +69,86 @@ def test_lapse_distribution():
     random_sample_b = rv.rng_fn(rng2, *[0.5, 0.5, 0.5, 0.3], 0.05, 10)
 
     np.testing.assert_array_equal(random_sample_a, random_sample_b)
+
+
+def test_apply_param_bouds_to_loglik():
```
AlexanderFengler, line 74: test_apply_param_bouds_to_loglik --> test_apply_param_bounds_to_loglik ?

tests/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+import numpy as np
+import pandas as pd
+import ssms
+
+import hssm
+
+
+@pytest.fixture(scope="module")
+def data_ddm():
+    v_true, a_true, z_true, t_true = [0.5, 1.5, 0.5, 0.5]
```
AlexanderFengler, line 12: Just a note: Might be good to test a range of parameter sets here for robustness in some cases. Those parameters are hard-coded here, and we might miss some strange things with the likelihoods if we fixate only on this (and the one below) parameter set.

AlexanderFengler (approved):
This looks fine to me.

tests/test_distribution_utils.py
digicosmos86: Nice catch!

digicosmos86:
```suggestion
def test_apply_param_bounds_to_loglik():
```

PR #267, opened 2023-08-23
Added blackbox likelihoods for `ddm` and `ddm_sdv` models

digicosmos86:
This PR addresses two issues:

1. Have blackbox likelihood defaults for `ddm` and `ddm_sdv` models
2. Added `full_ddm` blackbox likelihood function support
3. Modify default parent behavior. `parent` used to be the first parameter. Now parent is the first regression parameter. If there is no regression, it is the first parameter.

### DIFF
--- a/src/hssm/likelihoods/blackbox.py
+++ b/src/hssm/likelihoods/blackbox.py
@@ -6,16 +6,32 @@ import hddm_wfpt
 import numpy as np
 
 
+def hddm_to_hssm(func):
+    """Make HDDM likelihood function compatible with HSSM."""
+
+    def outer(data: np.ndarray, *args, **kwargs):
+        x = data[:, 0] * np.where(data[:, 1] == 1, 1.0, -1.0).astype(np.float64)
+        size = len(data)
+
+        args_list = [_broadcast(param, size) for param in args]
+        kwargs = {k: _broadcast(param, size) for k, param in kwargs.items()}
+
+        logp = func(x, *args_list, *kwargs)
+        logp = np.where(np.isfinite(logp), logp, -66.1)
+
+        return logp.astype(data.dtype)
+
+    return outer
+
+
+@hddm_to_hssm
 def logp_ddm_bbox(data: np.ndarray, v, a, z, t) -> np.ndarray:
     """Compute blackbox log-likelihoods for ddm models."""
-    x = (data[:, 0] * data[:, 1]).astype(np.float64)
     size = len(data)
-
-    v, a, z, t = [_broadcast(param, size) for param in [v, a, z, t]]
     zeros = np.zeros(size, dtype=np.float64)
 
-    logp = hddm_wfpt.wfpt.wiener_logp_array(
-        x=x,
+    return hddm_wfpt.wfpt.wiener_logp_array(
+        x=data,
         v=v,
         sv=zeros,
         a=a * 2,  # Ensure compatibility with HSSM.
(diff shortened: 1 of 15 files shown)


tests/test_mcmc.py
```diff
@@ -26,14 +32,38 @@ def test_reg_models(data_ddm_reg):
...
+    model1 = hssm.HSSM(data_ddm_reg, a=param_reg)
+    model1.sample(cores=1, chains=1, tune=10, draws=10)
+    model1.sample(sampler="nuts_numpyro", cores=1, chains=1, tune=10, draws=10)
+
+    model2 = hssm.HSSM(data_ddm_reg, loglik_kind="approx_differentiable", a=param_reg)
+    model2.sample(cores=1, chains=1, tune=10, draws=10)
+    model2.sample(sampler="mcmc", cores=1, chains=1, tune=10, draws=10)
 
-    model3 = hssm.HSSM(data_ddm_reg, a=param_reg)
+    model3 = hssm.HSSM(data_ddm_reg, loglik_kind="blackbox", a=param_reg)
```
digicosmos86, line 65: @AlexanderFengler

src/hssm/likelihoods/blackbox.py
```diff
@@ -0,0 +1,87 @@
...
+
+from __future__ import annotations
+
+import hddm_wfpt
+import numpy as np
+
+
+def logp_ddm_bbox(data: np.ndarray, v, a, z, t) -> np.ndarray:
+    """Compute blackbox log-likelihoods for ddm models."""
+    x = (data[:, 0] * data[:, 1]).astype(np.float64)
```
AlexanderFengler, line 11: This step assume -1, 1 coding for choices. (negative rts correspond to -1 choices)
For 0, 1 coding this will break likelihood evaluation.
Do we enforce that actually?

src/hssm/likelihoods/blackbox.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+    logp = np.where(np.isfinite(logp), logp, -66.1)
+
+    return logp.astype(data.dtype)
+
+
+def logp_ddm_sdv_bbox(data: np.ndarray, v, a, z, t, sv) -> np.ndarray:
+    """Compute blackbox log-likelihoods for ddm_sdv models."""
+    x = (data[:, 0] * data[:, 1]).astype(np.float64)
```
AlexanderFengler, line 36: same as above

src/hssm/likelihoods/blackbox.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+    logp = np.where(np.isfinite(logp), logp, -66.1)
+
+    return logp.astype(data.dtype)
+
+
+def logp_full_ddm(data: np.ndarray, v, a, z, t, sv, sz, st):
+    """Compute blackbox log-likelihoods for full_ddm models."""
+    x = (data[:, 0] * data[:, 1]).astype(np.float64)
```
AlexanderFengler, line 61: same

tests/test_mcmc.py
AlexanderFengler: this is resolved now right?

AlexanderFengler (commented):
few small comments.

src/hssm/likelihoods/blackbox.py
digicosmos86: Maybe we want to enforce that? What was the assumption in HDDM?

AlexanderFengler: It's more that the LAN likelihoods specifically do not enforce this and operate under the assumption that decisions are [0, 1] (if I am not gravely misremembering). 

On the LAN side this is important because that's what the networks are trained on.

For consistency, I think we should stick with `[0, 1]` and just apply the change to `[-1, 1]` inside the blackbox likelihood wrapper for the c-functions. 

We can also allow both on the user-facing side, check for it and apply necessary changes before likelihood construction, but it's probably better to give users strict / simple advice on what to do (supply choices as [0, 1] for 2-choice models) and then stick with it.

AlexanderFengler: To answer the question directly: In `hddm` the guidance was [-1, 1] :).

digicosmos86: Okay I changed this behavior. Now it can deal with both assumptions :)

AlexanderFengler (approved):
ok looks good now!

PR #269, opened 2023-08-29
Fixed a bug in sample_posterior_predictive()

digicosmos86:
The bug originates from the handling of `size` parameter passed to `rng_fn`

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -34,7 +34,7 @@ def data_angle():
 def data_ddm_reg():
     # Generate some fake simulation data
     intercept = 0.3
-    x = np.random.uniform(0.5, 0.2, size=1000)
+    x = np.random.uniform(0.5, 0.7, size=1000)
     y = np.random.uniform(0.4, 0.1, size=1000)
 
     v = intercept + 0.8 * x + 0.3 * y
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -33,9 +33,9 @@ def data_angle():
 @pytest.fixture(scope="module")
 def data_ddm_reg():
     # Generate some fake simulation data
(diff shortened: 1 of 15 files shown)


tests/conftest.py
```diff
@@ -34,7 +34,7 @@ def data_angle():
 def data_ddm_reg():
     # Generate some fake simulation data
     intercept = 0.3
-    x = np.random.uniform(0.5, 0.2, size=1000)
+    x = np.random.uniform(0.5, 0.7, size=1000)
     y = np.random.uniform(0.4, 0.1, size=1000)
```
AlexanderFengler, line 38: intercept = 1.5
x, y = np.random.uniform(-.5, .5, size = 1000)

tests/test_utils.py
```diff
@@ -64,11 +64,13 @@ def test_get_alias_dict():
 
     assert alias_regression["c(rt, response)"] == "rt,response"
     assert alias_regression["Intercept"] == "v_Intercept"
+    assert alias_regression["x"] == "v_x"
```
AlexanderFengler, line 67: is this test_util still ok after demoting `v` from default parent?

AlexanderFengler (commented):
Looks good just minimal comment!

tests/test_utils.py
digicosmos86: Yep. In this case the parent is still `"v"`. With `alias_regression_a` the parent becomes `"a"` which is why I had to change the tests

tests/conftest.py
digicosmos86: fixed

AlexanderFengler (approved):
ok looks good.

PR #270, opened 2023-08-29
Support for arbitrary additional fields

digicosmos86:
We now support additional data fields passed to the likelihood function. This is achieved through manipulating `dist()` method in SSM_Distribution so that `extra_fields`, which is now saved as a class attribute, is also passed to `logp()` method as part of `dist_params`. This way, any `RandomVariable` class will also receive these extra fields, but the classes themselves do not keep an extra reference to this object. This makes it easier to keep these extra fields consistent.

Please note that `extra_fields` is now a part of `model_config`. It is now possible to set `extra_fields` as a default for a likelihood function, as long as the likelihood function accepts these extra_fields as arguments.

### DIFF
--- a/tests/test_likelihoods.py
+++ b/tests/test_likelihoods.py
@@ -10,7 +10,8 @@ import pytest
 from numpy.random import rand
 
 # pylint: disable=C0413
-from hssm.likelihoods.analytical import compare_k, logp_ddm_sdv
+from hssm.likelihoods.analytical import compare_k, logp_ddm, logp_ddm_sdv
+from hssm.likelihoods.blackbox import logp_ddm_bbox, logp_ddm_sdv_bbox
 
 
 def test_kterm(data_ddm):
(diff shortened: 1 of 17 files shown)


tests/test_likelihoods.py
```diff
@@ -97,3 +98,21 @@ def test_no_inf_values_v(data_ddm, shared_params):
         assert np.all(
             np.isfinite(logp.eval())
         ), f"log_pdf_sv() returned non-finite values for v = {v}."
+
+
+def test_bbox(data_ddm):
```
AlexanderFengler, line 103: for these tests, it might make sense to take a sample of a bunch of parameter vectors instead of one fixed such vector.

AlexanderFengler (commented):
Looks good don't see anything wrong per se. 
Just some small comments.

tests/test_likelihoods.py
digicosmos86: Let's fix these tests in a different PR. I feel like there is a lot of improvements that we can do to these tests to make them much faster

AlexanderFengler (approved):
tests for another pr. this looks good then.

PR #275, opened 2023-09-06
Set hddm-wfpt as required, separate dependency

digicosmos86:


AlexanderFengler (approved):
ok this looks good :)

PR #283, opened 2023-09-21
Bumped PyMC version and limited numpy version

digicosmos86:
PyMC fixed pytensor import error in 5.8.2. Using that version onwards.

Also numpy 1.26 is causing some compatibility issues with pytensor. Pytensor is limiting the version of numpy and we are doing the same until they resolve this.

### DIFF
--- a/.github/workflows/build_and_publish.yml
+++ b/.github/workflows/build_and_publish.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python-version: ["3.9", "3.10", "3.11"]
+        python-version: ["3.10", "3.11"]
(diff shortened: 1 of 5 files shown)


AlexanderFengler:
This still needs some work?

AlexanderFengler (approved):
Did you take our 3.12 in some places here?
I thought I reviewed and approved this (including 3.12 here) earlier today.

digicosmos86:
> Did you take our 3.12 in some places here? I thought I reviewed and approved this (including 3.12 here) earlier today.

That was for `hddm_wfpt`. I am cautious about adding 3.12 immediately because some dependencies might not have support for it yet.

PR #286, opened 2023-09-25
Fix bug in extra fields

digicosmos86:
The bug is caused by not updating model_config in the update method of `hssm.ModelConfig`


AlexanderFengler (approved):
looks good. 
so this was basically about lack of deepcopy?

digicosmos86:
No it was actually sillier than that. I just forgot to update the `ModelConfig.update()` method to take into account a user-specified `extra_field` field 🤪

PR #303, opened 2023-10-20
Posterior predictive plotting

digicosmos86:
This PR adds the ability to perform posterior predictive plotting to HSSM. Many other improvements are also included in this PR. The main changes are:

- [ ] #290 Posterior predictive plots (both from `hssm.plotting.plot_posterior_predictve()` and `model.plot_posterior_predictive()`

The API design allows for maximum compatibility with `matplotlib` and `seaborn`. The plotting functions are designed to be thin wrappers around `seaborn` `histplot` and `FacetGrid`

- [] Added the ability to sample posterior predictives from a subset of posterior samples
- [] Functions to wrangle arviz InferenceData objects
- [] Tests for plotting functions and the above helper functions
- [] Restructured tests a bit and removed unnecessary files
- [] Documentation for plotting functions


src/hssm/plotting/posterior_predictive.py
```diff
@@ -0,0 +1,199 @@
...
+        plotting_df = _get_plotting_df(
+            idata, data, extra_dims=extra_dims, n_samples=None
+        )
+    else:
+        plotting_df = _get_plotting_df(
+            idata, data, extra_dims=extra_dims, n_samples=n_samples
+        )
+
+    # Flip the rt values if necessary
+    if np.any(plotting_df["rt"] == 0):
```
AlexanderFengler, line 131: this should happen only if `n_choices == 2`. 
if you apply this on other models you turn all non-zero responses to `1` ?

src/hssm/plotting/posterior_predictive.py
```diff
@@ -0,0 +1,199 @@
...
+    Parameters
+    ----------
+    model : hssm.HSSM
+        A fitted HSSM model.
+    idata : optional
+        The InferenceData object with posterior samples. If not provided, will use the
+        traces object stored inside the model. If posterior predictive samples are not
+        present in this object, will generate posterior predictive samples using the
+        this InferenceData object and the original data.
+    data : optional
```
AlexanderFengler, line 45: Data here has two functions:

1. Specify the covariates to allow posterior predictive sampling (potentially out-of-sample --> pure predictions) (note that we can literally make the values of the covariates up, and still get posterior predictive samples from our model)
2. Supply the real `rts, responses`. This is used only when we want to add real data to our plot.

Note that the first aspect can be used to generate posterior predictive samples WITHOUT having any real `rts, responses`.

The second aspect should be optional in this plot. 
We should be able to get the posterior predictive plot with and without overlay of real `rts, responses`.

src/hssm/plotting/posterior_predictive.py
```diff
@@ -0,0 +1,199 @@
...
+        [`sns.histplot` function]
+        (https://seaborn.pydata.org/generated/seaborn.histplot.html) for how to use
+        them. in combination to control the bins. This argument overrides bins but can
+        be used with `binrange.`
+    grid_kwargs : optional
+        Additional keyword arguments are passed to the [`sns.FacetGrid` constructor]
+        (https://seaborn.pydata.org/generated/seaborn.FacetGrid.html#seaborn.FacetGrid.__init__)
+        when any of row or col is provided. When producing a single plot, these
+        arguments are ignored.
+    kwargs : optional
```
AlexanderFengler, line 88: histplot also allows setting to `kde` right?
will this override (or correspondingly ignore) the setting of binwidth?

src/hssm/plotting/utils.py
```diff
@@ -0,0 +1,105 @@
...
+        .droplevel(["chain", "draw"], axis=0)
+    )
+
+    # Rename the columns
+    stacked.columns = ["rt", "response"]
+
+    return stacked
+
+
+def _get_plotting_df(
```
AlexanderFengler, line 56: Note that the definition of this function intersect with a comment I make above in the context of the posterior_predictive plot. 

The addition or "real" data, should be optional.

src/hssm/datasets.py
```diff
@@ -19,7 +19,12 @@
         filename="cavanagh_theta",
         path=os.path.join(base_dir, "hssm/datasets/cavanagh_theta_nn.csv"),
         description="Description for cavanagh_theta dataset",
-    )
+    ),
+    "cavanagh_theta_old": FileMetadata(
```
AlexanderFengler, line 23: why do we actually need the old version?

AlexanderFengler (changes requested):
Wow, big pr for adding one (plus a few wrapper) plot(s) :).
Hopefully the infrastructure will prove useful of more plotting stuff.

Left a whole bunch of comments, mostly minor but some things concerning functionality and logic.

Might make sense to meet to clarify.

src/hssm/datasets.py
digicosmos86: For backward compatibility. Some people might have already written some code with the old dataset. This way they can adapt their code to the new version without much trouble. We can remove this in the future.

src/hssm/hssm.py
```diff
@@ -514,6 +573,103 @@ def graph(self, formatting="plain", name=None, figsize=None, dpi=300, fmt="png")
...
+        [arviz documentation]
+        (https://arviz-devs.github.io/arviz/api/generated/arviz.plot_trace.html)
+        for additional parameters that can be specified.
+
+        Parameters
+        ----------
+        data : optional
+            An ArviZ InferenceData object. If None, the traces stored in the model will
+            be used.
+        include_computed_values : optional
```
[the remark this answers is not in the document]
digicosmos86: I actually used `include_deterministic` but then with the `rt,response_mean` thing I thought `computed_values` might be more inclusive here. The official term for this is `posterior pushforward quantities` from [Arviz documentation](https://python.arviz.org/en/stable/getting_started/WorkingWithInferenceData.html). What do you think?

src/hssm/plotting/posterior_predictive.py
```diff
@@ -0,0 +1,199 @@
...
+
+_logger = logging.getLogger("hssm")
+
+
+def plot_posterior_predictive(
+    model,
+    idata: az.InferenceData | None = None,
+    data: pd.DataFrame | None = None,
+    n_samples: int | float | None = 20,
+    row: str | None = None,
```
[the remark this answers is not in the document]
digicosmos86: Yep. I am not sure about the usefulness of another dimension tbh

src/hssm/plotting/posterior_predictive.py
digicosmos86: Right now we are not allowing `kde`, but I can add this option. Then it will work however seaborn decides

src/hssm/plotting/posterior_predictive.py
digicosmos86: Right now I am assuming `n_choices == 2` since we are not supporting more than 2 choices yet. We can revisit this once we do support more than 2 responses

src/hssm/plotting/utils.py
digicosmos86: Cool. Will work on this

src/hssm/plotting/posterior_predictive.py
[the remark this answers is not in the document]
AlexanderFengler: Consider the case where a dataset actually has three categorical variables that specify groupings. 

In this case we could e.g. 

1. Take one of them, and and make a plot for each of it's values --> subplots across the other two

This would generalize to 4,5 etc. groupings

2. Warn users that they should subset their data according to the leading groups, and only pass data that has (up to) two relevant group dimensions left.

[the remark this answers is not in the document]
AlexanderFengler: As another way to think about it:

We could organize the computation by sub graph. 

1. Kernel:
You subset the dataframe, run the corresponding posterior predictive, use it to fill in one axis object. 

2. Scaffolding code:
Organizes the returned axes objects into a grid (potentially splitting them across plots etc.)

This way of thinking also allows (because prob more efficient from a simulation overhead viewpoint):
1. Run big posterior predictive simulation up front
2. subset posterior predictives and pass them to kernels
3. build larger plot

I acknowledge that this way of doing things might make seaborn a little less appealing, however retains flexibility on our end. 

What I mean is that we can think of `plots` as kernels, and retain and reuse the `scaffolding wrapper`.

So for new plots we just think about building plotting kernels.

src/hssm/plotting/posterior_predictive.py
AlexanderFengler: yeah that would make sense, because some people I think are biased towards liking smooth plots :D

src/hssm/plotting/posterior_predictive.py
AlexanderFengler: ok for now then

src/hssm/plotting/posterior_predictive.py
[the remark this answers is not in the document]
digicosmos86: I totally see how it is possible to implement this. I am just not sure if this is a good visualization practice, once we have these high-dimensional visualizations. There is already a lot of information in the two dimensional case. So I guess it's more about me not wanting to encourage practice like this lol. However, if you have seen enough of a demand for higher dimensional charts, I am happy to implement this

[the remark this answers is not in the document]
AlexanderFengler: Take it on faith for now that people might want that :D.
I will motivate this a bit more during our next meeting.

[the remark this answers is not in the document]
digicosmos86: Then maybe let's allow for another dimension called `group`? Anything more than that, I feel like we should just make it possible for the users to use for loops themselves with the 2D version of the plotting function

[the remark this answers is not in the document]
AlexanderFengler: To make it general, just allow for group to be a list of covariates?
Then turn the combination of values into a top-level title for the respective 2D plots?

And in principle users can also go and loop over 2D plots themselves by passing the correct data-subsets as of now right?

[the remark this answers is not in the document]
digicosmos86: Okay @AlexanderFengler this is now added. The `groups` and `groups_order` arguments now supports arbitrary grouping dimensions.

> And in principle users can also go and loop over 2D plots themselves by passing the correct data-subsets as of now right?

Yep. This should be used in conjunction with `_get_plotting_df`, but now that the main function supports arbitrary grouping, this should not be necessary (though still possible)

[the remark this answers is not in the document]
AlexanderFengler: Great!

AlexanderFengler (approved):

PR #319, opened 2023-11-13
First attempt at Quantile probability plot

digicosmos86:
I have done the following in this PR:

- [ ] Updating posterior probability plots. The plotting backend went from `seaborn` to basic `matplotlib` plots with the ability to plot regions of uncertainty
- [ ] Added quantile probability plots. Tests will fail for the plots for now but I will update the fixtures soon to fix the tests
- [ ] Fixed a bug in posterior predictive sampling


src/hssm/distribution_utils/dist.py
```diff
@@ -279,19 +279,24 @@ def rng_fn(
...
+                replace_mask = np.stack([replace, replace], axis=-1)
+                n_draws = np.prod(replace_shape)
                 lapse_rt = pm.draw(
                     get_distribution_from_prior(cls._lapse).dist(**cls._lapse.args),
                     n_draws,
                     random_seed=rng,
-                ).reshape(out_shape)
-                lapse_response = rng.binomial(n=1, p=0.5, size=out_shape)
+                ).reshape(replace_shape)
+                lapse_response = rng.binomial(n=1, p=0.5, size=replace_shape)
```
AlexanderFengler, line 293: here I would go with np.random.choice(n_choices, ...) to accomodate models that have more than 2 choices.

if n_choices == 2, we change 0 --> -1, otherwise keep choices as they are.

src/hssm/plotting/quantile_probability.py
```diff
@@ -0,0 +1,427 @@
...
+    _get_title,
+    _process_df_for_qp_plot,
+    _subset_df,
+    _use_traces_or_sample,
+)
+
+_logger = logging.getLogger("hssm")
+
+
+def _plot_quantile_probability_1D(
```
AlexanderFengler, line 26: some of these plotting parameters might be passed through to the plot as **kwargs?
(I know I didn't do this well in HDDM :))

AlexanderFengler: reading a bit more I think you already tried to account for that quite a bit, but took the choice e.g. to keep xlabel and ylabel as named arguments here?

src/hssm/plotting/utils.py
```diff
@@ -79,9 +109,19 @@ def _get_plotting_df(
...
+    extra_dims = [] if extra_dims is None else extra_dims
+
+    if idata is None:
+        data = _process_data(data, extra_dims)
+
+        data.insert(0, "observed", "observed")
+        return data
 
     # get the posterior samples
+    idata_posterior = idata["posterior_predictive"]["rt,response"]
```
AlexanderFengler, line 124: seems like logic wise, you could have `idata is None --> True` but then still access `idata` in this line?

src/hssm/plotting/utils.py
```diff
@@ -200,3 +239,154 @@ def _get_title(cols: Iterable["str"], col_values: Iterable[Any]) -> str:
...
+        The column for whether the answer is correct.
+
+    Returns
+    -------
+    pd.DataFrame
+        The dataframe with `quantile`, `rt`, and `cond` variables.
+    """
+    if isinstance(q, int):
+        q = np.linspace(0, 1, q)[1:-1]
+
```
AlexanderFengler, line 269: consider adding a check here that checks if min(q) >= 0 and max(q) <= 1.
potentially also check if len(q) < 10 or if int q < 10 --> warning that this is a strange choice and that usually one would pick 4-5 quantiles.

src/hssm/plotting/utils.py
```diff
@@ -200,3 +239,154 @@ def _get_title(cols: Iterable["str"], col_values: Iterable[Any]) -> str:
...
+        model.sample_posterior_predictive(
+            idata=idata,
+            data=data,
+            inplace=True,
+            n_samples=n_samples,
+        )
+        idata = model.traces
+        sampled = True
+
+    return cast(az.InferenceData, idata), sampled
```
AlexanderFengler, line 377: is this not coming back as inference data?
could find it immediately, but the `sampled` return here is just for testing or is that used somewhere in the calling function?

AlexanderFengler (commented):
Looks great. 
Left a few comments that should be very simple to accomodate. 
Otherwise, let's do final debugging simply by battle-testing :)

src/hssm/distribution_utils/dist.py
digicosmos86: Can we tab this for now since it might take a bit more engineering to figure out where that `n_choices` comes from. `ssms` would need that to figure out how many responses to generate right?

src/hssm/plotting/utils.py
digicosmos86: The function returns if `idata is None` at line 121, so it would not reach here if `idata is None`?

src/hssm/plotting/utils.py
digicosmos86: Sounds good!

src/hssm/plotting/utils.py
digicosmos86: It is coming out as `InferenceData`, but since `idata` is defined as `InferenceData | None` in function signature, `mypy` thinks the return type here is also `InferenceData | None` which isn't consistent the type annotation, so we need to explicitly tell it that it cannot be `None`

And yes, `sampled` here just indicates whether to skip certain steps downstream

src/hssm/distribution_utils/dist.py
AlexanderFengler: yeah. That should be part of model configs actually so should be easily accessible. (can double check but I think it is for ssm-simulators?)

AlexanderFengler: Also outputs from ssms.simulator() have the 'metadata' key, which also has that.

src/hssm/plotting/utils.py
AlexanderFengler: ok overlooked the return

AlexanderFengler (approved):
ok looks good now :)

src/hssm/distribution_utils/dist.py
digicosmos86: I mean we do have one step to figure out number of responses from the data, but we might want the user to be a bit more explicit in non-binary cases. In any case we might need a separate PR to address this

src/hssm/plotting/quantile_probability.py
digicosmos86: Sorry just saw this. Yeah I am doing `xlabel` and `ylabel` from experience with R lol. It's hard to decide which ones you explicitly declare and which ones can be included in `kwargs`. I just thought that the users might want to be able to change these labels quickly

PR #320, opened 2023-11-14
Override defaults for link functions

digicosmos86:
Updates in this PR:

- [ ] Added hssm.Link that represents more generalized link functions.
- [ ] Added a `override_default_link` method to `hssm.Param` class
- [ ] Updated hssm.HSSM class so the `override_default_link` method is called on each `Param` before each Param is finalized
- [ ] Added a `default` parameter to hssm.HSSM to control which override strategy to use
- [ ] Updated tests to reflect these changes


src/hssm/distribution_utils/dist.py
```diff
@@ -279,19 +279,24 @@ def rng_fn(
                 )
                 out_shape = sims_out.shape[:-1]
                 replace = rng.binomial(n=1, p=p_outlier, size=out_shape).astype(bool)
-                replace = np.stack([replace, replace], axis=-1)
-                n_draws = np.prod(out_shape)
+                replace_n = int(np.sum(replace, axis=None))
+                if replace_n == 0:
+                    return sims_out
```
AlexanderFengler, line 284: I think this stuff is inherited from the other pr

src/hssm/link.py
```diff
@@ -0,0 +1,78 @@
...
+                    )
+                self.link = self._make_generalized_logit_simple(*bounds)
+                self.linkinv = self._make_generalized_sigmoid_simple(*bounds)
+                self.linkinv_backend = self._make_generalized_sigmoid_simple(*bounds)
+        else:
+            bmb.Link.__init__(name, link, linkinv, linkinv_backend)
+
+        self.bounds = bounds
+
+    def _make_generalized_sigmoid_simple(self, a, b):
```
AlexanderFengler, line 64: I think you want to use pytensor operations here no?
At least the linkinv_backend should be pytensor, the others are prob ok.
The example I gave had all pytensor? I think so. (So all pytensor should work either way, since I tested that)

src/hssm/plotting/__init__.py
```diff
@@ -1,5 +1,6 @@
 """Plotting functionalities for HSSM."""
 
 from .posterior_predictive import plot_posterior_predictive
+from .quantile_probability import plot_quantile_probability
```
AlexanderFengler, line 4: from here it's inherited from previous pr I think?

AlexanderFengler (changes requested):
General comment:
- we should probably rename the `default` argument towards `prior_settings` or something like that.

Otherwise looks good, apart from the pytensor operations for the link functions.

src/hssm/link.py
digicosmos86: Actually `bambi` uses `numpy` functions, so I just want to make it consistent with other link functions in bambi. https://github.com/bambinos/bambi/blob/main/bambi/families/link.py

AlexanderFengler: kind of confusing because in the docs for Link it says:
'''
    linkinv_backend : function
        Same than ``linkinv`` but must be something that works with PyMC backend (i.e. it must
        work with PyTensor tensors). Does not need to be specified when ``name`` is a known
        name.
        '''

AlexanderFengler: do numpy --> pt happen automatically?
because without pt it's not going to be differentiable...

digicosmos86: I wouldn't worry too much about this. I basically copied the docs from Bambi, and numpy should be compatible with pytensor. Also, we are not touching `linkinv_backend` here. Actually none of the bambi-defined link functions uses `link_backend`

digicosmos86: > do numpy --> pt happen automatically? because without pt it's not going to be differentiable...

I think there's some interpretability between numpy and pytensor. Otherwise none of the bambi link functions would work

AlexanderFengler: ok!

AlexanderFengler (approved):
I think all conversations are resolved?
Approved.

PR #331, opened 2023-11-28
Added the ability to use safe priors for hierarchical models

digicosmos86:


src/hssm/hssm.py
```diff
@@ -24,6 +24,7 @@
 import seaborn as sns
 import xarray as xr
 from bambi.model_components import DistributionalComponent
+from bambi.transformations import transformations_namespace
```
AlexanderFengler, line 27: what is that one for actually?

src/hssm/param.py
```diff
@@ -531,3 +654,14 @@ def _make_default_prior(bounds: tuple[float, float]) -> bmb.Prior:
         return bmb.Prior("TruncatedNormal", mu=lower, lower=lower, sigma=2.0)
     else:
         return bmb.Prior(name="Uniform", lower=lower, upper=upper)
+
+
+def merge_dicts(dict1: dict, dict2: dict) -> dict:
```
AlexanderFengler, line 659: strictly speaking this is not a merge right?

seems like if `key in merged` but instances are not dicts, you override the value of dict1 with whatever you find in dict2.

So maybe something like `override_dict` ?

AlexanderFengler (approved):
I think this looks good now. 
Looked over it again and taking into account our discussion from yesterday (thurs)

src/hssm/param.py
digicosmos86: This actually raises a question: should the override be recursive? For example, if `1|participant_id` has a defined prior, with only `sigma` having a defined hyperprior, should we also add our default `mu`? It seems that if we don't then a `bambi` or `pymc` default will be applied

AlexanderFengler: I would say no, because we would in fact not cover the case in all generality even then. 
The logic would only apply if the user defined distribution `1|participant_id` matches the one for which we have hyper-priors available. 

If the user defines their own prior, they should define it all the way down I would say.

pyproject.toml
```diff
@@ -23,14 +23,14 @@ numpy = ">=1.23.4,<1.26"
 onnx = "^1.12.0"
 jax = "^0.4.0"
 jaxlib = "^0.4.0"
-ssm-simulators = "0.5.1"
+ssm-simulators = "0.5.3"
```
AlexanderFengler, line 26: can be bumped to 0.6.1 now

AlexanderFengler (approved):
looks good now :).
couldn't find any serious issue anymore.

PR #344, opened 2024-01-16
Replace negative non-decision time in likelihood computation with a small value

digicosmos86:
In the analytical likelihood computation, wherever `rt - t <= epsilon`, we replace the likelihood value with a very small number. However, this is not done for other types of likelihoods.

We add a `ensure_positive_ndt` function and apply it at the end of all likelihood computation to ensure this.

In doing so, we assume that all parameter named "t" means non-decision time. There is a default parameter in `race_no_bias_angle_4` named `ndt`. We are now changing it to `t` to conform with this assumption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,13 +31,13 @@ HSSM is a Python toolbox that provides a seamless combination of state-of-the-ar
 
 ## Installation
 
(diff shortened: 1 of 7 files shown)


AlexanderFengler (approved):
ok looks good :).
thanks.

PR #358, opened 2024-02-20
Support for Go-Nogo and deadlines with missing data

digicosmos86:


src/hssm/distribution_utils/onnx/onnx.py
```diff
@@ -51,6 +52,63 @@ def make_jax_logp_funcs_from_onnx(
...
+                The value of vmap_logp at which the VJP is evaluated, typically is just
+                vmap_logp(data, *dist_params)
+
+            Returns
+            -------
+            list[ArrayLike]
+                The VJP of the log-likelihood function computed at gz.
+            """
+            _, vjp_fn = vjp(vmap_logp_no_data, *dist_params)
+            return vjp_fn(gz)[1:]
```
AlexanderFengler, line 108: just as a question. Can't immediately remember why the return here skips `0` index.

src/hssm/distribution_utils/onnx/onnx.py
```diff
@@ -253,8 +311,147 @@ def logp_op_dispatch(op, **kwargs):  # pylint: disable=W0612,W0613
...
+
+        def grad(self, inputs, output_gradients):
+            """Perform the pytensor.grad() operation.
+
+            Parameters
+            ----------
+            inputs
+                The same as the inputs produced in `make_node`.
+            output_gradients
+                Holds the results of the perform `perform` method.
```
AlexanderFengler, line 386: drop one perform

src/hssm/distribution_utils/onnx/onnx.py
```diff
@@ -280,14 +477,14 @@ def make_pytensor_logp(
 
     def logp(data: np.ndarray, *dist_params: list[float | ArrayLike]) -> ArrayLike:
         # Specify input layer of MLP
-        data = data.reshape((-1, 2))
-        inputs = pt.zeros((data.shape[0], len(dist_params) + 2))
+        data = data.reshape((-1, data_dim)) if data_dim > 1 else data
```
AlexanderFengler, line 480: don't get the purpose of this reshape

src/hssm/hssm.py
```diff
@@ -219,9 +217,17 @@ def __init__(
         link_settings: Literal["log_logit"] | None = None,
         prior_settings: Literal["safe"] | None = None,
         extra_namespace: dict[str, Any] | None = None,
+        missing_data: bool = False,
+        deadline: bool = False,
+        na_value: float = -999.0,
```
AlexanderFengler, line 222: maybe `missing_data_value` here instead of `na_value`?

AlexanderFengler (commented):
Definitely taking shape! :)

pyproject.toml
```diff
@@ -18,18 +18,17 @@ keywords = ["HSSM", "sequential sampling models", "bayesian", "bayes", "mcmc"]
...
 huggingface-hub = "^0.15.1"
-onnxruntime = "^1.15.0"
-bambi = "^0.12.0"
-numpyro = "^0.12.1"
+bambi = "^0.13.0"
+numpyro = "^0.13.2"
 hddm-wfpt = "^0.1.1"
-seaborn = "^0.13.0"
+seaborn = "^0.13.2"
 pytensor = "<2.17.4"
```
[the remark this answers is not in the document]
digicosmos86: Yes, especially in non-conda environments

src/hssm/config.py
```diff
@@ -86,19 +88,21 @@ def from_defaults(
...
                     model_name,
-                    loglik_kind,
+                    loglik_kind=loglik_kind,
+                    response=default_config["response"],
                     list_params=default_config["list_params"],
                     description=default_config["description"],
                 )
 
-            return Config(model_name, loglik_kind)
+            return Config(model_name, loglik_kind, response=["rt", "response"])
```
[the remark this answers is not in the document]
digicosmos86: The logic here is that when "model_name" supplied is not in any of the supported models (meaning the user is using a custom model), we by default hard-code `response` to be `["rt", "response"]`. This will get overridden by the `response` field in the `model_config` that they supply. This way, we don't require them to define `response` fields for compatibility with existing code base

src/hssm/defaults.py
```diff
@@ -48,9 +49,19 @@ class LoglikConfig(TypedDict):
 LoglikConfigs = dict[LoglikKind, LoglikConfig]
 
 
+class MissingDataNetwork(Enum):
+    """Enum for the missing data network."""
+
+    NONE = 0
+    CPN = 1
+    OPN = 2
+    OPN_WITH_MISSING_DATA = 3
```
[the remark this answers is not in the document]
digicosmos86: Okay sounds good :)

src/hssm/distribution_utils/blackbox.py
```diff
@@ -0,0 +1,117 @@
...
+
+            Parameters
+            ----------
+            data
+                A two-column numpy array with response time and response.
+            dist_params
+                A list of parameters used in the likelihood computation. The parameters
+                can be both scalars and arrays.
+            """
+            inputs = [
```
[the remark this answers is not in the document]
digicosmos86: The function signatures are also different, which means there is no way of using if-statements

src/hssm/distribution_utils/dist.py
```diff
@@ -592,109 +510,87 @@ def create_extra_pps_coord(self):
         return np.arange(2)
 
 
-def make_blackbox_op(logp: Callable) -> Op:
-    """Wrap an arbitrary function in a pytensor Op.
+def make_likelihood_callable(
+    loglik: pytensor.graph.Op | Callable | PathLike | str,
+    loglik_kind: Literal["analytical", "approx_differentiable", "blackbox"],
+    backend: Literal["pytensor", "jax", "cython", "other"] | None,
```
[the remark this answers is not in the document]
digicosmos86: We might want this. It is possible to implement analytical likelihoods in cython right? Our likelihood kinds might need some rethinking, since they are not mutually exclusive...

src/hssm/distribution_utils/onnx/onnx.py
digicosmos86: Here's the Jax Doc

> If has_aux is False, returns a (primals_out, vjpfun) pair, where primals_out is fun(*primals). If has_aux is True, returns a (primals_out, vjpfun, aux) tuple where aux is the auxiliary data returned by fun.

> vjpfun is a function from a cotangent vector with the same shape as primals_out to a tuple of cotangent vectors with the same number and shapes as primals, representing the vector-Jacobian product of fun evaluated at primals.

Looks like we don't need the primals

src/hssm/distribution_utils/onnx/onnx.py
digicosmos86: This handles a corner case when there is only one trial and `rt, response` is provided as a 1d vector

src/hssm/hssm.py
digicosmos86: I am fine with both. `na_value` is shorter and is also a convention in `pandas`. Let me know which one you prefer

pyproject.toml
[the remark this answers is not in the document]
AlexanderFengler: Ok let's follow gameplan (this is the wrong place to put it but nevertheless :)):
1. add the new network types
2. fix initialization issues 
3. get to conda
4. finish RL piece
5. web interface

src/hssm/distribution_utils/blackbox.py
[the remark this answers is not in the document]
AlexanderFengler: hm ok. Let's shelf the discussion, until some stroke of genius :)

src/hssm/distribution_utils/dist.py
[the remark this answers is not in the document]
AlexanderFengler: agree that we may want to change the naming conventions entirely eventually. 
For now, cython stuff falls under blackbox (since no gradients) and I am just questioning if it really falls under 'backend' category in any relevant sense. 
It's just some pyhton callable in the end.

src/hssm/distribution_utils/onnx/onnx.py
AlexanderFengler: Gotta love how complicated this reads...

src/hssm/distribution_utils/onnx/onnx.py
AlexanderFengler: perform 'perform'.
just about the docs :)

src/hssm/hssm.py
AlexanderFengler: If we have `missing_data` kwarg, then probably go with `missing_data_value` kwarg to make it consistent. 

We can also allow both `missing_data_value` and `na_value` and pool internally, if you think it's worth the burden. 

In general appreciate the convention argument, but locally I think going for `missing_data` AND `na_value` is confusing.

src/hssm/distribution_utils/dist.py
```diff
@@ -594,3 +593,74 @@ def make_likelihood_callable(
...
+        loglik, "approx_differentiable", backend, params_is_reg, 0 if is_cpn_only else 1
+    )  # Just assume that the missing data network is always approx_differentiable
+
+
+def assemble_callables(
+    callable: pytensor.graph.Op | Callable,
+    missing_data_callable: pytensor.graph.Op | Callable,
+    is_cpn_only: bool,
+) -> Callable:
+    """Assemble the likelihood callables into a single callable.
```
AlexanderFengler, line 612: drop duplicate sentence

AlexanderFengler (approved):
Looks good now. 
Couldn't find any apparent issue. 
Just left a comment on some doc-string nonsense :).

src/hssm/distribution_utils/blackbox.py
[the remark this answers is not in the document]
digicosmos86: Actually I figured this out. Have already implemented this in the other file (src/onnx/onnx.py). Will fix this one too

PR #383, opened 2024-04-02
by default trial wise parameters defined as deterministics are now ex…

AlexanderFengler:
…cluded from the posteriors


AlexanderFengler:
Stale at this point.

Issue #387, opened 2024-04-05
Specifying priors for categorical variables in regression does not work


eort:
**Describe the bug**

When running a (hierarchical) regression model and I try to specify priors for categorical variables I receive following error: 
`TypeError: Wrong number of dimensions: expected 1, got 0 with shape ().` (full stack trace below). I only tried this for the `angle` model 

**HSSM version**

0.2

**To Reproduce**
```
import hssm
from pandas import read_csv

data = read_csv('cavanagh_theta_nn.csv')
model = hssm.HSSM(data=data,
                  model='angle',
                  loglik='angle.onnx',
                  loglik_kind='approx_differentiable',
                  include=[
                    {
                        "name": "v",
                        "formula":"v ~ (1 | participant_id) + conf",
                        "prior": {
                            "Intercept": {"name": "Uniform", "lower": -1, "upper": 1, "initval":0},
                            "conf": {"name": "Normal", "mu": 0, "sigma": 1, "initval":0}
                        }
                    }
                ],
                  hierarchical=True,
                  link_settings='log_logit'
)
```

Full stack trace:
Traceback (most recent call last):
```
  File "/gpfs/project/ort/lorasick/ddm/code/hssm_reg.py", line 50, in <module>
    model = hssm.HSSM(data=data,
            ^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/project/projects/bpsydm/tools/pyEnvs/hssm/lib/python3.11/site-packages/hssm/hssm.py", line 334, in __init__
    self.set_alias(self._aliases)
  File "/gpfs/project/projects/bpsydm/tools/pyEnvs/hssm/lib/python3.11/site-packages/hssm/hssm.py", line 594, in set_alias
    self.model.build()
  File "/gpfs/project/projects/bpsydm/tools/pyEnvs/hssm/lib/python3.11/site-packages/bambi/models.py", line 350, in build
    self.backend.build(self)
  File "/gpfs/project/projects/bpsydm/tools/pyEnvs/hssm/lib/python3.11/site-packages/bambi/backend/pymc.py", line 70, in build
    self.components[name].build(self, spec)
  File "/gpfs/project/projects/bpsydm/tools/pyEnvs/hssm/lib/python3.11/site-packages/bambi/backend/model_components.py", line 60, in build
    self.build_common_terms(pymc_backend, bmb_model)
[...]

PR #391, opened 2024-04-12
Moved fixed effects before random effects in formula specifications in HSSM documentation

digicosmos86:
Per #387, sometimes specifying random effects before fixed effects can cause weird issues (which could be a bambi bug). In this PR, fixed effects are moved before random effects in HSSM documentation to avoid such issues. It is also more consistent with Bambi documentation

`numpy` and `scipy` are also removed from dependencies to prevent dependency conflicts, since they always come with `pymc` and `arviz`.


AlexanderFengler (approved):
yep looks good :). 
strange one, if we are fairly certain it's bambi, we should prob raise there. 
Might be an issue with the formula parser?

PR #408, opened 2024-05-01
change order between sz and sv in list_params

AlexanderFengler:


digicosmos86:
BTW you also need to change the order of `sz` and `sv` here in the definition of the function

https://github.com/lnccbrown/HSSM/blob/64e356e2e60820e6573dfee1e71b401f11d00e90/src/hssm/likelihoods/blackbox.py#L64-L76

digicosmos86 (approved):
Looks great! Just some minor comments

AlexanderFengler:
ok now @digicosmos86 ?

digicosmos86:
> ok now @digicosmos86 ?

Yep! Feel free to merge 😃

PR #420, opened 2024-05-08
Drop black in favor of ruff formatter

### LINKED ISSUE #419: Drop Black in favor of ruff formatter
As of v0.1.2, Ruff has introduced a formatter as a drop-in replacement for `black`. This PR removes `black` as a dependency in favor of `ruff` as the formatter.

The following tasks needs to be done:

- [x] Remove black as a dependency and update ruff version
- [x] Update pre-commit config to use ruff instead of black
- [x] Update ci script to perform ruff formatter checks instead of black.

Locally, also use ruff plug-ins for formatting

digicosmos86:
Closes #419

### DIFF
--- a/.github/workflows/build_and_publish.yml
+++ b/.github/workflows/build_and_publish.yml
@@ -41,13 +41,13 @@ jobs:
           pip install -e .
 
       - name: Run mypy
-        run: mypy src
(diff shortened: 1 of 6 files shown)


AlexanderFengler (approved):
ok looks good!

PR #436, opened 2024-05-20
414 long running time of sample posterior predictive and eventual death by oom

AlexanderFengler:
- posterior predictive now has a `safe_mode` that chunks computations
- the `n_samples` argument was renamed to `draws`, and one can pass None | int | list | np.ndarray
- when running posterior predictive with `kind='mean'`, the posterior naming cleans up `rt,response_mean` --> `v`
- prior predictives get assigned to `.traces` now, and naming is also cleaned up
- our `sample_prior_predictive()` will include the parent parameter as well now via an internal call to `.predict()`


AlexanderFengler:
I will try to add a few more tests to this before merging.

src/hssm/hssm.py
```diff
@@ -537,7 +539,8 @@ def sample_posterior_predictive(
         inplace: bool = True,
         include_group_specific: bool = True,
         kind: Literal["pps", "mean"] = "pps",
-        n_samples: int | float | None = None,
+        draws: int | float | list[int] | None = None,
+        safe_mode: bool = True,
```
jainraj, line 543: I suggest adding `safe_mode` in the function documentation.

src/hssm/hssm.py
```diff
@@ -678,7 +804,25 @@ def sample_prior_predictive(
...
+        prior_predictive.prior = prior_predictive.posterior
+        del prior_predictive.posterior
+
+        if self._inference_obj is None:
+            self._inference_obj = prior_predictive
+        else:
+            self._inference_obj.extend(prior_predictive)
+
+        # clean up `rt,response_mean` to `v`
+        self._drop_parent_str_from_idata(self._inference_obj)
```
digicosmos86, line 824: The type checker is a bit stupid here since the `if` block above basically indicates `self._inference_obj` is not None at this point. Maybe the easiest way to do this is to move this line above Line 818 and change the argument to `prior_predictive`?

src/hssm/hssm.py
```diff
@@ -678,7 +804,25 @@ def sample_prior_predictive(
...
+        return self._inference_obj
```
digicosmos86, line 825: I don't quite understand the logic here. `sample_prior_predictive` can be called before any sampling is done, right? Then why should we attach it to self._inference_obj? Then when you call `sample`, this object will be overridden after sampling. It doesn't quite work with some of the logic in the rest of the code either.

digicosmos86 (changes requested):
Looks good! Two higher-level comments:

1. Since the only call to `simulator` is done here: https://github.com/lnccbrown/HSSM/blob/19f786d9abeca00aa6af6d0740598bebb80ae6d3/src/hssm/distribution_utils/dist.py#L309-L315, maybe we can use a for-loop here over n_samples to make the sampling safe, instead of patching the higher-level functions themselves? This way we can avoid running many intermediate-level code multiple times.

2. `InferenceData` object does not come with attributes like `posterior`, or `posterior_predictive` by default, so type checker complains. The use of the square bracket notation is preferred. Or if this is too annoying we can disable this check (`attr-defined`) globally in `pyproject.toml` `mypy` section, but that can be a bit risky

src/hssm/hssm.py
```diff
@@ -514,15 +516,15 @@ def sample(
                 pass
 
         self._inference_obj = self.model.fit(
-            inference_method=sampler, init=init, **kwargs
+            inference_method=sampler, init=init, include_mean=True, **kwargs
```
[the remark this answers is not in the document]
AlexanderFengler: Ok will refactor towards simply preprocessing kwargs

src/hssm/hssm.py
AlexanderFengler: Adressed

src/hssm/hssm.py
```diff
@@ -404,6 +405,7 @@ def __init__(
             self.model, self._parent_param, self.response_c, self.response_str
         )
         self.set_alias(self._aliases)
+        print(self.pymc_model.initial_point())
```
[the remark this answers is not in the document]
AlexanderFengler: done

src/hssm/hssm.py
```diff
@@ -514,15 +516,15 @@ def sample(
                 pass
```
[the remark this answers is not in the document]
AlexanderFengler: done

src/hssm/hssm.py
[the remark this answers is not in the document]
AlexanderFengler: done

src/hssm/hssm.py
```diff
@@ -592,34 +595,157 @@ def sample_posterior_predictive(
                     + "Please either provide an idata object or sample the model first."
                 )
             idata = self._inference_obj
+            print("idata=None, we use the traces assigned to the HSSM object as idata.")
+
+        if "posterior_predictive" in idata.groups():
+            del idata.posterior_predictive
+            print("pre-existing posterior_predictive group deleted from idata. \n")
```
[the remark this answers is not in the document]
AlexanderFengler: done

src/hssm/hssm.py
```diff
@@ -592,34 +595,157 @@ def sample_posterior_predictive(
                     + "Please either provide an idata object or sample the model first."
                 )
             idata = self._inference_obj
+            print("idata=None, we use the traces assigned to the HSSM object as idata.")
```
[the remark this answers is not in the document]
AlexanderFengler: done

src/hssm/hssm.py
```diff
@@ -592,34 +595,157 @@ def sample_posterior_predictive(
...
+                # we can simply ignore settings, since the computational overhead
+                # should be very small --> nudges user towards good outputs.
+                _logger.warning(
+                    "The kind argument is set to 'mean', but 'draws' argument "
+                    + "is not None: The draws argument will be ignored"
+                )
+                return self.model.predict(
+                    idata, kind, data, inplace, include_group_specific
+                )
+        else:
```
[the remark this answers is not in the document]
AlexanderFengler: ok followed this suggestion, any format of draws is now first transformed --> `np.ndarray`

src/hssm/hssm.py
```diff
@@ -1353,6 +1497,33 @@ def _get_deterministic_var_names(self, idata) -> list[str]:
 
         return var_names
 
+    def _drop_parent_str_from_idata(
+        self, idata: az.InferenceData, inplace=True
+    ) -> az.InferenceData | None:
```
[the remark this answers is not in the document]
AlexanderFengler: changed

src/hssm/utils.py
```diff
@@ -452,3 +452,8 @@ def _rearrange_data(data: pd.DataFrame | np.ndarray) -> pd.DataFrame | np.ndarra
     split_not_missing = data[~missing_indices, :]
 
     return np.concatenate([split_missing, split_not_missing])
+
+
+def _split_array(data: np.ndarray | list[int], divisor: int):
```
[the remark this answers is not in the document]
AlexanderFengler: done

src/hssm/hssm.py
AlexanderFengler: ok worked out the type checking stuff

src/hssm/hssm.py
AlexanderFengler: had discussion offline on this one.

src/hssm/hssm.py
```diff
@@ -510,24 +511,41 @@ def sample(
...
-                self.model.predict(self._inference_obj, kind="mean", inplace=True)
+                # self.model.predict(self._inference_obj, kind="mean", inplace=True)
                 # rename 'rt,response_mean' to 'v' so in the traces everything
                 # looks the way it should
                 self._inference_obj.rename_vars(
                     {"rt,response_mean": self._parent}, inplace=True
                 )
+            elif (
+                self._parent in self._inference_obj.posterior.data_vars.keys()
+                and "rt,response_mean" in self._inference_obj.posterior.data_vars.keys()
```
digicosmos86, line 544: `data_vars` are `dicts`, so the Python 3 style is to not use `keys()`

src/hssm/hssm.py
```diff
@@ -592,35 +614,120 @@ def sample_posterior_predictive(
...
+        # of the original (draw dimension gets sub-sampled)
+
+        idata_copy = idata.copy()
+
+        if (draws.shape != idata["posterior"].draw.values.shape) or (
+            (draws.shape == idata["posterior"].draw.values.shape)
+            and not np.allclose(draws, idata["posterior"].draw.values)
+        ):
+            # Reassign posterior to sub-sampled version
+            setattr(idata_copy, "posterior", idata["posterior"].isel(draw=draws))
```
digicosmos86, line 656: Are there any differences between `setattr()` and `idata.add_groups()`?

src/hssm/hssm.py
```diff
@@ -592,35 +614,120 @@ def sample_posterior_predictive(
...
                 return None
-
+            else:
+                # Not safe mode and not inplace
+                # Function acts as very thin wrapper around
+                # .predict(). It just operates on the
+                # idata_copy object
+                return self.model.predict(
+                    idata_copy, kind, data, inplace, include_group_specific
+                )
```
digicosmos86, line 717: This if block looks slightly confusing. I think I understand what you mean, but would 

```python
if safe_mode:
    if inplace:
        ...
    else:
        ...
else:
    if inplace:
         ...
    else:
         ...
```
be more readable?

src/hssm/hssm.py
```diff
@@ -592,35 +614,120 @@ def sample_posterior_predictive(
...
+            # we can simply ignore settings, since the computational overhead
+            # should be very small --> nudges user towards good outputs.
+            _logger.warning(
+                "The kind argument is set to 'mean', but 'draws' argument "
+                + "is not None: The draws argument will be ignored!"
+            )
             return self.model.predict(
-                idata_copy, kind, data, False, include_group_specific
+                idata, kind, data, inplace, include_group_specific
             )
```
digicosmos86, line 730: Add an `else` clause here to throw an error whenever other values are specified?

src/hssm/hssm.py
```diff
@@ -1353,6 +1477,35 @@ def _get_deterministic_var_names(self, idata) -> list[str]:
 
         return var_names
 
+    def _drop_parent_str_from_idata(
+        self, idata: Union[az.InferenceData, None]
```
digicosmos86, line 1481:
```suggestion
        self, idata: az.InferenceData | None
```

digicosmos86 (approved):
Looks awesome! Just some style suggestions at this point. Feel free to merge after the fixes :)

src/hssm/hssm.py
```diff
@@ -592,35 +607,120 @@ def sample_posterior_predictive(
                     + "Please either provide an idata object or sample the model first."
                 )
             idata = self._inference_obj
+            _logger.info(
+                "idata=None, we use the traces assigned to the HSSM object as idata."
+            )
+
+        if "posterior_predictive" in idata.groups():
+            if idata is not None:
```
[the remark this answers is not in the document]
AlexanderFengler: @digicosmos86 changed. This was useless to begin with, just an artifact appeasing mypy...

src/hssm/hssm.py
```diff
@@ -10,7 +10,7 @@
 from copy import deepcopy
 from inspect import isclass
 from os import PathLike
-from typing import Any, Callable, Literal
+from typing import Any, Callable, Literal, Union
```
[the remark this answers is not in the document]
AlexanderFengler: changed

src/hssm/hssm.py
```diff
@@ -404,6 +404,7 @@ def __init__(
             self.model, self._parent_param, self.response_c, self.response_str
         )
         self.set_alias(self._aliases)
+        # _logger.info(self.pymc_model.initial_point())
```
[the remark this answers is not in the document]
AlexanderFengler: Stylistically eventually yes, but rn, I think it can sometimes still help future PRs that interact with this code. Here I literally have the next PR that I need to work on in mind. So in general agree, but let's skip here :)

src/hssm/hssm.py
AlexanderFengler: changed

src/hssm/hssm.py
AlexanderFengler: to be honest I don't know... let me look into that independently to understand it properly.

src/hssm/hssm.py
AlexanderFengler: fixed that

src/hssm/hssm.py
AlexanderFengler: changed

src/hssm/hssm.py
AlexanderFengler: Actually, at least used somewhat semantically here, `add_groups` is about new groups, `setattr` is about reassigning to pre-existing group.

PR #441, opened 2024-05-22
One-command support for JAX GPU

digicosmos86:
After this pr is merged, one can install `hssm` with GPU support with `pip install hssm[cuda12]` (not available on cuda)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,7 +25,7 @@ bambi = "^0.13.0"
 numpyro = "^0.15.0"
 hddm-wfpt = "^0.1.4"
 seaborn = "^0.13.2"
-jax = "^0.4.23"
+jax = { version = "^0.4.23", extras = ["cuda12"], optional = true }
 
 [tool.poetry.group.dev.dependencies]
 pytest = "^8.2.0"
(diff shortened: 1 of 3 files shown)


AlexanderFengler:
not available on `conda` :)

pyproject.toml
```diff
@@ -25,7 +25,7 @@ bambi = "^0.13.0"
 numpyro = "^0.15.0"
 hddm-wfpt = "^0.1.4"
 seaborn = "^0.13.2"
-jax = "^0.4.23"
+jax = { version = "^0.4.23", extras = ["cuda12"], optional = true }
```
AlexanderFengler, line 28: I think for now we have to cap jax roughly at <= 0.4.26, because of the config issue... setting floatX breaks on newest.

AlexanderFengler (approved):
high level question: 
Does this operate via a "synonym" on pypi (listed under HSSM), or do we basically end up with two packages on pypi with this?

pyproject.toml
digicosmos86: Looks like they fixed this? I have jax==0.4.28 and this issue did not pop up for me...

digicosmos86:
> 

Correct. conda lol

pyproject.toml
digicosmos86: Fixed

PR #446, opened 2024-05-24
fix key error bug when going through initialization, make safe priors…

AlexanderFengler:
- fixes key error bug when going through initial value setters
- make initial value setters optional
- make safe priors the default
- if user supplied initial value to a parameter that we try to overwrite --> skip overwrite
- allow map as starting point in sampler and add `find_MAP()` method.
- fixing some redundant starting value settings

### DIFF
--- a/src/hssm/hssm.py
+++ b/src/hssm/hssm.py
@@ -10,7 +10,7 @@ import logging
 from copy import deepcopy
 from inspect import isclass
 from os import PathLike
-from typing import Any, Callable, Literal
+from typing import Any, Callable, Literal, Union
 
 import arviz as az
 import bambi as bmb
@@ -1035,6 +1035,26 @@ class HSSM:
 
         return az.summary(data, **kwargs)
 
+    def initial_point(self, transformed: bool = False) -> dict[str, np.ndarray]:
+        """Compute the initial point of the model.
+
+        This is a slightly altered version of pm.initial_point.initial_point().
+
+        Parameters
+        ----------
+        transformed : bool, optional
+            If True, return the initial point in transformed space.
+
+        Returns
+        -------
+        dict
+            A dictionary containing the initial point of the model parameters.
+        """
+        fn = pm.initial_point.make_initial_point_fn(
+            model=self.pymc_model, return_transformed=transformed
+        )
+        return pm.model.Point(fn(None), model=self.pymc_model)
+
     def __repr__(self) -> str:
         """Create a representation of the model."""
         output = []
@@ -1682,10 +1702,6 @@ class HSSM:
             )
             return None
 
-        # Store overall link setting string
-        # Note the call to str(), turns None to `None`
-        link_setting_str = str(self.link_settings)
-
         # Set initial values for particular parameters
         for name_, starting_value in self.pymc_model.initial_point().items():
             # strip name of `_log__` and `_interval__` suffixes
@@ -1701,7 +1717,7 @@ class HSSM:
             # If the user actively supplies a link function, the user
             # should also have supplied an initial value insofar it matters.
             if self.params[self._get_prefix(name_tmp)].is_regression:
-                param_link_setting = link_setting_str
+                param_link_setting = self.link_settings
             else:
                 param_link_setting = None
             if name_tmp in initval_settings[param_link_setting].keys():
@@ -1733,7 +1749,11 @@ class HSSM:
             name_str_prefix = name_str
         return name_str_prefix
 
-    def _check_if_initval_user_supplied(self, name_str: str) -> bool:
+    def _check_if_initval_user_supplied(
+        self,
+        name_str: str,
+        return_value: bool = False,
+    ) -> Union[bool, float, int, np.ndarray]:
         """Check if initial value is user-supplied."""
         # The function assumes that the name_str is either raw parameter name
         # or `paramname_Intercept`, because we only really provide special default
@@ -1745,13 +1765,15 @@ class HSSM:
         if "_" in name_str:
             if "p_outlier" not in name_str:
                 name_str_prefix = name_str.split("_")[0]
-                name_str_suffix = name_str.split("_")[1]
+                # name_str_suffix = "".join(name_str.split("_")[1:])
+                name_str_suffix = name_str[len(name_str_prefix + "_") :]
             else:
                 name_str_prefix = "p_outlier"
                 if name_str == "p_outlier":
                     name_str_suffix = ""
                 else:
-                    name_str_suffix = name_str.split("_")[2]
+                    # name_str_suffix = "".join(name_str.split("_")[2:])
+                    name_str_suffix = name_str[len("p_outlier_") :]
         else:
             name_str_prefix = name_str
             name_str_suffix = ""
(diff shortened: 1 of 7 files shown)


src/hssm/hssm.py
```diff
@@ -1619,24 +1671,103 @@ def _postprocess_initvals_deterministic(
...
+                        " skipping overwrite with default value.",
+                        name_tmp,
+                    )
+                    continue
+                else:
+                    # Apply specific settings from initval_settings dictionary
+                    self.pymc_model.set_initval(
+                        self.pymc_model.named_vars[name_tmp],
+                        initval_settings[param_link_setting][name_tmp],
+                    )
```
digicosmos86, line 1715: No need for `else` here because of `continue`

src/hssm/hssm.py
```diff
@@ -1619,24 +1671,103 @@ def _postprocess_initvals_deterministic(
         # Consider case where link functions are set to 'log_logit'
         # or 'None'
         if self.link_settings not in ["log_logit", "None", None]:
```
digicosmos86, line 1673: We do not allow `link_settings` to be a string `"None"` per type definition of argument

src/hssm/hssm.py
```diff
@@ -1619,24 +1671,103 @@ def _postprocess_initvals_deterministic(
...
+            # If not, we don't actually apply a link function to it as per default.
+            # Therefore we need to apply the initial value strategy corresponding
+            # to 'None' link function.
+
+            # If the user actively supplies a link function, the user
+            # should also have supplied an initial value insofar it matters.
+            if self.params[self._get_prefix(name_tmp)].is_regression:
+                param_link_setting = link_setting_str
+            else:
+                param_link_setting = "None"
```
digicosmos86, line 1701: I'd rather use `None` instead of the string

digicosmos86 (changes requested):
Looks great! Just a few stylistic suggestions. Do we have tests in place for the added functions? I'm a bit nervous about things going production without corresponding tests

src/hssm/hssm.py
```diff
@@ -405,17 +405,32 @@ def __init__(
...
 
     def sample(
         self,
         sampler: (
             Literal["mcmc", "nuts_numpyro", "nuts_blackjax", "laplace", "vi"] | None
         ) = None,
         init: str | None = None,
+        start_at_map: bool = False,
         **kwargs,
     ) -> az.InferenceData | pm.Approximation:
```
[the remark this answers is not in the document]
AlexanderFengler: tried to simplify this, now we allow 'initvals' to be passed as 'map' / dict / None.

src/hssm/hssm.py
```diff
@@ -444,6 +464,20 @@ def sample(
...
+                    "start_at_map = True but no map estimate precomputed. \n"
+                    "Running map estimation first..."
+                )
+                self.find_MAP()
+                kwargs["initvals"] = self._map_dict
+        elif start_at_map and ("initvals" in kwargs):
+            _logger.warning(
+                "start_at_map = True, but initvals is provided. "
+                + "The initvals argument will be used."
+            )
```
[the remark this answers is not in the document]
AlexanderFengler: see comment above

src/hssm/hssm.py
```diff
@@ -1619,24 +1671,103 @@ def _postprocess_initvals_deterministic(
...
+        # `p_outlier` is the only basic parameter floating around that has
+        # an underscore in it's name.
+        # We need to handle it separately. (Renaming might be better...)
+        if "_" in name_str and "p_outlier" not in name_str:
+            name_str_prefix = name_str.split("_")[0]
+        elif "_" in name_str and "p_outlier" in name_str:
+            name_str_prefix = "p_outlier"
+        else:
+            name_str_prefix = name_str
+        return name_str_prefix
```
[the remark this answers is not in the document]
AlexanderFengler: looks slightly different now

src/hssm/hssm.py
```diff
@@ -1619,24 +1671,103 @@ def _postprocess_initvals_deterministic(
...
+            name_str_suffix = name_str.split("_")[1]
+        elif ("_" in name_str) and ("p_outlier" in name_str):
+            name_str_prefix = "p_outlier"
+            if name_str == "p_outlier":
+                name_str_suffix = ""
+            else:
+                name_str_suffix = name_str.split("_")[2]
+        else:
+            name_str_prefix = name_str
+            name_str_suffix = ""
```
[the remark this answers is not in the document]
AlexanderFengler: this also looks slightly different now.

src/hssm/hssm.py
```diff
@@ -1619,24 +1671,103 @@ def _postprocess_initvals_deterministic(
...
+            if name_str_prefix == self._parent:
+                tmp_param = "c(rt, response)"
+            else:
+                tmp_param = name_str_prefix
+
+            if (name_str_suffix == "Intercept") or (name_str_prefix == self._parent):
+                if "initval" in self.priors[tmp_param]["Intercept"].args:
+                    return True
+                else:
+                    return False
```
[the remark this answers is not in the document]
AlexanderFengler: logic of this function became a bit more complicated.

src/hssm/hssm.py
```diff
@@ -1619,24 +1671,103 @@ def _postprocess_initvals_deterministic(
...
+            else:
+                if "initval" in self.priors[tmp_param].args:
+                    return True
+                else:
+                    return False
```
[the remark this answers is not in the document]
AlexanderFengler: as above.

tests/test_initvals.py
```diff
@@ -0,0 +1,258 @@
...
+    # ("approx_differentiable", "ddm", "mcmc", "map"), # parallel sampling is an issue here
+    ("analytical", "ddm", "mcmc", "map"),
+    # ("approx_differentiable", "angle", "mcmc", "map"), # parallel sampling is an issue here
+    ("approx_differentiable", "ddm", "nuts_numpyro", "initial_point"),
+    ("analytical", "ddm", "nuts_numpyro", "initial_point"),
+    ("approx_differentiable", "angle", "nuts_numpyro", "initial_point"),
+    # ("approx_differentiable", "ddm", "mcmc", "initial_point"),
+    # ("analytical", "ddm", "mcmc", "initial_point"), # parallel sampling is an issue
+    # ("approx_differentiable", "angle", "mcmc", "initial_point"),
+]
```
digicosmos86, line 26: Why not set `cores=1` so the commented out tests can run?

tests/test_initvals.py
```diff
@@ -0,0 +1,258 @@
...
+        # should also have supplied an initial value insofar it matters.
+        if model.params[model._get_prefix(name_)].is_regression:
+            param_link_setting = model.link_settings
+        else:
+            param_link_setting = None
+
+        # Go through parameters that are specified in the initial value defaults
+        # If not specified in there, we won't touch the parameter during post-processing
+        # anyways
+        if name_ in hssm.defaults.INITVAL_SETTINGS[param_link_setting].keys():
```
digicosmos86, line 76: No need for `.keys()` in Python 3

tests/test_initvals.py
```diff
@@ -0,0 +1,258 @@
...
+
+                assert np.allclose(
+                    model_initial_point,
+                    default_initial_point,
+                    atol=1e-3,
+                ), f"""Initial value for {name_} is supposed to be {default_initial_point},
+                       and does not match the initial point set by model,
+                       which is {model_initial_point}."""
+        else:
+            pass
```
digicosmos86, line 117: We don't need this right?

src/hssm/hssm.py
```diff
@@ -1618,26 +1700,156 @@ def _postprocess_initvals_deterministic(
...
+                name_str_prefix = "p_outlier"
+        else:
+            name_str_prefix = name_str
+        return name_str_prefix
+
+    def _check_if_initval_user_supplied(
+        self,
+        name_str: str,
+        return_value: bool = False,
+    ) -> Union[bool, float, int, np.ndarray, None]:
```
digicosmos86, line 1761: For Python 3.10+ we use `|` instead of importing `Union`

digicosmos86 (approved):
Looks great! Nit picking a bit now LOL.

BTW, I also noticed some debug output while using HSSM. Can you check the print statements in the HSSM code? For printing out information we should be using `logging.info` anyway. Feel free to merge afterwards :)

tests/test_initvals.py
AlexanderFengler: just thought it enhances clarity

src/hssm/hssm.py
AlexanderFengler: @digicosmos86 got rid of Union in multiple places now. 
The comment is without loss of generality?

AlexanderFengler: (I guess can't get rid of it everywhere got complaints when doing so in some places)

digicosmos86: Only place where it might complain is when the `|` operator is not used in type hints, for example in `isinstance` or `cast`, but as type hints you should be all set :)

AlexanderFengler:
@digicosmos86 good to merge this now? (saw that you made one commit re print statements, is there anything else in the works?)

digicosmos86:
> @digicosmos86 good to merge this now? (saw that you made one commit re print statements, is there anything else in the works?)

Good to merge now :)

PR #452, opened 2024-06-04
Fix ensure positive ndt

digicosmos86:
This PR fixes a bug in which `ensure_positive_ndt` is applied after likelihood computation with lapse distribution, resulting in wrong likelihood calculations. This fix will likely significantly improve convergence.


AlexanderFengler (approved):
Looks good :)

PR #467, opened 2024-06-20
put in prep paper up front in readme

AlexanderFengler:


AlexanderFengler:
Just a small change in readme.

PR #468, opened 2024-06-20
Fix nan gradients in analytical likelihood

digicosmos86:


src/hssm/likelihoods/analytical.py
```diff
@@ -262,15 +241,17 @@ def logp_ddm(
...
+
+    p = pt.maximum(ftt01w(tt, a, z_flipped, err, k_terms), pt.exp(LOGP_LB))
+
+    logp = negative_rt * LOGP_LB + (1 - negative_rt) * (
         pt.log(p)
         - v_flipped * a * z_flipped
-        - (v_flipped**2 * rt / 2.0)
-        - 2.0 * pt.log(a),
+        - (v_flipped**2 * tt / 2.0)
+        - 2.0 * pt.log(pt.maximum(epsilon, a))
```
AlexanderFengler, line 254: reflecting on this a bit, 
I think this maximum business is actually corrupting the gradients, so we should just a priori restrict `a > epsilon` (via prior essentially?).

AlexanderFengler: on the other hand, apart from initialization (which 1. our strategies should already avoid, 2. we generally can impact) `a` should basically never come close to 0, so this should basically never be the culprit...

src/hssm/likelihoods/analytical.py
```diff
@@ -262,15 +241,17 @@ def logp_ddm(
...
     )
 
     checked_logp = check_parameters(logp, a >= 0, msg="a >= 0")
```
AlexanderFengler, line 257: in the spirit of above, this check could be `a>0` but honestly we shouldn't really ever get there.

src/hssm/likelihoods/analytical.py
```diff
@@ -220,7 +199,7 @@ def logp_ddm(
     z: float,
     t: float,
     err: float = 1e-15,
-    k_terms: int = 20,
+    k_terms: int = 7,
     epsilon: float = 1e-15,
```
AlexanderFengler, line 203: I don't know what was used for testing / is used as actual value for inference, but I guess it is this default?

The epsilon for the `rt` part should rather be on the order of 1e-3, or even 1e-2.

If we are reusing the same epsilon in multiple places, we should probably separate it out.

src/hssm/likelihoods/analytical.py
```diff
@@ -262,15 +241,17 @@ def logp_ddm(
     z_flipped = pt.switch(flip, 1 - z, z)  # transform z if x is upper-bound response
     rt = rt - t
 
-    p = pt.maximum(ftt01w(rt, a, z_flipped, err, k_terms), pt.exp(LOGP_LB))
+    negative_rt = rt <= epsilon
```
AlexanderFengler, line 244: Ok reflecting on this a bit, the logic that we want should probably look something like:

- flag all rts lower than epsilon
- go through with ftt01w
- then set all flagged rts to `LOGB_LB`

This should actually cut the gradient for problematic rts.
Potentially we put this as a `logp_ddm_2` and compare results / gradients.
Alternatively, if any `rt` breaches epsilon, directly send logp to -infty (this is probably not preferable).

AlexanderFengler (commented):
Looks good, this is mostly about iterating conceptually, not code quality.

AlexanderFengler:
@digicosmos86 is this stale for now?

digicosmos86:
> @digicosmos86 is this stale for now?

There doesn't seem to be a solution for really small RTs in the denominator, which can blow up

frankmj:
ah well maybe this is related to my "RT hack" I proposed as an interim
solution for cases where *t* is low. (Which I attributed to the possibility
that the sampler would end up proposing values of t that hit the lower
bound and lead to unstable gradients, but I know that pymc is supposed to
deal with such boundaries smoothly under the hood - so maybe the issue is
just the RTs in the denominator being small). In that case the RT hack
would still work (ie under the hood befor fitting just add a constant value
to all RTs (say 0.5), which should only shift the t parameter, and then
report t_new = t - 0.5).

digicosmos86:
@frankmj I ran a few more tests and the RT-hack did do the trick. It might be hard for us to implement this trick in our code though, mostly because people use `arviz` functions instead of the convenience functions that we provide, which could give us some control over the output. We could note this in our documentation somewhere about this trick so that the users can implement this themselves so that they have full control

frankmj:
Great. maybe one simple solution would be to simply add a link function
with t= t'-const  ?

digicosmos86:
@frankmj That's a great idea! I also noticed that the RT-hack only worked when `float64` is used, indicating some other issues that we might have with numerical stability. I'll look deeper into this

cpaniaguam (commented):
A few things here I might end up picking up myself.

src/hssm/likelihoods/analytical.py
```diff
@@ -342,11 +324,11 @@ def logp_ddm_sdv(
...
             - 2 * a * v_flipped * z_flipped
-            - (v_flipped**2) * rt
+            - (v_flipped**2) * tt
         )
-        / (2 * (sv**2) * rt + 2)
-        - 0.5 * pt.log(sv**2 * rt + 1)
-        - 2 * pt.log(a),
+        / (2 * (sv**2) * tt + 2)
+        - 0.5 * pt.log(sv**2 * tt + 1)
+        - 2 * pt.log(pt.maximum(epsilon, a)),
```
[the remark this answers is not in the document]
digicosmos86: We are probably not going to keep this one. I just tried this to see if we keep the log positive we can get somewhere. It helps a bit it seems, but the culprit is not this one

src/hssm/likelihoods/analytical.py
```diff
@@ -262,15 +241,17 @@ def logp_ddm(
     z_flipped = pt.switch(flip, 1 - z, z)  # transform z if x is upper-bound response
     rt = rt - t
 
-    p = pt.maximum(ftt01w(rt, a, z_flipped, err, k_terms), pt.exp(LOGP_LB))
+    negative_rt = rt <= epsilon
 
-    logp = pt.where(
-        rt <= epsilon,
-        LOGP_LB,
+    tt = negative_rt * epsilon + (1 - negative_rt) * rt
```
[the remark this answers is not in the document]
digicosmos86: This actually is done on purpose. `pt.switch` can cause some weird errors

src/hssm/likelihoods/analytical.py
```diff
@@ -34,9 +33,11 @@ def k_small(rt: np.ndarray, err: float) -> np.ndarray:
...
         A 1D at array of k_small.
     """
-    ks = 2 + pt.sqrt(-2 * rt * pt.log(2 * np.sqrt(2 * np.pi * rt) * err))
-    ks = pt.max(pt.stack([ks, pt.sqrt(rt) + 1]), axis=0)
-    ks = pt.switch(2 * pt.sqrt(2 * np.pi * rt) * err < 1, ks, 2)
+    _a = 2 * pt.sqrt(2 * np.pi * rt) * err < 1
+    _b = 2 + pt.sqrt(-2 * rt * pt.log(2 * pt.sqrt(2 * np.pi * rt) * err))
+    _c = pt.sqrt(rt) + 1
+    _d = pt.max(pt.stack([_b, _c]), axis=0)
+    ks = _a * _d + (1 - _a) * 2
```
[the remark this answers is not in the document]
digicosmos86: Please see comment below

src/hssm/likelihoods/analytical.py
```diff
@@ -56,9 +57,11 @@ def k_large(rt: np.ndarray, err: float) -> np.ndarray:
...
         A 1D at array of k_large.
     """
-    kl = pt.sqrt(-2 * pt.log(np.pi * rt * err) / (np.pi**2 * rt))
-    kl = pt.max(pt.stack([kl, 1.0 / (np.pi * pt.sqrt(rt))]), axis=0)
-    kl = pt.switch(np.pi * rt * err < 1, kl, 1.0 / (np.pi * pt.sqrt(rt)))
+    _a = np.pi * rt * err < 1
+    _b = 1.0 / (np.pi * pt.sqrt(rt))
+    _c = pt.sqrt(-2 * pt.log(np.pi * rt * err) / (np.pi**2 * rt))
+    _d = pt.max(pt.stack([_b, _c]), axis=0)
+    kl = _a * _b + (1 - _a) * _b
```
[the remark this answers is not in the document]
digicosmos86: Please see comment below

src/hssm/likelihoods/analytical.py
```diff
@@ -34,9 +33,11 @@ def k_small(rt: np.ndarray, err: float) -> np.ndarray:
     np.ndarray
         A 1D at array of k_small.
     """
-    ks = 2 + pt.sqrt(-2 * rt * pt.log(2 * np.sqrt(2 * np.pi * rt) * err))
-    ks = pt.max(pt.stack([ks, pt.sqrt(rt) + 1]), axis=0)
-    ks = pt.switch(2 * pt.sqrt(2 * np.pi * rt) * err < 1, ks, 2)
+    _a = 2 * pt.sqrt(2 * np.pi * rt) * err < 1
```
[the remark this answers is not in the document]
digicosmos86: It's actually equivalent but I was just playing around

src/hssm/likelihoods/analytical.py
```diff
@@ -34,9 +33,11 @@ def k_small(rt: np.ndarray, err: float) -> np.ndarray:
...
+    _b = 2 + pt.sqrt(-2 * rt * pt.log(2 * pt.sqrt(2 * np.pi * rt) * err))
+    _c = pt.sqrt(rt) + 1
```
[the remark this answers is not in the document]
digicosmos86: Sure! Feel free to change this

src/hssm/likelihoods/analytical.py
```diff
@@ -262,15 +241,17 @@ def logp_ddm(
...
 
-    p = pt.maximum(ftt01w(rt, a, z_flipped, err, k_terms), pt.exp(LOGP_LB))
+    negative_rt = rt <= epsilon
 
-    logp = pt.where(
-        rt <= epsilon,
-        LOGP_LB,
+    tt = negative_rt * epsilon + (1 - negative_rt) * rt
+
+    p = pt.maximum(ftt01w(tt, a, z_flipped, err, k_terms), pt.exp(LOGP_LB))
```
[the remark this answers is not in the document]
digicosmos86: Just playing around here. Not actually changing

src/hssm/likelihoods/analytical.py
digicosmos86: But this did help a bit, for some reason...

src/hssm/likelihoods/analytical.py
digicosmos86: Same as above

src/hssm/likelihoods/analytical.py
digicosmos86: Was playing around. It seems that changing `k_terms` to 7 did not improve speed or computation

digicosmos86:
@cpaniaguam Thanks for the suggestions! I committed all excluding those involving `pt.switch`. I thought `pt.switch` is a more readable alternative, but it caused some `switch_sink` errors that shows up only when `float64` is used. Actually removing `switch` Ops allowed me to sample with `float64` with out errors.

Please feel free to take this further. This PR wasn't final - was just a placeholder for some of my experiments

AlexanderFengler:
@digicosmos86 let's use this PR to switch to float64 overall? 

Also, the latest state of affairs with all changes in this PR is that it's still breaking right?

digicosmos86:
You are correct. It is still broken. This PR is kind of my mess though. I'd rather start a new one and just switch out all the `switch` ops, which should get us over the float64 issue

AlexanderFengler:
@digicosmos86 I am good with that approach.

digicosmos86:
Since this is still in the works, I am going to convert it to a draft PR

AlexanderFengler:
@digicosmos86 to be closed now that the other PR is up?

PR #478, opened 2024-06-30
fix bug where resampling doesnt update the traces argument

AlexanderFengler:
Core of the bug were two things:
- calling `az.extend()` by default prioritizes the "previous" idata wherever overlap...
- when returning the `.traces` via `.sample()` the results were overwritten if `.sample()` was called again --> now returning deepcopy instead


AlexanderFengler:
This also downgrades pymc version to `<5.15` for now.
Some refactoring needs to happen to accommodate the changes in `pymc>=5.15`.

PR #484, opened 2024-07-08
Fix compatibility with PyMC>=5.16

digicosmos86:
In PyMC>=5.16, which depends on pytensor>=2.23.0, the way to define `RandomVariable` class has changed to use a `signature` class property, which requires changes in HSSM codebase. In this PR, the following was changed:

1. removed `ndim_supp` which is no longer supported in pytensor
2. changed some of the ways `extra_fields` is passed due to this change. `extra_fields`, once defined, can no longer be changed (we will remove support for `extra_fields` for a different class for RL support in the future. This is just a temporary fix)
3. changed how missing data and deadline is handled
4. Also implemented a temporary fix to graphing


AlexanderFengler (approved):
Looks good!

PR #485, opened 2024-07-08
Reconciliate re-sampling behavior

### LINKED ISSUE #481: Reconciliate re-sampling behavior
**Describe the bug**

Currently, HSSM keeps a reference to the result of the last sampling. In #476, a bug was found that this internal `InferenceData` object does not get updated when `InferenceData` object already exists. This issue clarifies HSSM's behavior once the users resample:

**Expected behavior**

1. HSSM still keeps reference to the result of the last call to `model.sample()`. All results of the previous samples will be overwritten.
2. However, if there is an existing reference to a previous result, that reference will still be pointed to the result when the reference is created:

```python
trace1 = model.sample(...) #sample1
model.sample(...) #sample2
```

We expect that `trace1` will still point to the first sample. Implicitly, if the user does not assign the result of the call to `model.sample()`, then the result of the previous call will be overwritten, and the `InferenceData` will be garbage collected.

If there is a new reference to the second `model.sample()` call, then we expect the samples to be different.

```python
trace1 = model.sample(...)
trace2 = model.sample(...)

trace1 is trace2 #False
```

3. Any changes to the `InferenceData` object will directly change the object directly attached to the model (default Python behavior)

```python
trace1 = model.sample(...)
trace2 = model.traces

trace1 is trace2 #True

del trace2["posterior_predictive"]
```

Then `trace1` should not no longer have `posterior_predictive` group.

digicosmos86:
Resolves #481. Please see the issue for details. We have written tests to ensure the behavior described in #481

### DIFF
--- a/tests/test_hssm.py
+++ b/tests/test_hssm.py
@@ -267,3 +267,14 @@ def test_override_default_link(caplog, data_ddm_reg):
 
     assert "t" in caplog.records[0].message
     assert "strange" in caplog.records[0].message
+
+
+def test_resampling(data_ddm):
+    model = HSSM(data=data_ddm)
+    sample_1 = model.sample(draws=10, chains=1, tune=0)
+    assert sample_1 is model.traces
+
+    sample_2 = model.sample(draws=10, chains=1, tune=0)
+    assert sample_2 is model.traces
+
+    assert sample_1 is not sample_2
--- a/src/hssm/hssm.py
+++ b/src/hssm/hssm.py
@@ -537,18 +537,16 @@ class HSSM:
         # If sampler is finally `numpyro` make sure
         # the jitter argument is set to False
         if sampler == "nuts_numpyro":
-            if "jitter" not in kwargs.keys():
-                kwargs["jitter"] = False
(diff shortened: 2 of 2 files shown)


tests/test_hssm.py
```diff
@@ -267,3 +267,14 @@ def test_override_default_link(caplog, data_ddm_reg):
 
     assert "t" in caplog.records[0].message
     assert "strange" in caplog.records[0].message
+
+
+def test_resampling(data_ddm):
```
AlexanderFengler, line 272: we should put a not somewhere concerning the resulting behavior.
If I understand correctly, if you change sample_2 you automatically change model.traces?

AlexanderFengler (approved):
This is good, just see one comment.

tests/test_hssm.py
digicosmos86: I updated the `return` section of the `sample()` method and `traces` object to indicate that `.sample()` returns a reference to `model.traces` object, which stores the traces of the last call to `.sample()`.

PR #513, opened 2024-07-21
Fix categorical variables bug

AlexanderFengler:
Resolve inconsistency in how `jitter` was applied to initial values.
Jitter is now applied on the "unstransformed" variables.


digicosmos86 (approved):
Looks good to me!

PR #516, opened 2024-07-24
Fix compatibility with Bambi

digicosmos86:
We will have a series of PRs that fix various compatibility issues after the bambi 0.14.0 updates. We will merge all these PRs to this one before we merge it into main.


AlexanderFengler (approved):
Let's go :)

PR #517, opened 2024-07-24
Fix aliasing for bambi

digicosmos86:
In bambi 0.14.0, the naming convention for the parent parameter changed from `"{response}_mean"` to `"{parent_name}"`. We need to update how we alias the parameter names. Specifically, we need to:

**If the parent parameter is not a regression:**

1. Map `"{parent_name}"` to `"{parent_name}_mean"`
2. Map `"Intercept"` to `"{parent_name}"`

**If the parent parameter is a regression:**

We only need to add `"{parent_name}_"` prefix to every single covariate of the regression on parent.

PR #518, opened 2024-07-24
Fix posterior predictive sampling

digicosmos86:
In Bambi 0.14.0, the `kind` argument for `model.predict()` went from `"pps"` and `"mean"` to `"response"` and `"response_params"`. We need to update bambi code to accommodate for that

### DIFF
--- a/src/hssm/hssm.py
+++ b/src/hssm/hssm.py
@@ -404,7 +404,8 @@ class HSSM:
             self.model, self._parent_param, self.response_c, self.response_str
         )
         self.set_alias(self._aliases)
-        # _logger.info(self.pymc_model.initial_point())
+        self.model.build()
+        _logger.info(self.pymc_model.initial_point())
 
         if process_initvals:
             self._postprocess_initvals_deterministic(initval_settings=INITVAL_SETTINGS)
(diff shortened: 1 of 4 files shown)


src/hssm/utils.py
```diff
@@ -113,35 +113,35 @@ def _get_alias_dict(
...
-                            alias_dict[response_c] |= {name: f"{parent_name}_{name}"}
-            else:
-                alias_dict[component_name] = {component_name: component_name}
+                    alias_dict[name] = f"{parent_name}_{name}"
 
-    for name in model.constant_components.keys():
-        alias_dict |= {name: name}
+        return alias_dict
+
+    for component_name, component in model.distributional_components.items():
```
AlexanderFengler, line 132: I think I just saw these changes in #517 as well.

src/hssm/hssm.py
```diff
@@ -404,7 +404,8 @@ def __init__(
             self.model, self._parent_param, self.response_c, self.response_str
         )
         self.set_alias(self._aliases)
-        # _logger.info(self.pymc_model.initial_point())
+        self.model.build()
+        _logger.info(self.pymc_model.initial_point())
```
AlexanderFengler, line 408: saw this change in #517 not sure what the right spot for comment is.

AlexanderFengler (commented):
Looks good, apart from comment passed through from #517

PR #519, opened 2024-07-24
Fix plotting

digicosmos86:
In Bambi 0.14.0, the dimension for response has changed from `"{response}_obs"` to `"__obs__"`. We need to change this in our code and also the test files to ensure that the plotting tests pass


AlexanderFengler (commented):
Looks good, at this point I guess the idea is to stack the changes across PR numbers?

So it's going to be held back by comments to #517 ?

PR #520, opened 2024-07-24
Fix sampler specification

digicosmos86:
In Bambi 0.14.0, all JAX-based sampling is now done via `bayeux`, which breaks our sampler specification. For the moment, we are sticking to JAX-based sampling via pymc, through the `nuts_sampler` argument of `pm.sample`


src/hssm/hssm.py
```diff
@@ -538,35 +544,39 @@ def sample(
...
-            kwargs["jitter"] = False
-        else:
-            if "jitter" in kwargs:
-                _logger.warning(
-                    "The jitter keyword argument is "
-                    + "supported only by the nuts_numpyro sampler. \n"
-                    + "The jitter argument will be ignored."
-                )
-                del kwargs["jitter"]
+            if "nuts_numpyro_kwargs" in kwargs:
```
AlexanderFengler, line 547: `nuts_sampler_kwargs` ?

AlexanderFengler (approved):
this one (looking at the incremental change beyond #519) looks good.

src/hssm/hssm.py
digicosmos86: Yep need to fix this one

PR #521, opened 2024-07-24
Fix graphing

digicosmos86:
Many changes to the graphing facilities were introduced in PyMC 5.16.x, and HSSM's internal graphing hacks stopped working. Fortunately, in Bambi 0.14.0, we no longer need these hacks. Instead, we simply need to implement some modifications to the graph generated from Bambi in the case where `parent` is not a regression.


src/hssm/utils.py
```diff
@@ -113,159 +111,37 @@ def _get_alias_dict(
...
-    """
-    return function([], var, mode="FAST_COMPILE")()
+                    alias_dict[name] = f"{parent_name}_{name}"
 
+        return alias_dict
 
-VarName = NewType("VarName", str)
-
-
-class HSSMModelGraph(ModelGraph):
```
AlexanderFengler, line 162: really nice that we can grop this!

AlexanderFengler (approved):

PR #522, opened 2024-07-24
Fix InferenceData

digicosmos86:
For some reason, `f"{model._parent}_mean"` is included in InferenceData but this wasn't the case before aliasing. Two related issues for bambi: https://github.com/bambinos/bambi/issues/828 and https://github.com/bambinos/bambi/issues/827 that can potentially fix this problem with HSSM

Also modified some tests so that tests pass


AlexanderFengler (approved):
Looks good!

PR #534, opened 2024-08-03
Bambi fix shape issues

AlexanderFengler:
fixes failing tests in #522 .


src/hssm/hssm.py
```diff
@@ -975,27 +975,39 @@ def plot_trace(
...
-                        raise ValueError(
-                            "`var_names` must be a string, a list of strings, or None."
-                        )
-                else:
-                    kwargs["var_names"] = var_names
+                        kwargs["var_names"] = var_names
+        elif (not isinstance(data, az.InferenceData)) and (data is not None):
+            raise ValueError("data must be an InferenceData object.")
+        elif data is None:
+            raise ValueError("Please sample to model first.")
```
digicosmos86, line 1010: For clarity, I'd use just one if block. We already have an assert statement to ensure that `data` is of type `az.InferenceData`, so just an else block that throws an error should be sufficient

digicosmos86 (approved):
Looks great! Just one minor issue with if statements lol

digicosmos86:
> Looks great! Just one minor issue with if statements lol

@AlexanderFengler also I am changing the merge target to the previous PR (fix InferenceData). We need to merge this one first before merging that one

src/hssm/hssm.py
AlexanderFengler: Ok thinking about the logic, I think we were able to just take the whole if logic out and rely on the fact that `data` must be `InferenceData` at this point anyways. @digicosmos86

AlexanderFengler: This needed to change in two places.
Will let the tests run now, then merge?

PR #545, opened 2024-08-13
537 initial vi support and tutorial

AlexanderFengler:
Adds:
- functionality for variational inference native to HSSM class
- basic tests for new functionality mirroring the MCMC counterpart
- tutorial on variational inference illustrating both the route via HSSM and the route via PyMC directly


src/hssm/hssm.py
```diff
@@ -598,8 +609,97 @@ def sample(
...
+
+        # Get posterior samples from vi-approximation
+        if self._vi_approx is not None:
+            self._inference_obj_vi = self._vi_approx.sample(draws)
+        else:
+            raise ValueError(
+                "VI approximation object has not been initialized yet."
+                " However this check happens after VI is supposed to have run,"
+                " so something is wrong!"
+            )
```
digicosmos86, line 679: This check is unnecessary, no? We know that `pm.fit()` would never return `None` without throwing an error

src/hssm/hssm.py
```diff
@@ -598,8 +609,97 @@ def sample(
...
+
+        # Post-processing
+        if hasattr(self, "pymc_model") and self._inference_obj_vi is not None:
+            vars_to_keep = set(
+                [var.name for var in self.pymc_model.free_RVs]
+            ).intersection(
+                set(list(self._inference_obj_vi["posterior"].data_vars.keys()))
+            )
+        else:
+            raise ValueError("Model has not been initialized yet. Something is wrong!")
```
digicosmos86, line 689: This could be nick-picking, but I never liked the error message "Something is wrong." We should either provide something more informative or leave it out

src/hssm/hssm.py
```diff
@@ -598,8 +609,97 @@ def sample(
...
+        niter: int = 10000,
+        draws=1000,
+        return_idata=True,
+        **vi_kwargs,
+    ) -> pm.Approximation | az.InferenceData:
+        """Perform Variational Inference.
+
+        Parameters
+        ----------
+        niter : int
```
digicosmos86, line 640: The docstring format is `numpy-notype`. Since we already use type annotation, we don't need to provide the types for the variables here. `mkdocs` can figure it out.

src/hssm/hssm.py
```diff
@@ -598,8 +609,97 @@ def sample(
...
+        )
+
         return self.traces
 
+    def vi(
+        self,
+        method: str = "advi",
+        niter: int = 10000,
+        draws=1000,
+        return_idata=True,
```
digicosmos86, line 633: Types are missing here

src/hssm/hssm.py
```diff
@@ -598,8 +609,97 @@ def sample(
...
+            ).intersection(set(list(self._inference_obj["posterior"].data_vars.keys())))
+        else:
+            raise ValueError("Model has not been initialized yet. Something is wrong!")
+
+        setattr(
+            self._inference_obj,
+            "posterior",
+            self._inference_obj["posterior"][list(vars_to_keep)],
+        )
+
```
digicosmos86, line 625: Since we have a separate `.vi()` method here, this function will only return an `InferenceData` type. Let's change the type annotation for the function signature and docstrings as well

digicosmos86 (approved):
Looks good to me! Just documentation changes here and there. Once the tests pass it's safe to merge.

PR #547, opened 2024-08-14
Fix initval setters leading to NotImplementedError

AlexanderFengler:
This changes the logic on how we store `initvals`. 
We now attach and update a class property `.initvals` that holds a dictionary.
This dictionary is passed as `initvals` argument to the `.sampler()` method.

All pre-processing of initial values now targets this dictionary, instead of attempting changes to the `pymc_model` directly.


src/hssm/hssm.py
```diff
@@ -629,8 +633,9 @@ def vi(
         self,
         method: str = "advi",
         niter: int = 10000,
-        draws=1000,
-        return_idata=True,
+        draws: int = 1000,
+        return_idata: bool = True,
+        ignore_mcmc_start_point_defaults=False,
```
digicosmos86, line 638: Type for this parameter

src/hssm/hssm.py
```diff
@@ -664,19 +669,17 @@ def vi(
...
+        if ("start" not in vi_kwargs) and not ignore_mcmc_start_point_defaults:
+            _logger.info("Using MCMC starting point defaults.")
+            vi_kwargs["start"] = self._initvals
+
         # Run variational inference directly from pymc model
         with self.pymc_model:
             self._vi_approx = pm.fit(n=niter, method=method, **vi_kwargs)
 
         # Get posterior samples from vi-approximation
         if self._vi_approx is not None:
```
digicosmos86, line 681: This check is still redundant

src/hssm/hssm.py
```diff
@@ -2048,9 +2075,12 @@ def __jitter_initvals_all(self, jitter_epsilon: float) -> None:
             starting_value_tmp = starting_value + np.random.uniform(
                 -jitter_epsilon, jitter_epsilon, starting_value.shape
             ).astype(np.float32)
-            self.pymc_model.set_initval(
-                self.pymc_model.named_vars[name_tmp], starting_value_tmp
-            )
+
+            if not isinstance(self._initvals, dict):
+                raise ValueError("self._initvals should not be a dictionary.")
+            else:
```
digicosmos86, line 2081: `self._initvals` can only be of type `dict` right? It's initialized as a `dict`

AlexanderFengler: most of these are just in response to mypy bugging me endlessly :D
will try to do a round of improvements. @digicosmos86

PR #568, opened 2024-08-21
Fix float32 vs float64 issue

### LINKED ISSUE #156: Fix `float32` vs `float64` issue
Using analytical `ddm` likelihood while not forcing `pytensor.config.floatX = "float32"` will result in warning messages while sampling still continues and finishes successfully. Posted on PyMC Discourse and confirmed that this is a bug on PyMC which still exists in the latest v5.6.0.

@AisOmar While we wait for their permanent fix, we should probably add `pytensor.config.floatX = "float32"` in the getting started code for the time being just to avoid these confusing warnings.

digicosmos86:
closes #468 
closes #156 

@AlexanderFengler @cpaniaguam After working on this PR, I realized that this is not very different from #468, but I have only incorporated changes that we agreed on. @cpaniaguam it seems that the code did not quite work after merging your suggestions in #468, so I didn't incorporate all of them. Can you take a look again, and if there's anything missing, let me know?

### DIFF
--- a/src/hssm/likelihoods/analytical.py
+++ b/src/hssm/likelihoods/analytical.py
@@ -25,7 +25,7 @@ def k_small(rt: np.ndarray, err: float) -> np.ndarray:
     Parameters
     ----------
     rt
-        A 1D numpy array of flipped R.... T.....s. (0, inf).
+        A 1D numpy array of flipped RTs. (0, inf).
     err
         Error bound.
 
@@ -34,9 +34,11 @@ def k_small(rt: np.ndarray, err: float) -> np.ndarray:
     np.ndarray
         A 1D at array of k_small.
     """
-    ks = 2 + pt.sqrt(-2 * rt * pt.log(2 * np.sqrt(2 * np.pi * rt) * err))
-    ks = pt.max(pt.stack([ks, pt.sqrt(rt) + 1]), axis=0)
-    ks = pt.switch(2 * pt.sqrt(2 * np.pi * rt) * err < 1, ks, 2)
+    _a = 2 * pt.sqrt(2 * np.pi * rt) * err < 1
+    _b = 2 + pt.sqrt(-2 * rt * pt.log(2 * pt.sqrt(2 * np.pi * rt) * err))
+    _c = pt.sqrt(rt) + 1
+    _d = pt.maximum(_b, _c)
+    ks = _a * _d + (1 - _a) * 2
 
     return ks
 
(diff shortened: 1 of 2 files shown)


tests/test_likelihoods.py
```diff
@@ -116,12 +92,51 @@ def test_bbox(data_ddm, logp_func, logp_bbox_func, true_values):
...
+    z = pt.dvector()
+    t = pt.dvector()
+    sv = pt.dvector()
+    size = cav_data_numpy.shape[0]
+    logp = logp_ddm(cav_data_numpy, v, a, z, t).sum()
+    grad = pt.grad(logp, wrt=[v, a, z, t])
+    grad_func = pytensor.function(
+        [v, a, z, t],
+        grad,
+        mode=NanGuardMode(nan_is_error=True, inf_is_error=True, big_is_error=False),
```
cpaniaguam, line 114: Define mode once and reuse it twice?
```python
    mode = NanGuardMode(nan_is_error=True, inf_is_error=True, big_is_error=False)
```

src/hssm/likelihoods/analytical.py
```diff
@@ -34,9 +34,11 @@ def k_small(rt: np.ndarray, err: float) -> np.ndarray:
     np.ndarray
         A 1D at array of k_small.
     """
-    ks = 2 + pt.sqrt(-2 * rt * pt.log(2 * np.sqrt(2 * np.pi * rt) * err))
-    ks = pt.max(pt.stack([ks, pt.sqrt(rt) + 1]), axis=0)
-    ks = pt.switch(2 * pt.sqrt(2 * np.pi * rt) * err < 1, ks, 2)
+    _a = 2 * pt.sqrt(2 * np.pi * rt) * err < 1
```
AlexanderFengler, line 37: I guess if you want to go really crazy, you can define `__a=2 * np.pi * rt` only once :)

src/hssm/likelihoods/analytical.py
```diff
@@ -56,9 +58,11 @@ def k_large(rt: np.ndarray, err: float) -> np.ndarray:
     np.ndarray
         A 1D at array of k_large.
     """
-    kl = pt.sqrt(-2 * pt.log(np.pi * rt * err) / (np.pi**2 * rt))
-    kl = pt.max(pt.stack([kl, 1.0 / (np.pi * pt.sqrt(rt))]), axis=0)
-    kl = pt.switch(np.pi * rt * err < 1, kl, 1.0 / (np.pi * pt.sqrt(rt)))
+    _a = np.pi * rt * err < 1
```
AlexanderFengler, line 61: same here with `np.pi * rt * err`

AlexanderFengler (approved):
Is the behavior actually a bit more robust now?
Otherwise looks good, left only a minor comments.

PR #572, opened 2024-08-22
Added `HSSM.restore_traces()` method

digicosmos86:
The method will either attach an existing `az.InferenceData` or `pm.Approximation` object to `self._inference_obj` or load from a netcdf and attach the object


AlexanderFengler (approved):
Loogs good.

digicosmos86:
@AlexanderFengler Looks like some tests ran for 6 hours. Can you take a look? Looks like one of those tests for vi

PR #576, opened 2024-08-23
Fixed error of parent parameter specification

digicosmos86:
@AlexanderFengler after this change is made `process_initvals` breaks. Can you take a look?


digicosmos86:
@AlexanderFengler we should merge this so we have a semi-working `main`. Tests in CI not passing has nothing to do with the changes introduced in this PR

PR #579, opened 2024-08-30
Added UserParam class for user parameter specification

digicosmos86:
Note: To avoid conflict with `src/jhssm/param.py`, I am naming the directory `src/hssm/param_refactor` for now. Once all integration is tested, I will rename this directory

What was done in this PR:

1. Added a few methods in `prior.py` to prepare for model output
2. Added a UserParam (which will eventually be exported as `hssm.Param`) for users to specify priors and store user specification


AlexanderFengler (approved):
Prob warrants a comprehensive tutorial on prior setting?
Looks good otherwise.

tests/test_prior.py
cpaniaguam, line 1: Why not use the truthy/falsy nature of the values in many of these checks? Makes for more consice code.

cpaniaguam (changes requested):
Looks good. I pointed out a couple of things I would change but might not be needed perhaps due to conventions that I'm not aware of yet.

AlexanderFengler:
@digicosmos86 is this on hold or pushed through #590 at this point?

AlexanderFengler:
From what I can see this is redundant now.

PR #583, opened 2024-09-06
Revert changes to analytical likelihood function

digicosmos86:


AlexanderFengler (approved):
looks good, tested that speed is as it should be. 
let's merge this for now and continue eventual investigation elsewhere.

Issue #432, opened 2024-05-14
Suggestion for Improving Prior Setup in Hierarchical Models


gingjehli:
It would be helpful if one could specify priors for classes such as intercepts, coefficients, and covariance matrices across all hierarchical levels of a model (e.g., groups, treatments, or subjects). This way, one can ensure that the chosen priors are applied consistently throughout the model, which makes results more reliable.

Also, it would be great to incorporate warning messages if one hasn't set priors for every parameter in a provided regression equation. A warning message could pop up to remind that some priors are missing, and the software is going to use the default ones instead. This feature would prevent from accidentally overlooking a setting and help avoid the issues that come with unintended default priors.

Adding these features would make the modeling process smoother and more user-friendly, giving better control and clarity over how our models are built.

AlexanderFengler:
Hey @gingjehli , 

we will look into this. Adding the warnings you suggest is trivial, we can definitely do that. 
Priors for parameter-groups will demand a bit more thought, but is doable as well. Thank you for raising those points.

digicosmos86:
This is done. There is a `global_formula` argument to `HSSM` that allows you to globally set a formula

PR #590, opened 2024-10-04
[Refactor] How parameters are represented in HSSM

### LINKED ISSUE #431: Custom Group-Level Priors Not Propagating to Individual-Level Priors When Hierarchical=True
When the 'hierarchical=True' option is set and custom priors are specified, the group-level priors do not automatically apply to the individual-level priors. Instead, individual-level priors continue to default to their predefined settings. Would be good to fix this a/o mention it in the documentation. For now, it seems that all priors for all levels need to be set individually (AND hierarchical has to be set to False) if one wants to fully leverage the use of self-specified priors.

Here is an example: 

`
sm1_full2 = hssm.HSSM(
    data=data,
    model="angle",
    hierarchical=True, 
    prior_settings=None, 
    loglik_kind="approx_differentiable",
    include=[
        {
            "name": "v",
            "formula": "v ~ 1 + (sub_cond_rewardLOG + sub_cond_averse)*conflictDomain*condition2 + (1 + (sub_cond_rewardLOG + sub_cond_averse)*conflictDomain*condition2 | participant_id)", 
            "prior": {
                "Intercept": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_rewardLOG": {"name": "Normal", "mu": 0.0, "sigma": 2.5},  # Using a "large" prior for covariates
                "sub_cond_averse": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_rewardLOG:conflictDomain": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_averse:conflictDomain": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_rewardLOG:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_averse:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_rewardLOG:conflictDomain:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_averse:conflictDomain:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "conflictDomain:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "conflictDomain": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
            },
            "link": "identity",
            "bounds": (-10.0, 10.0)
        },],
)
sm1_full2 
`

**Below is a snippet of the corresponding output - the participant-level priors are still the default priors. This is misleading:** 

Formula: v ~ 1 + (sub_cond_rewardLOG + sub_cond_averse)*conflictDomain*condition2 + (1 + (sub_cond_rewardLOG + sub_cond_averse)*conflictDomain*condition2 | participant_id)
    Priors:
        v_Intercept ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_rewardLOG ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_averse ~ Normal(mu: 0.0, sigma: 2.5)
        v_conflictDomain ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_rewardLOG:conflictDomain ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_averse:conflictDomain ~ Normal(mu: 0.0, sigma: 2.5)
        v_condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_rewardLOG:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_averse:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_conflictDomain:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_rewardLOG:conflictDomain:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_averse:conflictDomain:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_1|participant_id ~ Normal(mu: 0.0, sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_sub_cond_rewardLOG|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_sub_cond_averse|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_conflictDomain|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_sub_cond_rewardLOG:conflictDomain|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_sub_cond_averse:conflictDomain|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_condition2|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_sub_cond_rewardLOG:condition2|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_sub_cond_averse:condition2|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_conflictDomain:condition2|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_sub_cond_rewardLOG:conflictDomain:condition2|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
        v_sub_cond_averse:conflictDomain:condition2|participant_id ~ Normal(mu: Normal(mu: 0.0, sigma: 0.25), sigma: Weibull(alpha: 1.5, beta: 0.30000001192092896))
    Link: identity
    Explicit bounds: (-10.0, 10.0)

**So, to fully propagate the self-defined priors, 'hierarchical=True' had to be set to false as shown below:**

`
sm1_full1 = hssm.HSSM(
    data=data,
    model="angle",
    hierarchical=False, 
    prior_settings=None, 
    loglik_kind="approx_differentiable",
    include=[
        {
            "name": "v",
            "formula": "v ~ 1 + (sub_cond_rewardLOG + sub_cond_averse)*conflictDomain*condition2 + (1 + (sub_cond_rewardLOG + sub_cond_averse)*conflictDomain*condition2 | participant_id)",
            "prior": {          
                "Intercept": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_rewardLOG": {"name": "Normal", "mu": 0.0, "sigma": 2.5},  # Using a "large" prior for covariates
                "sub_cond_averse": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_rewardLOG:conflictDomain": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_averse:conflictDomain": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_rewardLOG:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_averse:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_rewardLOG:conflictDomain:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "sub_cond_averse:conflictDomain:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "conflictDomain:condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "conflictDomain": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
                "condition2": {"name": "Normal", "mu": 0.0, "sigma": 2.5},
            },
            "link": "identity",
            "bounds": (-10.0, 10.0)
        },],
)
sm1_full1 
`

**In this case, the priors are now correctly propagated to the participant layer :)**

Here an output snippet:

v:
    Formula: v ~ 1 + (sub_cond_rewardLOG + sub_cond_averse)*conflictDomain*condition2 + (1 + (sub_cond_rewardLOG + sub_cond_averse)*conflictDomain*condition2 | participant_id)
    Priors:
        v_Intercept ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_rewardLOG ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_averse ~ Normal(mu: 0.0, sigma: 2.5)
        v_conflictDomain ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_rewardLOG:conflictDomain ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_averse:conflictDomain ~ Normal(mu: 0.0, sigma: 2.5)
        v_condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_rewardLOG:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_averse:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_conflictDomain:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_rewardLOG:conflictDomain:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_sub_cond_averse:conflictDomain:condition2 ~ Normal(mu: 0.0, sigma: 2.5)
        v_1|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 2.5))
        v_sub_cond_rewardLOG|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 2.5373001098632812))
        v_sub_cond_averse|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 2.486599922180176))
        v_conflictDomain|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 5.008999824523926))
        v_sub_cond_rewardLOG:conflictDomain|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 4.127600193023682))
        v_sub_cond_averse:conflictDomain|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 3.574899911880493))
        v_condition2|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 5.0))
        v_sub_cond_rewardLOG:condition2|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 3.5920000076293945))
        v_sub_cond_averse:condition2|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 3.4658000469207764))
        v_conflictDomain:condition2|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 5.855800151824951))
        v_sub_cond_rewardLOG:conflictDomain:condition2|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 5.573299884796143))
        v_sub_cond_averse:conflictDomain:condition2|participant_id ~ Normal(mu: 0.0, sigma: HalfNormal(sigma: 4.914400100708008))
    Link: identity
    Explicit bounds: (-10.0, 10.0)

**Though, you can see that the scaling of the sigmas at the individual level isn't done automatically. 
Would be good to fix this and/or mention all of the above in the documentation).**

### LINKED ISSUE #559: [Refactor] `Params` class, which wraps around a dictionary and stores all ParamDef information
This class also handles the following:

1. Determines and keeps track of which Param to use as parent
2. Iterates over all params and produces Bambi definitions of `formula` and `priors`
3. Produces printouts for all parameters
4. Produces user specifications in all parameters for model saving/loading

### LINKED ISSUE #560: [Refactor] `Param` class, which is a base class for `SimpleParam` and `RegressionParam` class

### LINKED ISSUE #561: [Refactor] `SimpleParam` class

### LINKED ISSUE #562: [Refactor] RegressionParamDef class

### LINKED ISSUE #564: [Refactor] `UserParam` class only for user class specification

digicosmos86:
This PR introduces a refactored `Param` class to represent parameter specifications in HSSM.

Before the refactor, `Param` was a single class that both accepts user inputs and performs internal processing, such as merging with default values and truncating priors, which made the code in the class very messy. Some of the functionalities also bled into the HSSM class, which bloated the class as well. The refactor aims to separate the concerns of the old `Param` class into multiple classes, each taking one responsibility of the Param class. This completely moves the concerns out of the HSSM class, making the main class much less bloated (even though there is still room for improvement).

The improvements are:
1. A user-facing `UserParam` class, which is exported as `hssm.Param`, for storing user input. (closes #564)
2. An internal `Param` class that defines some common functionalities for the parameter representation. (closes #560)
3. Inheriting the `Param` class is the `SimpleParam` class, which handles the situation where the parameter is not a regression. Its subclass `DefaultParam` handles the situation where the prior is not specified in the non-regression case (where we do not truncate the prior). (closes #561)
4. The `RegressionParam` class for handling the regression case (closes #562)
5. A dict-like `Params` class for creating these params both from user specification, or defaults, or both. It also takes away some of the responsibilities of the HSSM class (closes #559)

Tests for all these new classes are under `tests/param` directory.

The refactor also removes "hierarchical" argument to `HSSM`, adding `global_formula` instead. (closes #431, #432)

Some other changes to other code files are necessary for this gigantic refactor to work, but most of the changes are under `src/hssm/param`.


src/hssm/param/utils.py
```diff
@@ -0,0 +1,43 @@
+"""Utility functions for the parameter classes."""
+
+import bambi as bmb
+import numpy as np
+
+
+def validate_bounds(bounds: tuple[float, float]) -> None:
+    """Validate the bounds."""
```
cpaniaguam, line 8: I think the docstring could be improved by adding some more context or maybe an example.
```suggestion
    """
    Validate the given bounds.

    This function checks that the `bounds` are a valid tuple containing two values,
    where the first value is less than the second also handling unbounded cases.

        Parameters
        ----------
        bounds (tuple[float, float]): A tuple containing two float values representing 
                                       the lower and upper bounds.

        Raises
        -------
        ValueError: If `bounds` does not contain exactly two values, or if the 
                    lower bound is greater than or equal to the upper bound.

        Examples:
        >>> validate_bounds((0.0, 1.0))  # Valid input, should not raise an exception

        >>> validate_bounds((1.0, 0.0))  # Invalid input, should raise ValueError
        Traceback (most recent call last):
            ...
        ValueError: Invalid bounds: (1.0, 0.0)

        >>> validate_bounds((0.0, 0.0))  # Invalid input, should raise ValueError
        Traceback (most recent call last):
            ...
        ValueError: Invalid bounds: (0.0, 0.0)

        >>> validate_bounds((-inf, inf))  # Valid input, should not raise an exception

        >>> validate_bounds((inf, -inf))  # Invalid input, should raise ValueError
        Traceback (most recent call last):
            ...
        ValueError: Invalid bounds: (inf, -inf)
    """
```

src/hssm/param/regression_param.py
```diff
@@ -0,0 +1,402 @@
...
+    def validate(self) -> None:
+        """Validate the parameter."""
+        if self.formula is None:
+            raise ValueError(f"Formula not specified for parameter {self.name}.")
+        self.reformat_formula()
+        if isinstance(self.prior, bmb.Prior):
+            raise ValueError(
+                "Please specify priors for each individual parameter in the "
+                + f"regression for {self.name}."
+            )
```
cpaniaguam, line 153:
```suggestion
            raise ValueError(
                "Please specify priors for each individual parameter in the "
                f"regression for {self.name}."
            )
```

Also consider using a template or function as this one has popped twice already.

src/hssm/param/regression_param.py
```diff
@@ -0,0 +1,402 @@
...
+        if np.isneginf(lower) and np.isposinf(upper):
+            self.link = "identity"
+        elif lower == 0.0 and np.isposinf(upper):
+            self.link = "log"
+        elif not np.isneginf(lower) and not np.isposinf(upper):
+            self.link = Link("gen_logit", bounds=self.bounds)
+        else:
+            _logger.warning(
+                "The bounds for parameter %s (%f, %f) seem strange. Nothing is done to"
+                + " the link function. Please check if they are correct.",
```
cpaniaguam, line 193:
```suggestion
                "The bounds for parameter %s (%f, %f) seem strange. Nothing is done to "
                "the link function. Please check if they are correct.",
```

src/hssm/param/regression_param.py
```diff
@@ -0,0 +1,402 @@
...
+
+        This is most likely because both default prior and default bounds are supplied.
+        """
+        if self.link is not None:
+            return
+
+        if self.bounds is None:
+            raise ValueError(
+                "Cannot override the default link function. "
+                f"Bounds are not specified for parameter {self.name}."
```
cpaniaguam, line 179: 👍

src/hssm/param/regression_param.py
```diff
@@ -0,0 +1,402 @@
...
+
+    def validate(self) -> None:
+        """Validate the parameter."""
+        if self.formula is None:
+            raise ValueError(f"Formula not specified for parameter {self.name}.")
+        self.reformat_formula()
+        if isinstance(self.prior, bmb.Prior):
+            raise ValueError(
+                "Please specify priors for each individual parameter in the "
+                + f"regression for {self.name}."
```
cpaniaguam, line 152: I think you can drop the `+` here.
```suggestion
                "Please specify priors for each individual parameter in the "
                f"regression for {self.name}."
```

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+    elif user_param.is_simple:
+        param = SimpleParam.from_user_param(user_param)
+    else:
+        # If we cannot tell definitively that the parameter is a regression parameter
+        # or a simple parameter from the UserParam object,
+        # we need to determine the type based on whether the model has a global formula
+        if model.global_formula is not None:
+            param = RegressionParam.from_user_param(user_param)
+        else:
+            param = SimpleParam.from_user_param(user_param)
```
cpaniaguam, line 311: Why not add the type hints here?

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+
+    if model.global_formula is not None:
+        param: Param = RegressionParam.from_defaults(
+            name,
+            formula=model.global_formula,
+            bounds=default_bounds,
+            link_settings=model.link_settings,
+        )
+    else:
+        param = DefaultParam.from_defaults(name, default_prior, default_bounds)
```
cpaniaguam, line 351: Type hint?

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+    is_ddm = (
+        model.model_name in ["ddm", "ddm_sdv", "full_ddm"]
+        and model.loglik_kind != "approx_differentiable"
+    )
+
+    for name in model.list_params:
+        if name in user_params:
+            param: Param = make_param_from_user_param(model, name, user_params[name])
+        else:
+            param = make_param_from_defaults(model, name)
```
cpaniaguam, line 267: Type hint?

cpaniaguam (changes requested):
Thanks for this redesign! Added some suggestions and a question about `default_model_config`.

src/hssm/hssm.py
```diff
@@ -386,24 +380,19 @@ def __init__(
         if self.has_lapse and self.list_params[-1] != "p_outlier":
             self.list_params.append("p_outlier")
 
-        # Process kwargs and p_outlier and add them to include
-        include, other_kwargs = self._add_kwargs_and_p_outlier_to_include(
-            include, kwargs, p_outlier
+        # Process all parameters
+        self.params = Params.from_user_specs(
+            model=self,
+            include=[] if include is None else include,
```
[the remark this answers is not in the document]
digicosmos86: We need to explicitly account for the case when `include` is an empty list, which would cause `not include` to evaluate to true

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+
+        self.parent_param = self[self.parent]
+
+    @classmethod
+    def from_user_specs(
+        cls,
+        model: HSSM,
+        include: list[dict[str, Any] | UserParam],
+        kwargs: dict[str, Any],
+        p_outlier: float | dict | bmb.Prior | None,
```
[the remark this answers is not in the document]
digicosmos86: I thought about this too, but can't think of anything better than `kwargs` , because it is indeed the kwargs argument in the constructor of `HSSM`

src/hssm/param/params.py
digicosmos86: It's actually implicitly the same type as the `Param` in the other branch. I had to have explicit type hint for the other branch because that's mypy will use for this branch as well, which causes a mypy error

src/hssm/param/params.py
digicosmos86: Again, the type mypy uses is when it first sees the `param` variable in the first branch. Without the type hint, mypy will use `RegressionParam` for `param`, which will cause errors. So once I declare `param` to be `Param`, that works for all branches

src/hssm/param/simple_param.py
```diff
@@ -0,0 +1,283 @@
...
+            else:
+                prior = f"Value: {self.prior}"
+        elif isinstance(self.prior, (dict, bmb.Prior)):
+            prior = f"Prior: {self.prior}"
+        else:
+            raise ValueError(f"Invalid prior type {type(self.prior)}")
+
+        if self.bounds is None:
+            return f"{self.name}:\n" f"    {prior}"
+        return f"{self.name}:\n" f"    {prior}\n" f"    Explicit bounds: {self.bounds}"
```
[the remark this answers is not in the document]
digicosmos86: Yep, but `\t` produces 8 spaces, and I specifically need 4 spaces :)

src/hssm/param/user_param.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def to_dict(
+        self,
+    ) -> dict:
+        """Convert the UserParam object to a dictionary with shallow copy."""
+        return {
+            f.name: getattr(self, f.name)
+            for f in fields(self)
+            if getattr(self, f.name) is not None
+        }
```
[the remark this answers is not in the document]
digicosmos86: `asdict` produces a deep copy, but for this case I only need a shallow copy

src/hssm/param/utils.py
digicosmos86: Sounds great!

tests/param/test_user_param.py
```diff
@@ -0,0 +1,73 @@
...
+    """Test that the to_dict method works correctly."""
+    param = UserParam(name="test", prior=0, formula="x", link="identity", bounds=(0, 1))
+    param_dict = param.to_dict()
+    assert param_dict == {
+        "name": "test",
+        "prior": 0,
+        "formula": "x",
+        "link": "identity",
+        "bounds": (0, 1),
+    }
```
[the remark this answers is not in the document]
digicosmos86: Great idea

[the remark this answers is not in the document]
digicosmos86: The next test actually tests for that

tests/param/test_params.py
```diff
@@ -0,0 +1,408 @@
...
+    {"name": "v", "prior": {"name": "Normal", "mu": 0, "sigma": 1}},
+    {"name": "a", "prior": bmb.Prior("HalfNormal", sigma=1)},
+]
+
+
+def kw():
+    return {
+        "z": UserParam(prior=bmb.Prior("Normal", mu=0, sigma=1)),
+        "other_param": 1,
+    }
```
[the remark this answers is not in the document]
digicosmos86: Tried that. I haven't figured out a way for a new dictionary to be created every time the fixture is used by simply changing the scope of the fixture. Because the object is modified in-place, that caused some chaos. If there's a way for a new dictionary to be created every time a fixture is used, please let me know, and I'll definitely use a fixture.

tests/param/test_params.py
```diff
@@ -0,0 +1,408 @@
...
+    assert reg_user_param.is_regression
+
+    param = make_param_from_user_param(model, "a", reg_user_param)
+    assert isinstance(param, RegressionParam)
+    assert param.name == "a"
+    assert param.formula == "a ~ x"
+    assert param.prior is None
+    assert (
+        param.bounds
+        == default_model_config["ddm"]["likelihoods"]["analytical"]["bounds"]["a"]
```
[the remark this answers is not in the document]
digicosmos86: It's meant to be a large dictionary, so that one day it could be re-written as a JSON or YAML file. We also thought about using shared configs between the `hssm` package and `ssm-simulators` package. As I mentioned, we definitely welcome ideas on how to implement this differently, including `lazy` ways to delay loading all this information until specifically requested

tests/param/test_params.py
```diff
@@ -0,0 +1,408 @@
...
+        has_lapse=True,
+        loglik_kind=loglik_kind,
+        global_formula=global_formula,
+        link_settings=link_settings,
+        prior_settings=prior_settings,
+        model_config=config,
+        response_c="c(rt, response)",
+        additional_namespace={},
+    )
+    return model
```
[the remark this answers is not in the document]
digicosmos86: Added on top of the function

tests/param/test_params.py
```diff
@@ -0,0 +1,408 @@
...
+    param = make_param_from_user_param(model, "z", ambiguous_user_param)
+    assert isinstance(param, RegressionParam)
+    assert param.name == "z"
+    assert param.formula == "t ~ x + z"  # By this time, the formula is not yet updated
+    assert param.prior is None
+    assert param.bounds == (0, 1)
+    assert isinstance(param.link, Link)
+
+
+def test_make_param_from_defaults():
```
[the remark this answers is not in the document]
digicosmos86: I understand the rationale with writing DRY code. However, with tests, sometimes it makes sense to know exactly where things go wrong?

tests/param/test_params.py
```diff
@@ -0,0 +1,408 @@
...
+    return model
+
+
+include = [
+    {"name": "v", "prior": {"name": "Normal", "mu": 0, "sigma": 1}},
+    {"name": "a", "prior": bmb.Prior("HalfNormal", sigma=1)},
+]
+
+
+def kw():
```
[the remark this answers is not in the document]
digicosmos86: We chatted about this with fixtures

src/hssm/param/regression_param.py
```diff
@@ -0,0 +1,402 @@
...
+        Process the prior specification.
+    set_loglogit_link
+        Override the default link function.
+    make_safe_priors
+        Override the default priors.
+    reformat_formula
+        Reformat the formula.
+    """
+
+    terms: list[str]
```
cpaniaguam, line 75: I'm a bit confused by this. Is it just a type hint for `terms` for a type checker?

src/hssm/param/regression_param.py
```diff
@@ -0,0 +1,402 @@
...
+        else:
+            _logger.warning(
+                "The bounds for parameter %s (%f, %f) seem strange. Nothing is done to"
+                + " the link function. Please check if they are correct.",
+                self.name,
+                lower,
+                upper,
+            )
+
+    def make_safe_priors(
```
cpaniaguam, line 199: This method seems to be doing a lot of things. You might want to consider splitting it into a couple of methods, one for each of the main blocks of code, or perhaps add some high-level comments describing the what/why for each?

tests/param/test_params.py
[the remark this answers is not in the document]
cpaniaguam: Was a comment added? I don't see any.

cpaniaguam (commented):
Couple more observations.

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+from typing import TYPE_CHECKING, Any
+
+import bambi as bmb
+
+from .param import Param
+from .regression_param import RegressionParam
+from .simple_param import DefaultParam, SimpleParam
+from .user_param import UserParam
+
+if TYPE_CHECKING:
```
AlexanderFengler, line 20: don't understand the purpose of this `if` statement tbh

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+                # parent is a regression
+                else:
+                    rhs = formula.split(" ~ ")[1]
+                    parent_formula = f"{model.response_c} ~ {rhs}"
+                    if prior is not None:
+                        priors[param_name] = prior
+                    if link is not None:
+                        links[param_name] = link
+            else:
+                # non-regression case
```
AlexanderFengler, line 141: maybe add explanation for why only non-regression case needs to be processed here.

src/hssm/param/utils.py
```diff
@@ -0,0 +1,43 @@
...
+    -------
+        A bmb.Prior object representing the default prior for the provided bounds.
+    """
+    if bounds is None:
+        raise ValueError("Parameter unspecified.")
+    lower, upper = bounds
+    if np.isinf(lower) and np.isinf(upper):
+        prior = bmb.Prior("Normal", mu=0.0, sigma=2.0)
+    elif np.isinf(lower) and not np.isinf(upper):
+        prior = bmb.Prior("TruncatedNormal", mu=upper, upper=upper, sigma=2.0)
```
AlexanderFengler, line 34: the idea here is to have a `halfnormal` starting at the upper bound?

AlexanderFengler (commented):
This looks in good shape!

src/hssm/config.py
```diff
@@ -5,14 +5,17 @@
...
+import bambi as bmb
 
 from .defaults import LogLik, LoglikKind, SupportedModels, default_model_config
 
 if TYPE_CHECKING:
     from pytensor.tensor.random.op import RandomVariable
 
-    from .param import ParamSpec
+
+ParamSpec = Union[float, dict[str, Any], bmb.Prior, None]
```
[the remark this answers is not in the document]
digicosmos86: Unfortunately the union `|` operator outside of type annotation is actually a valid operator for dictionaries, so in regular python syntax `Union` is necessary. It's also necessary for forward references that was used in some cases here. This is an inconsistency that Python needs to resolve lol

src/hssm/param/param.py
```diff
@@ -0,0 +1,179 @@
...
+        prior
+            The prior specification for the parameter.
+        bounds
+            The bounds for the parameter.
+        """
+        if self.bounds is None:
+            self.bounds = bounds
+        if self.prior is None:
+            self.prior = prior
+        for key, value in kwargs.items():
```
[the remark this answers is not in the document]
digicosmos86: This is done strictly to have the same function pattern for all sub-classes. See implementation of the same method in subclasses

src/hssm/param/utils.py
digicosmos86: Yep :)

src/hssm/param/param.py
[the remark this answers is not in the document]
AlexanderFengler: ok I see didn't appreciate that.

src/hssm/config.py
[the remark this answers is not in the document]
AlexanderFengler: ok in that case I find it weird to encourage people to NOT use `Union` consistently... 
but ok not our doing.

src/hssm/param/params.py
digicosmos86: This is actually just a bad comment that wasn't deleted

src/hssm/param/regression_param.py
digicosmos86: Added some high level comments

src/hssm/param/regression_param.py
digicosmos86: Yes, for some reason mypy cannot figure out types of properties in the parent class, so we are being specific here

AlexanderFengler (approved):
lgtm

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+            if param_name in user_params:
+                raise ValueError(
+                    f"Parameter `{param_name}` specified in both"
+                    " `include` and `kwargs`."
+                )
+            user_params[param_name] = UserParam.from_kwargs(
+                param_name,
+                # Use pop to remove the key from kwargs
+                kwargs.pop(param_name),
+            )
```
cpaniaguam, line 233: Consider a guard clause:
```suggestion
        if param_name in user_params and param_name in kwargs:
            raise ValueError(
                f"Parameter {param_name} specified in both "
                "include and kwargs."
            )

        # Update user_params only if param_name is in kwargs
        if param_name in kwargs:
            user_params[param_name] = UserParam.from_kwargs(
                param_name,
                kwargs.pop(param_name),
            )
```

src/hssm/param/regression_param.py
```diff
@@ -0,0 +1,406 @@
...
+        if prior is not None:
+            raise ValueError(
+                f"{self.name} is a regression parameter. "
+                + "It should not have a default prior."
+            )
+        super().fill_defaults(prior=None, bounds=bounds)
+        if self.formula is None:
+            if "formula" not in kwargs:
+                raise ValueError(f"Formula not specified for parameter {self.name}.")
+            self.formula = kwargs["formula"]
```
cpaniaguam, line 140: Use a guard clause and avoid the extra indentation?

cpaniaguam (approved):
Approving but left some comments for your considerations about the `parse_bambi` method. I think it's doing many things that are a bit too intertwined.

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+            else:
+                if formula is not None:
+                    other_formulas.append(formula)
+                if prior is not None:
+                    priors[param_name] = prior
+                if link is not None:
+                    links[param_name] = link
+
+        if parent_formula is None:
+            raise ValueError("Parent parameter not found.")
```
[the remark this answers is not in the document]
digicosmos86: This is more to satisfy `mypy` than anything else. If you check the initialization code for `Params`, you'll see that parent_formula is guaranteed to exist. However, if only set inside the loop, type checker doesn't know that `parent_formula` is guaranteed to exist at this point, so the right thing to do is to initialize it to `None` and then use this check to ensure type checker that it's not none.

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+        result_formula: bmb.Formula = bmb.Formula(parent_formula, *other_formulas)
+        result_priors = None if not priors else priors
+        result_links: dict | str = "identity" if not links else links
```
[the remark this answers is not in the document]
digicosmos86: I really appreciate the thinking on this. However, the code was written this way intentionally. In addition to DRY, I also follow these rules:

1. Avoid unnecessary loops. This is why I did everything within the same loop. Separating the logic would result in multiple loops that could impact performance. Maybe it's negligible with fast GPUs and lots of memory, but still a good idea anyway :)
2. Avoid side effects. Distilling the `update_priors_and_links` function means that `priors` and `links` will be modified in-place inside a function, which is not ideal (I did break this rule once in this PR, but that was the only way to modularize the code. That still caused me problems during testing so it was really the last resort).
3. Avoid dangling values. This line `other_formulas.append(formula) if formula is not None else None` results in a `None` value when evaluated if `formula` not `None`, which I avoid. I think GC will have to catch this value and delete it if I am not mistaken. Some languages treat this as an error. Personally I don't use ternary statements if it's not on the right side of an assignment or if the resulting value is immediately used.

I'd love to have other suggestions that also adhere to these rules :)

src/hssm/param/params.py
digicosmos86: This way `param_name in kwargs` is checked twice right?

src/hssm/param/params.py
```diff
@@ -0,0 +1,353 @@
...
+        """
+        user_params = collect_user_params(model, include, kwargs, p_outlier)
+        params = make_params(model, user_params)
+        return cls(params)
+
+    def parse_bambi(
+        self,
+        model: HSSM,
+    ) -> tuple[bmb.Formula, dict | None, dict[str, str | bmb.Link] | str]:
+        """Retrieve three items that helps with bambi model building.
```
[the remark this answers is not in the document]
digicosmos86: Good catch!

src/hssm/param/params.py
cpaniaguam: Yes, but it avoids the nested if statement. One could save the check to a variable but that would add an extra line of code anyway. If `param_name` is always in `kwargs` then the update could be done without the check.

digicosmos86: I think I am OK with just one layer of nesting if that avoids one unnecessary check :). Is the readability that different?

cpaniaguam: To me it is. What about a compromise saving the check to a variable: `param_in_user_params = param_name in user_params` and use it in the second if block?

digicosmos86: Okay. I think I'll compromise without adding this other variable. But in the future maybe we can agree to disagree and not making this as an issue in code reviews?

cpaniaguam: Surely. Do note that I already approved the PR!

Issue #594, opened 2024-10-18
Suppress annomalous output in pytensor.function  


cpaniaguam:
Currently calls to `pytensor.function` in the `tests/test_likelihoods.py::test_analytical_gradient` test function report a "bug" that doesn't throw any exceptions and do not prevent the test from passing. The message comes from the `pytensor.function function` (see bug report https://github.com/pymc-devs/pytensor/issues/1037).

I suggest suppressing this kind of output from the `pytensor.fucntion` function, at least temporarily while the pymc devs resolve this problem. Something like the following seems to work:
```python
import contextlib
import logging
import os

class SuppressOutput:
    def __enter__(self):
        self._null_file = open(os.devnull, "w")
        self._stdout_context = contextlib.redirect_stdout(self._null_file)
        self._stderr_context = contextlib.redirect_stderr(self._null_file)
        self._stdout_context.__enter__()
        self._stderr_context.__enter__()
        logging.disable(logging.CRITICAL)  # Disable logging

    def __exit__(self, exc_type, exc_value, traceback):
        self._stdout_context.__exit__(exc_type, exc_value, traceback)
        self._stderr_context.__exit__(exc_type, exc_value, traceback)
        self._null_file.close()
        logging.disable(logging.NOTSET)  # Re-enable logging

# Example Usage
with SuppressOutput():
    buggy_function()

digicosmos86:
PR merged

PR #595, opened 2024-10-18
fix: suppress buggy output from pytensor.function

cpaniaguam:
See #594.

### DIFF
--- a/src/hssm/utils.py
+++ b/src/hssm/utils.py
@@ -9,8 +9,10 @@ these representations in Bambi-compatible formats through convenience function
 _parse_bambi().
 """
 
+import contextlib
 import itertools
 import logging
+import os
 from copy import deepcopy
 from typing import Any, Literal, cast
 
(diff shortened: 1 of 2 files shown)


AlexanderFengler (commented):
Is this functionality meant to sit in `test_likelihoods.py` or is this wip?

cpaniaguam:
`pytensor.function` is also used in https://github.com/lnccbrown/HSSM/blob/fa12018b146596505b403a4bdf2d7fb9ea475570/src/hssm/distribution_utils/dist.py#L413

It could be used there as well right now.

@AlexanderFengler Where would a good home be for the output suppressor?

digicosmos86 (approved):
Looks good to me! Just one comment: we use numpy docstring format. The docstring is slightly non-standard :)

src/hssm/utils.py
```diff
@@ -548,3 +550,43 @@ def _rearrange_data(data: pd.DataFrame | np.ndarray) -> pd.DataFrame | np.ndarra
 def _split_array(data: np.ndarray | list[int], divisor: int) -> list[np.ndarray]:
     num_splits = len(data) // divisor + (1 if len(data) % divisor != 0 else 0)
     return [tmp.astype(int) for tmp in np.array_split(data, num_splits)]
+
+
+class SuppressOutput:
```
AlexanderFengler, line 555: sorry about the late response here, but his converged to the right spot :)

AlexanderFengler (approved):
lgtm

PR #600, opened 2024-10-30
596 add code coverage

cpaniaguam:
TODO:

- [x] Add secret as shown in https://docs.codecov.com/docs/github-2-getting-a-codecov-account-and-uploading-coverage#add-the-codecov-token-to-your-repo-or-your-org
- [x] Add badge


digicosmos86 (approved):
LGTM now

.github/workflows/run_tests.yml
```diff
@@ -7,7 +7,7 @@ on:
 jobs:
   run_tests:
     runs-on: ubuntu-latest
-    if: ${{ ! contains(github.event.head_commit.message, '[skip slow tests]') }}
+    if: ${{ ! contains(github.event.head_commit.message, '[skip fast tests]') }}
```
cpaniaguam, line 10: @digicosmos86 I am not sure this is needed. Can this line be dropped?

digicosmos86: So the setup is now to run all tests? Can `codecov` collect test results from separate runs? In principle I'd want fast tests to run so the tests fail fast

cpaniaguam: @digicosmos86 
> Can codecov collect test results from separate runs?

It seems like the tooling for that is not reliable -- see https://stackoverflow.com/questions/73688076/pytest-with-cov-append-flag-overwrites-coverage-file. Besides, we are using `xdist` to run the tests in parallel. We could mark the slow tests but it's not clear to me that would matter to `xdist`. We can have a fail fast strategy with the `--exitfirst` flag, however. I'll add that.

AlexanderFengler:
@digicosmos86 the approval here is after the last change? Or another review needed?

digicosmos86:
> @digicosmos86 the approval here is after the last change? Or another review needed?

We need to figure out the fast and slow tests situation. For now let's discuss this a bit more

cpaniaguam:
> > @digicosmos86 the approval here is after the last change? Or another review needed?
> 
> We need to figure out the fast and slow tests situation. For now let's discuss this a bit more

I could try to run the tests in succession and see if the tool for combining coverage reports works for us. However, please keep in mind that the tests are run in parallel, most of them are not slow, and there is a fail-fast strategy in place already.

digicosmos86:
> > > @digicosmos86 the approval here is after the last change? Or another review needed?
> > 
> > 
> > We need to figure out the fast and slow tests situation. For now let's discuss this a bit more
> 
> I could try to run the tests in succession and see if the tool for combining coverage reports works for us. However, please keep in mind that the tests are run in parallel, most of them are not slow, and there is a fail-fast strategy in place already.

Yeah the parallel tests does complicate things a bit, but even though parallelization may introduce some randomness, I think it's still possible to figure out the order in which tests are _scheduled_ to run, which is connected to how pytest searches and collects tests. There should be some documentation on that. I don't mind restructuring (i.e. put all fast tests in one separate directory) the tests a bit so that fast tests are run before the hundreds of slow ones

digicosmos86:
> > > > > @digicosmos86 the approval here is after the last change? Or another review needed?
> > > > 
> > > > 
> > > > We need to figure out the fast and slow tests situation. For now let's discuss this a bit more
> > > 
> > > 
> > > I could try to run the tests in succession and see if the tool for combining coverage reports works for us. However, please keep in mind that the tests are run in parallel, most of them are not slow, and there is a fail-fast strategy in place already.
> > 
> > 
> > Yeah the parallel tests does complicate things a bit, but even though parallelization may introduce some randomness, I think it's still possible to figure out the order in which tests are _scheduled_ to run, which is connected to how pytest searches and collects tests. There should be some documentation on that. I don't mind restructuring (i.e. put all fast tests in one separate directory) the tests a bit so that fast tests are run before the hundreds of slow ones
> 
> @digicosmos86 Please see [this](https://tinyurl.com/37mmd688) and the changes to `tests/conftest.py`.

Ah got it! You did exactly that. Cool then I am OK with merging. @AlexanderFengler would you like to give it another look?

PR #607, opened 2024-12-03
add compile_logp convenience function

AlexanderFengler:
adding small convenience function


src/hssm/distribution_utils/dist.py
```diff
@@ -356,8 +357,210 @@ def rng_fn(
     return SSMRandomVariable
 
 
+def make_custom_rv(
```
digicosmos86, line 360: Does this function have much in common with the previous one? I wonder if we can somehow combine these two into a more general function so we don't repeat the same code?

Also, seems like codecov is complaining that this function was not tested

src/hssm/distribution_utils/dist.py
```diff
@@ -356,8 +357,210 @@ def rng_fn(
...
+                )
+                np.putmask(sims_out, replace_mask, lapse_output)
+            return sims_out
+
+    return CustomRV
+
+
 def make_distribution(
-    rv: str | Type[RandomVariable],
+    rv: str | Type[RandomVariable] | RandomVariable | Callable,
```
digicosmos86, line 563: This argument can be confusing now since it also accepts a callable. Might be a good idea to have a separate `simulator_fn` argument if someone wants to pass a simulator

src/hssm/likelihoods/analytical.py
```diff
@@ -396,22 +396,21 @@ def logp_ddm_sdv(
 ddm_sdv_params = ddm_params + ["sv"]
 
 DDM: Type[pm.Distribution] = make_distribution(
-    "ddm",
-    logp_ddm,
+    rv="ddm",
```
digicosmos86, line 399: does `rv` always have to be passed now? Is there a way to intelligently figure this out from the model string?

digicosmos86 (approved):
Looks good! Only a fewminor comments. I haven't had a chance to review the tutorials but I trust you, @AlexanderFengler 😆

.github/workflows/run_fast_tests.yml
```diff
@@ -0,0 +1,36 @@
+name: Run fast tests
```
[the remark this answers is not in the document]
AlexanderFengler: They came from a recent update, this is some git weirdness :/. 
Seemed harmless enough so I let it pass, but annoying and not exactly sure what happened there.

PR #612, opened 2024-12-17
Replace Poetry with uv for dependency management and workflows

cpaniaguam:
This pull request includes significant changes to the setup and testing workflows, as well as updates to the project's dependency management. The most important changes include the replacement of Poetry with `uv` for environment setup and dependency management, consolidation of test workflows, and updates to the `pyproject.toml` file to align with the new dependency management tool.

### Workflow and Environment Setup:
* [`.github/setup-env/action.yml`](diffhunk://#diff-ba24ca0a9e1929f638a09a15dc999bd10cd678783b903984afec77eb3d7800b3R1-R28): Added a new action file to set up the Python environment using `uv` instead of Poetry. It necessary to add a step to install the blas and lapack libraries.
* [`.github/workflows/build_and_publish.yml`](diffhunk://#diff-ebf33a46dfb7b44353d6b00f5978292fa922bcd221f948c81cd7aa3afd70e5baL13-R15): Consolidated the fast and slow test jobs into a single `run_tests` job and updated the environment setup to use `uv`. [[1]](diffhunk://#diff-ebf33a46dfb7b44353d6b00f5978292fa922bcd221f948c81cd7aa3afd70e5baL13-R15) [[2]](diffhunk://#diff-ebf33a46dfb7b44353d6b00f5978292fa922bcd221f948c81cd7aa3afd70e5baL34-R43)
* `.github/workflows/build_docs.yml`, `.github/workflows/linting_and_type_checking.yml`, `.github/workflows/run_tests.yml`: Updated to use the new `setup-env` action and replaced Poetry commands with `uv` commands. [[1]](diffhunk://#diff-4dfe935b4852393622c2b414783d1d11e1a868fd8485709124d90208fa61ec8bL19-R25) [[2]](diffhunk://#diff-228789bb0049580d97e799082e4d5bf439b066726aeda792e1476f5dddb838a5L20-R32) [[3]](diffhunk://#diff-c2e9f7c1bfaf65da79bbd29ed3ce389b2acc8b4099a257914a7292a66e6b9bc9L23-R32)

### Dependency Management:
* [`pyproject.toml`](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L1-R56): Transitioned from Poetry to `uv` for dependency management. Updated the file structure to conform to the new tool, including changes to dependencies and build system. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L1-R56) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L175-R182)

### Documentation:
* [`docs/CONTRIBUTING.md`](diffhunk://#diff-4c6b93aa75d5affde60dc3849606c9acd75ed444d52e99f3055fc0c7aa77e9e0L64-R74): Updated the contribution guide to reflect the switch from Poetry to `uv` for managing dependencies. [[1]](diffhunk://#diff-4c6b93aa75d5affde60dc3849606c9acd75ed444d52e99f3055fc0c7aa77e9e0L64-R74) [[2]](diffhunk://#diff-4c6b93aa75d5affde60dc3849606c9acd75ed444d52e99f3055fc0c7aa77e9e0L88)

### Code Cleanup:
* Various files in `src/hssm/` and `tests/`: Minor formatting changes as per the latest version of the `ruff` formatter. [[1]](diffhunk://#diff-886b4f455536b5c662d101c272845eaf0d7ea4acc0475bc11473cbf11147546eL1333-R1335) [[2]](diffhunk://#diff-886b4f455536b5c662d101c272845eaf0d7ea4acc0475bc11473cbf11147546eL1358-R1358) [[3]](diffhunk://#diff-886b4f455536b5c662d101c272845eaf0d7ea4acc0475bc11473cbf11147546eL1400-R1401) [[4]](diffhunk://#diff-bb00556f25602c8a368f814d7d970629a480891007666b5981a51ec65c8cda24L175-R175) [[5]](diffhunk://#diff-e61fd6e4066e2fdf15ed8d19bf33eceb475e4c2758d42db8a315f09efd697417L175-R176) [[6]](diffhunk://#diff-e04eb437c8cd0ca4d88142b49bf40d703315e9144651a303d283703c299409e7L257-R257) [[7]](diffhunk://#diff-909736c07bcfd1677acaeb10229a5b3c6026609ab5189f1a3ac5aefcd45dbef6L104-R104) [[8]](diffhunk://#diff-909736c07bcfd1677acaeb10229a5b3c6026609ab5189f1a3ac5aefcd45dbef6L117-R126) [[9]](diffhunk://#diff-19a7aa3cce0c461d18f967d0ebb6a12458a1156c9f63ff9e84bd36a0af0a93c1L309-R321) [[10]](diffhunk://#diff-b90c0fb80d86b59ff2cf6d51ebeafaf45d1ed08f9155850441829df93a3ea337L73-R75)


AlexanderFengler:
@cpaniaguam this implies using uv all the way or specific to github workflows?

.github/setup-env/action.yml
```diff
@@ -0,0 +1,32 @@
...
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v4
+      with:
+        # Install a specific version of uv.
+        version: "0.5.9"
+
+    - name: Set up Python ${{ inputs.python-version }}
+      run: uv python install ${{ inputs.python-version }}
+      shell: bash
```
digicosmos86, line 27: It seems that uv determins the default Python version for the project by using the `.python-version` file in the project root. So I think the above processes can be simplified to:

1. Checkout repo
2. Install UV
3. Create `.python-version` in root directory using Python `inputs.python-version`. (this will require .python-version` to be included in `.gitignore`.
4. Run `uv sync --all-extras` to install all dependencies.

If we are going with the System python route, we don't need `uv`. We can just pip install all dependencies from `pyproject.toml`. Thoguhts?

.github/workflows/build_and_publish.yml
```diff
@@ -31,26 +27,31 @@ jobs:
...
+          python -m twine upload --repository testpypi dist/*
 
       - name: Publish to PyPI
         env:
-          POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI_TOKEN }}
-        run: poetry publish
+          TWINE_USERNAME: __token__
+          TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
+        run: |
+          python -m twine upload dist/*
```
digicosmos86, line 50: Since we set up `uv` in the previous step, let's use `uv` for this?

.github/workflows/build_and_publish.yml
```diff
@@ -31,26 +27,31 @@ jobs:
...
+
+      - name: Install doc dependencies
+        run: |
+          $HOME/.cargo/bin/uv pip install -e '.[dev]'
 
       - name: Build and publish docs
-        run: poetry run mkdocs gh-deploy --force
+        run: mkdocs gh-deploy --force
```
digicosmos86, line 57: Instead of `poetry run`, we can just do `uv run`

pyproject.toml
```diff
@@ -1,54 +1,54 @@
...
 name = "HSSM"
 version = "0.2.4"
 description = "Bayesian inference for hierarchical sequential sampling models."
 authors = [
-    "Alexander Fengler <alexander_fengler@brown.edu>",
-    "Paul Xu <yang_xu@brown.edu>",
-    "Aisulu Omar <aisulu_omar@brown.edu>",
-    "Michael Frank <michael_frank@brown.edu>",
+    {name = "Alexander Fengler", email = "alexander_fengler@brown.edu"},
+    {name = "Paul Xu", email = "yang_xu@brown.edu"},
```
digicosmos86, line 7: @cpaniaguam you can add your name here too

digicosmos86 (changes requested):
Looks good! The workflows can be a bit confusing since it both installs UV and system Python and the use of system Python and uv commands are intertwined. To disentangle this, we can do one of the following:

1. Use uv instead of system Python. UV will create a virtual environment with Python installed, and we run all commands with `uv`.
2. Use system Python without using UV. All dependencies are supplied in `pyproject.toml`, so there is no real need for `uv` in github actions.

Personally I think option 1 works a bit better since it replicates our local workflow, and `uv` installs dependencies quite a bit faster.

.github/setup-env/action.yml
cpaniaguam: I think it'd be simpler to just use the system's python. I was kind of trying to keep the same spirit of the current workflow which does something like this but with Poetry. Additionally, aren't the workflows run with different versions of Python? How could this be done if a single Python version is specified in a file? Currently this is accomplished programmatically within each of the workflows that depend on this `action.yml`.

digicosmos86: I am fine with either sticking to system Python or using UV to handle everything. The current approach is a mixture of both so can be a bit confusing at times. For example, you run the `Set up Python` step which installs a system Python and now you are using UV to install another version, which is a bit redundant.

If the only issue with using `uv` is how to pin the python version to use for the matrix run, you can run `uv python pin ${{ inputs.python-version }}` to pin the version to be used with the project, and then proceed to run `uv sync` to set up the virtual environment.

I still prefer the uv way because of the many nice functionalities and speed that uv offers. It also replicates our local workflow which is nice. However, if you prefer the system Python way, then let's not include uv in the workflow

.github/setup-env/action.yml
```diff
@@ -0,0 +1,28 @@
...
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v4
+      with:
+        # Install a specific version of uv.
+        version: "0.5.9"
+
+    - name: Install hssm
+      if: steps.cache.outputs.cache-hit != 'true'
+      run: uv sync --all-extras
```
digicosmos86, line 27: This is my fault. We only need to run `uv sync`. We have an optional dependency group that installs JAX with GPU support, which is unsupported in the dev environment. Let's not install those dependencies.

.github/workflows/build_and_publish.yml
```diff
@@ -30,27 +26,18 @@ jobs:
...
-      - name: Publish to test-PyPI
-        run: poetry publish -r testpypi
+        run: uv build
 
       - name: Publish to PyPI
-        env:
-          POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI_TOKEN }}
-        run: poetry publish
+        run: |
+          uv publish --token ${{ secrets.PYPI_TOKEN }}
```
digicosmos86, line 40: We still need to publish to testpypi first just to make sure that publishing works

digicosmos86 (approved):
Looks good to me. Just a few minor comments

.github/workflows/build_and_publish.yml
cpaniaguam: It's not clear to me how to do this with `uv` or whether it supports this. For now I could keep this part of the workflow using poetry. Thoughts?

digicosmos86: Seems to me this should not be difficult. This is their documentation on adding a different index than PyPI:

https://docs.astral.sh/uv/guides/publish/#building-your-package

Let me know if there's further issues

AlexanderFengler (approved):
lgtm.

PR #651, opened 2025-02-06
644 implement model registrator

cpaniaguam:
This pull request introduces model registration functionality.

### Model Registration:
* Added a new module `register.py` for registering custom models in HSSM, including functions `register_model`, `list_registered_models`, and `get_model_info`.
* Updated `src/hssm/__init__.py` to include the `register_model` function. [[1]](diffhunk://#diff-a680f8ceffe23227f4a96d6b698b47440b4855d861c6919cfc3eb06155dfff52R21) [[2]](diffhunk://#diff-a680f8ceffe23227f4a96d6b698b47440b4855d861c6919cfc3eb06155dfff52R39)
* Tests

### Dependency Updates:
* Added `ipython`, `nbconvert`, and `ptpython` to the `dev` dependencies in `pyproject.toml`. [[1]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R42) [[2]](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R51-R54)

### Initialization Refactoring:
* Refactored the `__init__` method in `src/hssm/hssm.py` to use a dictionary comprehension for initializing `self._init_args`.


src/hssm/register.py
```diff
@@ -0,0 +1,84 @@
+"""Module for registering custom models in HSSM."""
+
+from pprint import pformat
+from typing import Dict
```
digicosmos86, line 4: Since we are 3.10+ we can just use `dict` without the import

src/hssm/register.py
```diff
@@ -0,0 +1,84 @@
...
+    # Validate required keys
+    required_keys = set(DefaultConfig.__annotations__.keys())
+    missing_keys = required_keys - set(config.keys())
+    if missing_keys:
+        raise ValueError(f"Missing required keys in config: {missing_keys}")
+
+    # TODO: validate provided configs?
+
+    # Register the model
+    default_model_config[name] = config
```
digicosmos86, line 46: We are going to improve on this, but I would add some safeguard so that the original defaults are not accidentally overwritten. We can probably provide a `get_model_defaults` that returns a deepcopy of a `DefaultConfig` object for people to overwrite if they only want to change on certain aspects of an existing model

src/hssm/register.py
```diff
@@ -0,0 +1,84 @@
...
+from pprint import pformat
+from typing import Dict
+
+from .defaults import (
+    DefaultConfig,
+    default_model_config,
+)
+
+
+def register_model(name: str, config: DefaultConfig) -> None:
```
digicosmos86, line 12: I can be convinced otherwise, but I prefer to actually have the fields in the function signature, and when people have a `DefaultConfig`, they can simply pass `**config` to the function. I think it's more explicit to document the types of the argument

src/hssm/register.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Raises
+    ------
+    ValueError
+        If the model name is not found
+    """
+    if name not in default_model_config:
+        raise ValueError(f"Model '{name}' not found")
+
+    return f"Model: {name}\n" + pformat(default_model_config[name], indent=2)
```
digicosmos86, line 84: Why not include `pformat()` in the f-string to avoid creating 2 strings?

digicosmos86 (commented):
This looks great! I really like how clean it is, and it is exactly what we were thinking of.

Taking this further, we can now move on to thinking about how not to modify the `default_model_config` object (it is really supposed to be read-only) and also lazy-loading only certain model configs as needed, but this is already pretty great!

src/hssm/register.py
cpaniaguam: I want `pformat` to act on the dict only. The whole string gets messy when other stuff is added to it and does not render well with `print`.

src/hssm/register.py
digicosmos86: Yep, that's basically it :)

src/hssm/register.py
digicosmos86: So you mean that `f"Model: {name}\n{pformat(default_mode_config[name], indent=2)}"` would return something different?

cpaniaguam: Oh I see what you mean. I find that a bit more difficult to read, however. Maybe we can compromise a bit of performance for readability?

src/hssm/register.py
```diff
@@ -0,0 +1,96 @@
...
+from pprint import pformat, pp
+from typing import cast
+
+from .defaults import (
+    DefaultConfig,
+    LoglikConfigs,
+    SupportedModels,
+)
+from .defaults import (
+    default_model_config as registered_models,
```
digicosmos86, line 12: This is not exactly lazy. Once `default_model_config` is imported, the entire dictionary is loaded into memory.

One thing that we can make the loading process lazy is to separate the dictionary into variables with the same name as the model, and then we can dynamically import them as needed. Check out [`importlib.import_module`](https://docs.python.org/3/library/importlib.html#importlib.import_module)

There are definitely other ways in which we can do this. Feel free to use whatever you think works best

PR #677, opened 2025-02-28
659 import config functions on demand

cpaniaguam:
Two new modules `_types` with all types used in `defaults` and `modelmeta` for default model metadata accessors.


AlexanderFengler (approved):
lgtm! Thanks @cpaniaguam 
Still think we probably want to split `modelmeta.py` up into separate files per model, but that's not a blocker here, can do downstream.

AlexanderFengler:
@cpaniaguam is this still intended to be draft?

cpaniaguam:
> @cpaniaguam is this still intended to be draft?

Yes. I'll go back to this after merging #681.

PR #682, opened 2025-03-05
P outlier defaults

AlexanderFengler:
adding in missing defaults for p-outlier regression, and a small associated tutorial.


src/hssm/hssm.py
```diff
@@ -1737,12 +1737,12 @@ def __repr__(self) -> str:
...
-            if self.p_outlier.is_regression:
-                raise NotImplementedError(
-                    "Regression for `p_outlier` is not implemented yet."
-                )
+            # # TODO: Allow regression for self.p_outlier
+            # # Need to determine what the output should look like
+            # if self.p_outlier.is_regression:
+            #     raise NotImplementedError(
+            #         "Regression for `p_outlier` is not implemented yet."
+            #     )
```
cpaniaguam, line 1745: Keep commented code or ok to delete?

src/hssm/prior.py
```diff
@@ -321,16 +321,18 @@ def get_hddm_default_prior(
...
-    "z": {"dist": "Weibull", "alpha": 1.2, "beta": "0.25"},
+    "z": {"dist": "Weibull", "alpha": 1.2, "beta": 0.25},
     "t": {"dist": "HalfNormal", "sigma": 0.2},
-    "sv": {"dist": "Weibull", "alpha": 1.5, "beta": "0.3"},
-    "sz": {"dist": "Weibull", "alpha": 1.5, "beta": "0.3"},
-    "st": {"dist": "Weibull", "alpha": 1.5, "beta": "0.3"},
+    "sv": {"dist": "Weibull", "alpha": 1.5, "beta": 0.3},
+    "sz": {"dist": "Weibull", "alpha": 1.5, "beta": 0.3},
+    "st": {"dist": "Weibull", "alpha": 1.5, "beta": 0.3},
+    "p_outlier": {"dist": "Weibull", "alpha": 1.5, "beta": 0.3},
```
cpaniaguam, line 335: Why the switch to actual number types?

cpaniaguam (approved):
lgtm

src/hssm/hssm.py
```diff
@@ -1353,9 +1353,9 @@ def plot_trace(
             Whether to call plt.tight_layout() after plotting. Defaults to True.
         """
         data = data or self.traces
-        assert isinstance(data, az.InferenceData), (
-            "data must be an InferenceData object."
-        )
+        assert isinstance(
+            data, az.InferenceData
+        ), "data must be an InferenceData object."
```
digicosmos86, line 1358: I'd prefer throwing more informative error types like `TypeError` instead of using assertions here

digicosmos86 (approved):
LGTM! I think we also need to change some associated type annotations and for `p_outliers` in `HSSM` and other functions to make explicit that we do support regressions for it. It could be another PR though

src/hssm/prior.py
AlexanderFengler: To be honest I am confused how this became a string before. As far as I am concerned this should have always been float. @cpaniaguam @digicosmos86

PR #687, opened 2025-03-14
Update pymc version in pyproject.toml to 5.21.0

cpaniaguam:
Dependency updates:

* [`pyproject.toml`](diffhunk://#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711L25-R27): Updated the `pymc` dependency to version `5.21.0` and drop `numpy`.

Type casting syntax updates per `ruff` v0.11.


src/hssm/config.py
```diff
@@ -54,7 +54,7 @@ def from_defaults(
         loglik_kind
             The kind of the log-likelihood for the model.
         """
-        model_name_casted = cast(SupportedModels, model_name)
+        model_name_casted = cast("SupportedModels", model_name)
```
AlexanderFengler, line 57: what made it necessary to turn all these types into strings?

AlexanderFengler (commented):
The string casting is a bit scary.
Can you explain? Otherwise looks like the version change just passes through? @cpaniaguam

src/hssm/config.py
cpaniaguam: That's Ruff v.0.11 fault.

AlexanderFengler: Is this somehow equivalent?

AlexanderFengler: I may just not understand typing properly then

AlexanderFengler: Tried checking the release notes not but still confused :D

cpaniaguam: Seems like it. I guess `cast` might also do the casting of the type to `str`, but Ruff's linting was insisting on making those changes. We can always downgrade to a previous version of Ruff.

AlexanderFengler (approved):
lgtm. Thanks @cpaniaguam

PR #695, opened 2025-03-26
Add dependabot

cpaniaguam:
This pull request includes a change to the `.github/dependabot.yml` file to set an update schedule for GitHub Actions.

* [`.github/dependabot.yml`](diffhunk://#diff-dd4fbda47e51f1e35defb9275a9cd9c212ecde0b870cba89ddaaae65c5f3cd28R1-R10): Added configuration to check for updates to GitHub Actions every month.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every month
+      interval: "monthly"


digicosmos86:
I defer to @AlexanderFengler here, but we had to do lots of version fixes in the past due to pymc/numpy/bambi/jax compatibility issues, which can sometimes get pretty ugly. Not sure how dependabot would be helpful here. For example, JAX is still in development and frequently introduce breaking changes that breaks the package

cpaniaguam:
> I defer to @AlexanderFengler here, but we had to do lots of version fixes in the past due to pymc/numpy/bambi/jax compatibility issues, which can sometimes get pretty ugly. Not sure how dependabot would be helpful here. For example, JAX is still in development and frequently introduce breaking changes that breaks the package

I think dependabot suggests upgrades and creates a PR. If there are problems one can simply ignore and close the PR. But if the package builds and tests pass, that would be a win.

digicosmos86:
> > I defer to @AlexanderFengler here, but we had to do lots of version fixes in the past due to pymc/numpy/bambi/jax compatibility issues, which can sometimes get pretty ugly. Not sure how dependabot would be helpful here. For example, JAX is still in development and frequently introduce breaking changes that breaks the package
> 
> I think dependabot suggests upgrades and creates a PR. If there are problems one can simply ignore and close the PR. But if the package builds and tests pass, that would be a win.

I see. That makes sense. Thanks @cpaniaguam!

digicosmos86 (approved):
LGTM!

PR #712, opened 2025-04-07
Mark slow test plotting

cpaniaguam:


AlexanderFengler:
after merging main, pytorch can't be installed ?!?!?! @cpaniaguam

AlexanderFengler:
@cpaniaguam tried to source this error:

```
error: Distribution `torch==2.0.1 @ registry+[https://pypi.org/simple/`](https://pypi.org/simple/%60) can't be installed because it doesn't have a source distribution or wheel for the current platform
```

So far no luck, I don't get why we need to install exactly that version of torch.

PR #730, opened 2025-06-10
Remove Python 3.12 from test matrix in GitHub Actions workflow

cpaniaguam:
Trying reduced matrix with py 3.10 and 3.11.

### DIFF
--- a/tests/test_data_sanity.py
+++ b/tests/test_data_sanity.py
@@ -11,21 +11,29 @@ def cpn():
     return Path(__file__).parent / "fixtures" / "ddm_cpn.onnx"
 
 
-@pytest.mark.slow
-def test_data_sanity_check(data_ddm, cpn, caplog):
+pattern = r"Field\(s\) `.*` not found in data\."
+
+# The DataValidator class is tested in the test_data_validator.py file, so this file
+# can probably be removed in the future. CP
+
+
+def test_data_sanity_check(data_ddm):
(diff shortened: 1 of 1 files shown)


digicosmos86:
I found out what the issue is. The optional group `notebook` has `lanfactory` as the dependency, which will install PyTorch. We have 3 options:

1. Remove the notebooks that require `lanfactory` and move them to `lanfactory` tutorials
2. Only install this dependency group in one Python version for building notebooks. This will us to install the CPU version of PyTorch before `uv sync`
3. Update `lanfactory` to a PyTorch version that supports 3.12. We still shouldn't be installing `lanfactory` as a dependency in a testing environment though, because `torch` is just way too big to keep the tests light-weight

We shouldn't remove Python 3.12 tests from the matrix if we want to support Python 3.12.

AlexanderFengler:
@digicosmos86 we might just have to cut the next release of lanfactory here. 
On main, lanfactory is on `torch<=2.7.0` already.

digicosmos86:
> Thanks for the insight. This wasn't happening before, so I'm curious about what changed. Were the PyTorch wheels for Python 3.12 removed for this version?
> 
> In my opinion, **option 3** seems like the most robust long-term solution.

From what I can see, `run tests` actions have been failing since a few months ago. Also, unless you do `uv sync --all-groups` locally, uv won't install optional dependencies, so you wouldn't encounter this problem locally

PR #731, opened 2025-06-10
Update Python version in test matrix to 3.13

cpaniaguam:


AlexanderFengler:
@cpaniaguam since it's the same story, let's just cut the lanfactory release, then try adding 3.12 (back) and 3.13 to the matrix?

Issue #737, opened 2025-06-25
A more general process for wrapping JAX likelihood functions


digicosmos86:
Due to legacy reasons, the workflow of wrapping JAX functions in PyTensor ops is specific to converting ONNX networks. On the other hand, there is another wrapper function for non-differentiable JAX functions only. In light of the addition of more complex RLSSM likelihood, this process needs to be generalized to arbitrary differentiable and non-differentiable JAX likelihoods, treating ONNX as a special case

PR #741, opened 2025-06-25
Reorganized the `distribution_utils` submodule

digicosmos86:
As #737 indicates, we need a more general workflow for wrapping JAX likelihoods in pytensor ops. Currently this workflow is entangled with the conversion of ONNX likelihoods to pytensor ops, which should be treated as a special case. This PR addresses this entanglement a little bit. There is still more to be done.

The following was done:
1. Separated `make_vmap_func` from `make_jax_logp_funcs_from_onnx`, so it can be used separately to create vectorized functions, even though its utility outside of the `onnx` case still remains to be seen.
2. Elevated `make_jax_logp` and `make_pytensor_logp` functions from `distribution_utils.onnx` submodule, and renamed the submodule to `distribution_utils.onnx_utils`, so that it only contains utility functions for `onnx`-related stuff.
3. Separated `make_jax_logp` into a separate `jax.py` file, making it more general for all jax use cases.
4. Added a `return_jit` toggle for a few functions, so that the non-jitted onnx wrapper functions can be used in other JAX functions in the RL case.

@AlexanderFengler we might also need to move `make_jax_logp_funcs_from_jax_callable` from `onnx.py` to `jax.py` as well, but since it depends on another function in `onnx.py`, I didn't do anything to avoid circular imports. I think this function should stand alone for more general JAX use cases.


src/hssm/distribution_utils/jax.py
```diff
@@ -161,7 +78,10 @@ def vjp_vmap_logp(
         _, vjp_fn = vjp(vmap_logp, *inputs)
         return vjp_fn(gz) if params_only else vjp_fn(gz)[1:]
 
-    return jit(vmap_logp), jit(vjp_vmap_logp), vmap_logp
+    if return_jit:
+        return jit(vmap_logp), jit(vjp_vmap_logp), vmap_logp
+
+    return vmap_logp, cast("LogLikeGrad", vjp_vmap_logp)
```
cpaniaguam, line 98: @digicosmos86 Where do you stand on "type-stability"? We know in Python this isn't a thing, but adhering to it as a principle can make code easier to maintain and understand. 

Consider extracting the JIT-related logic into a separate function (e.g., `make_jitted_vmap_func`). This would keep each function type-stable, simplify the interface, and make the codebase easier to maintain and use. Callers could then explicitly choose between the jitted and non-jitted versions.

src/hssm/distribution_utils/dist.py
```diff
@@ -708,11 +705,14 @@ def make_likelihood_callable(
...
-                    params_only=False if params_only is None else params_only,
-                )
+                logp, logp_grad, logp_nojit = cast(
+                    "tuple[LogLikeFunc, LogLikeGrad, LogLikeFunc]",
+                    make_jax_logp_funcs_from_jax_callable(
+                        loglik,
+                        params_is_reg,
+                        params_only=False if params_only is None else params_only,
+                    ),
+                )  # type: ignore[assignment]
```
cpaniaguam: I think most of this casting would be simplified if the functions were type-stable.

src/hssm/distribution_utils/onnx.py
```diff
@@ -0,0 +1,208 @@
...
+            supports the case where only parameters are passed.
+
+        Returns
+        -------
+        jnp.ndarray
+            The element-wise log-likelihoods.
+        """
+        # Makes a matrix to feed to the LAN model
+        # print("scalars only: ", scalars_only)
+        # print("params only: ", params_only)
```
cpaniaguam:
```suggestion
```

src/hssm/distribution_utils/onnx.py
```diff
@@ -0,0 +1,208 @@
...
+            return jit(logp_vec), jit(grad(logp)), logp_vec
+        return logp_vec, grad(logp)
+
+    in_axes: list = [0 if is_regression else None for is_regression in params_is_reg]
+    if not params_only:
+        in_axes = [0] + in_axes
+
+    return make_vmap_func(
+        logp, in_axes=in_axes, params_only=params_only, return_jit=return_jit
+    )
```
cpaniaguam: Continuing in the spirit of type-stable functions, I think this is function could be modularized to make it type-stable. Maybe for a separate issue?

src/hssm/distribution_utils/jax.py
digicosmos86: I think I am on the fence about this. While I do like type-stability and how that leads to better readability and better maintainability, the change will lead to a separate function that the users will need to call, which is what this function is trying to achieve - all functionalities in one function. It'll also break existing tutorials on how to use this function. I'll leave this decision to you and @AlexanderFengler. Feel free to start a new PR to improve on this if you think the improvement is worth the effort.

digicosmos86: @cpaniaguam I found a way to achieve type narrowing without defining a new function. Did a few things to make mypy happy

pyproject.toml
```diff
@@ -61,14 +61,10 @@ test = [
...
     "ruff>=0.9.1",
-    ]
-
-notebook = [
-    "ipykernel>=6.29.5",
-    "lanfactory>=0.4.4",
-    "zeus-mcmc>=2.5.4",
 ]
 
+notebook = ["ipykernel>=6.29.5", "lanfactory>=0.4.4", "zeus-mcmc>=2.5.4"]
```
AlexanderFengler, line 66: can you put back into stacked format? 
more coherent with the rest of the file

src/hssm/distribution_utils/onnx.py
```diff
@@ -0,0 +1,248 @@
...
+        return make_jax_logp_funcs_from_onnx(
+            logp, params_is_reg, params_only, return_jit=True
+        )
+    return make_jax_logp_funcs_from_onnx(
+        logp, params_is_reg, params_only, return_jit=False
+    )
+
+
+@overload
+def make_jax_logp_funcs_from_onnx(
```
AlexanderFengler, line 84: The name says `make_jax_logp_funcs_from_onnx`, but we allow passing a callable. I guess that was always weird. Definitely we have to prioritize disentangling this properly (as suggested in the PR msg).

src/hssm/distribution_utils/onnx.py
```diff
@@ -0,0 +1,248 @@
...
+        forward pass, the second calculates the VJP, and the third is
+        the forward-pass that's not jitted. When `params_only` is True, and all
+        parameters are scalars, only a scalar function, its gradient, and the non-jitted
+        version of the function are returned.
+        If `return_jit` is `False`, a pair of jax functions. The first
+        calculates the forward pass, and the second calculates the VJP of the logp
+        function.
+    """
+    if isinstance(model, (str, PathLike, onnx.ModelProto)):
+        loaded_model = (
```
AlexanderFengler: loader_model --> `model_onnx`?

To help semantically.

src/hssm/distribution_utils/onnx.py
```diff
@@ -0,0 +1,248 @@
...
+        if params_only:
+            input_vector = jnp.array(inputs)
+        else:
+            data = inputs[0]
+            dist_params = inputs[1:]
+            param_vector = jnp.array([inp.squeeze() for inp in dist_params])
+            if param_vector.shape[-1] == 1:
+                param_vector = param_vector.squeeze(axis=-1)
+            input_vector = jnp.concatenate((param_vector, data))
+        if isinstance(model, (str, PathLike, onnx.ModelProto)):
```
AlexanderFengler, line 170: down here, is the model not already loaded, hence `onnx.ModelProto` ?

src/hssm/distribution_utils/onnx.py
```diff
@@ -0,0 +1,248 @@
...
+        Parameters that are regressions will not be vectorized in likelihood
+        calculations.
+
+    Returns
+    -------
+    Callable
+        The logp function that applies the ONNX model to data and returns the element-
+        wise log-likelihoods.
+    """
+    loaded_model: onnx.ModelProto = (
```
AlexanderFengler: speaking to my point about `loaded_model` typing above.

AlexanderFengler (commented):
Some small comments.

pyproject.toml
digicosmos86: Sure! I'll do this in future PRs to avoid merge conflicts

src/hssm/distribution_utils/onnx.py
digicosmos86: The loaded model is called `model_onnx`. I think the goal of this line is to determine that if `model` is not a callable, then proceed to use `interpret_onnx` to compute the loaded model. I think this weirdness is caused by the ad-hoc addition of `model_callable` to this function. As I pointed out, we should make a separate function for general jax function

src/hssm/distribution_utils/onnx.py
digicosmos86: I created a new issue for this

PR #742, opened 2025-06-25
Make the `make_jax_logp_op` function more general

digicosmos86:
Added a `n_params` parameter to the function to help the function figure out which inputs for which not to compute gradient. Helpful for RL likelihoods

### DIFF
--- a/src/hssm/distribution_utils/jax.py
+++ b/src/hssm/distribution_utils/jax.py
@@ -116,9 +116,9 @@ def make_jax_logp_ops(
     logp_nojit
         The non-jit version of logp.
     n_params : optional
-        Only required when `extra_fields` are used, in which case the resulting Op will
-        not compute gradients with respect to the extra fields. Supply the number of
-        parameters that are used in the likelihood computation.
+        Number of parameters used in the likelihood computation. Only required
+        when `extra_fields` are used, in which case the resulting Op will not
+        compute gradients with respect to the extra fields.
 
     Returns
     -------


src/hssm/distribution_utils/jax.py
```diff
@@ -128,7 +133,10 @@ def make_node(self, data, *dist_params):
             """
             inputs = [pt.as_tensor_variable(dist_param) for dist_param in dist_params]
             self.scalars_only = all(inp.ndim == 0 for inp in inputs)
+            # params_only means calculate gradients only with respect to the
+            # parameters, not the data.
             self.params_only = data is not None
```
cpaniaguam: Because this is a boolean, consider `is_params_only`.

src/hssm/distribution_utils/jax.py
```diff
@@ -128,7 +133,10 @@ def make_node(self, data, *dist_params):
             """
             inputs = [pt.as_tensor_variable(dist_param) for dist_param in dist_params]
             self.scalars_only = all(inp.ndim == 0 for inp in inputs)
```
cpaniaguam: `is_scalars_only`?

src/hssm/distribution_utils/jax.py
```diff
@@ -128,7 +133,10 @@ def make_node(self, data, *dist_params):
...
+            # params_only means calculate gradients only with respect to the
+            # parameters, not the data.
             self.params_only = data is not None
+            self.n_params = n_params
 
             if self.params_only:
```
cpaniaguam: This would read `if self.is_params_only`:.

AlexanderFengler (commented):
I think the renaming @cpaniaguam suggest makes sense, otherwise this looks good to me. 

All old use-cases basically mindlessly supply `n_params` as `None`? In other words, this is doesn't create any other sharp edges? 

Thanks @digicosmos86

digicosmos86:
> I think the renaming @cpaniaguam suggest makes sense, otherwise this looks good to me.
> 
> All old use-cases basically mindlessly supply `n_params` as `None`? In other words, this is doesn't create any other sharp edges?
> 
> Thanks @digicosmos86

Correct. In old code the gradient of `extra_fields` will be computed. It's just slightly more inefficient. However, other than RL use cases, I haven't seen much needs to use `extra_fields`, so this could just be specific to the RL use cases

digicosmos86:
@cpaniaguam @AlexanderFengler can you approve this one too? Thanks!

src/hssm/distribution_utils/jax.py
```diff
@@ -114,6 +115,10 @@ def make_jax_logp_ops(
         The Jax function that calculates the VJP of the logp function.
     logp_nojit
         The non-jit version of logp.
+    n_params : optional
+        Only required when `extra_fields` are used, in which case the resulting Op will
+        not compute gradients with respect to the extra fields. Supply the number of
+        parameters that are used in the likelihood computation.
```
cpaniaguam: I don't understand the last sentence here.

digicosmos86: So basically the signature of the function is `logp(data, *dist_params, *extra_fields)`, where `dist_params` is a list that contains all parameters from the model, and `extra_fields` is a list that contains all additional data required to compute the likelihood. We need to compute gradients for everything in `dist_params` but not `extra_fields`. Unforutnately, from inside the function, the only thing that the function can see is a gigantic list passed to it (Python only supports one `*args` when the function is defined). So we need the length of `dist_params` to figure out the gradients for which items needs to be computed

cpaniaguam: Thanks for the context.
```suggestion
    n_params : optional
        Number of parameters used in the likelihood computation. Only required when `extra_fields` are used, in which case the resulting Op will not compute gradients with respect to the extra fields.
```

PR #746, opened 2025-07-01
Incorporate RLDM likelihood function

digicosmos86:
This PR incorporates the likelihood function of RLDM into the likelihoods submodule. Some changes are from other PRs, but the main changes are in `src/hssm/likelihoods/rldm.py`. Main additions are:

- Incorporated @krishnbera's differentiable RLDM likelihood
- factory functions for building these likelihoods based on number of participants and trials per participant. This is because JAX `jit` requires these numbers to be static.
- a high-level, user-facing `make_rldm_logp_op` that handles the generation of the `pytensor` op to be used directly with HSSM


src/hssm/likelihoods/rldm.py
```diff
@@ -0,0 +1,260 @@
...
+
+import jax
+import jax.numpy as jnp
+from jax.lax import dynamic_slice, scan
+
+from ..distribution_utils.jax import make_jax_logp_ops
+from ..distribution_utils.onnx import make_jax_logp_funcs_from_onnx
+from ..utils import download_hf
+
+angle_onnx = download_hf("angle.onnx")
```
cpaniaguam, line 13: Will this be extended later?

src/hssm/likelihoods/rldm.py
```diff
@@ -0,0 +1,260 @@
...
+    subj_scaler = dynamic_slice(scaler, [subj * ntrials_subj], [ntrials_subj])
+    subj_a = dynamic_slice(a, [subj * ntrials_subj], [ntrials_subj])
+    subj_z = dynamic_slice(z, [subj * ntrials_subj], [ntrials_subj])
+    subj_t = dynamic_slice(t, [subj * ntrials_subj], [ntrials_subj])
+    subj_theta = dynamic_slice(theta, [subj * ntrials_subj], [ntrials_subj])
+
+    subj_trial = dynamic_slice(trial, [subj * ntrials_subj], [ntrials_subj])
+    subj_response = dynamic_slice(response, [subj * ntrials_subj], [ntrials_subj])
+    subj_rt = dynamic_slice(rt, [subj * ntrials_subj], [ntrials_subj])
+    subj_feedback = dynamic_slice(feedback, [subj * ntrials_subj], [ntrials_subj])
```
cpaniaguam, line 73: Maybe build a factory to produce these dynamic slices based on the first argument? It seems like the second and third parameters are the same for all  of them.

src/hssm/likelihoods/rldm.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def logp(data, *dist_params) -> jnp.ndarray:
+        """Compute the log likelihood for the RLDM model.
+
+        Parameters
+        ----------
+        args : tuple
+            A tuple containing the subject index, number of trials per subject,
+            data, and model parameters
+            (rl_alpha, scaler, a, z, t, theta, trial, feedback).
```
cpaniaguam: Seems like the docstring is out-of-date.

src/hssm/likelihoods/rldm.py
```diff
@@ -0,0 +1,260 @@
...
+            t,
+            theta,
+            trial,
+            feedback,
+        )
+
+    return logp
+
+
+def make_vjp_logp_func(logp) -> Callable:
```
cpaniaguam: Type hint missing.

cpaniaguam (commented):
Couple little things.

src/hssm/likelihoods/rldm.py
digicosmos86: Yep, there is already a function in #753 that addresses this

src/hssm/likelihoods/rldm.py
digicosmos86: Yeah this does look cumbersome. However, a better optimization might be to allow this function to accept a matrix, so the slices can be done with one line, since this function is no longer the outmost function that has to conform to a signature. We'll definitely come back to this

src/hssm/likelihoods/rldm.py
digicosmos86: Will fix this

src/hssm/likelihoods/rldm.py
digicosmos86: Will fix this too

src/hssm/likelihoods/rldm.py
```diff
@@ -0,0 +1,262 @@
...
+
+import jax
+import jax.numpy as jnp
+from jax.lax import dynamic_slice, scan
+
+from ..distribution_utils.jax import make_jax_logp_ops
+from ..distribution_utils.onnx import make_jax_logp_funcs_from_onnx
+from ..utils import download_hf
+
+angle_onnx = download_hf("angle.onnx")
```
krishnbera, line 13: Can we extend this make this more general later? Users might add their own likeilhoods in `rldm.py` which will require use of other LAN onnx models. One route could be specifying in the HSSM `model_config` what ssm/LAN will be used. The corresponding onnx file can then be directly loaded here.

src/hssm/likelihoods/rldm.py
```diff
@@ -0,0 +1,262 @@
...
+# Obtain the angle log-likelihood function from an ONNX model.
+angle_logp_jax_func, _ = make_jax_logp_funcs_from_onnx(
+    model=angle_onnx,
+    params_is_reg=[True] * 5,
+    params_only=False,
+    return_jit=False,
+)
+
+
+def jax_lan_wrapper(input_matrix):
```
krishnbera, line 24: Information about the specific ssm/LAN (eg. from the model_config as mentioned above) used would make this implementation generic.

src/hssm/likelihoods/rldm.py
```diff
@@ -0,0 +1,262 @@
...
+from jax.lax import dynamic_slice, scan
+
+from ..distribution_utils.jax import make_jax_logp_ops
+from ..distribution_utils.onnx import make_jax_logp_funcs_from_onnx
+from ..utils import download_hf
+
+angle_onnx = download_hf("angle.onnx")
+
+# Obtain the angle log-likelihood function from an ONNX model.
+angle_logp_jax_func, _ = make_jax_logp_funcs_from_onnx(
```
krishnbera, line 16: Information about the specific ssm/LAN (eg. from the model_config as mentioned above) used would make this implementation generic.

src/hssm/likelihoods/rldm.py
digicosmos86: Yes this is just a proof of concept. I think ideally, we can use a decorator for this:

```python
@rldm_likelihood(model="xxx.onnx", n_participants=5, n_trials=200)
def my_awesome_likelihood_function():
    ...
```

We can do all the wrapping in the decorator

src/hssm/likelihoods/rldm.py
krishnbera: Yeah, what Paul suggests might be a good idea. But passing a matrix here might hide the specific details/arguments to the function.

src/hssm/likelihoods/rldm.py
```diff
@@ -0,0 +1,262 @@
...
+        process_trial, (q_val, 0.0, LAN_matrix_init, 0), trials
+    )
+
+    # forward pass through the LAN to compute log likelihoods
+    LL = jax_lan_wrapper(LAN_matrix)
+
+    return LL.ravel()
+
+
+rldm_logp_inner_func_vmapped = jax.vmap(
```
krishnbera, line 128: Information about the rlssm `model_config` would help make this function a generic implementation.

src/hssm/likelihoods/rldm.py
digicosmos86: Just to clarify everything in this file is just a proof-of-concept. We'll get back to the final implementation once we figure out how we can generalize this

krishnbera (approved):

PR #753, opened 2025-07-03
Workflow for incorporating JAX functions into HSSM

digicosmos86:
The main goal of this PR is to disentangle `make_jax_logp_funcs_from_jax_callable` from `make_jax_logp_funcs_from_onnx`, so that jax functions can be used with HSSM. There are quite some refactoring that happened to make this process more flexible. The main changes are:

1. Moved `make_jax_logp_funcs_from_jax_callable` from `onnx` to `jax` module under `distribution_utils`, and also renamed it to `make_jax_logp_funcs_from_callable`. It no longer calls `make_jax_logp_funcs_from_onnx` under the hood. Instead, it calls the following two atomic functions refactored from `make_jax_logp_funcs_from_onnx`.
2. Separated `make_vmap_func` and `make_vjp_func` from `make_jax_logp_funcs_from_onnx` into their own separate module under `func_utils`, so that these functions can be reusable.
3. Refactored out a `load_onnx_model` function that loads the onnx models, which is now reused in multiple functions.
4. A new `make_simple_jax_logp_funcs_from_onnx` function that directly accepts a matrix. Can come in handy when constructing RL likelihoods and reduce wrangling
5. Simplified RLDM likelihood in light of these changes
6. Tests for all these changes

Now HSSM can directly accept functions written in JAX, the integration is a bit weird though, just because the `analytical`, `approx_differentiable` and `blackbox` really does not provide us enough information about the likelihood functions themselves. Now, when `loglik_kind` is `approx_differentiable` and `backend` is `jax`, HSSM will help create a vectorized function with `vmap`, and when `loglik_kind` is analytical and `backend` is `jax`, HSSM will assume that the JAX function provided can handle the vectorization and skip the vmap part. This is weird and unintuitive, but probably the only solution given how we get information about the likelihood function from the users.


digicosmos86:
Note to @AlexanderFengler we might need to update the tutorial on incorporating a jax function into HSSM

src/hssm/distribution_utils/dist.py
```diff
@@ -694,7 +692,13 @@ def make_likelihood_callable(
         if loglik_kind == "analytical":
             if backend is None or backend == "pytensor":
                 return loglik
-            return make_blackbox_op(loglik)
+            logp, logp_grad, logp_nojit = make_jax_logp_funcs_from_callable(
+                loglik,
+                vmap=False,
+                params_only=False if params_only is None else params_only,
+            )
+            lan_logp_jax = make_jax_logp_ops(logp, logp_grad, logp_nojit)
```
cpaniaguam: To abstract away complexity: because these are not used individually, would a pattern like this be sensible:

```suggestion
            jax_logp_funcs= make_jax_logp_funcs_from_callable(
                loglik,
                vmap=False,
                params_only=False if params_only is None else params_only,
            )
            lan_logp_jax = make_jax_logp_ops(*jax_logp_funcs)
```

src/hssm/distribution_utils/dist.py
```diff
@@ -705,11 +709,12 @@ def make_likelihood_callable(
...
-                logp, logp_grad, logp_nojit = make_jax_logp_funcs_from_jax_callable(
+                logp, logp_grad, logp_nojit = make_jax_logp_funcs_from_callable(
                     loglik,
-                    params_is_reg,
+                    vmap=True,
+                    params_is_reg=params_is_reg,
                     params_only=False if params_only is None else params_only,
-                )  # type: ignore[assignment]
+                )
                 lan_logp_jax = make_jax_logp_ops(logp, logp_grad, logp_nojit)
```
cpaniaguam: Here is another instance of the pattern used for the `loglik_kind == "analytical"` case. Consider a little helper function and reuse it twice?

src/hssm/distribution_utils/func_utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+    vjp_vmap_logp = make_vjp_func(
+        vmap_logp,
+        params_only=params_only,
+    )
+
+    if return_jit:
+        return jit(vmap_logp), jit(vjp_vmap_logp), vmap_logp
+
+    return vmap_logp, vjp_vmap_logp
```
cpaniaguam, line 77: Maybe for later, make this function "type-stable."

src/hssm/distribution_utils/jax.py
```diff
@@ -289,3 +204,107 @@ def logp_op_dispatch(op, **kwargs):  # pylint: disable=W0612,W0613
...
+                in_axes=in_axes,
+                params_only=params_only,
+                return_jit=True,
+            )
+        return make_vmap_func(
+            logp,
+            in_axes=in_axes,
+            params_only=params_only,
+            return_jit=False,
+        )
```
cpaniaguam, line 300: I guess this is the silly part. What about adding an exception and get rid of the silliness?

```suggestion
        return make_vmap_func(
            logp,
            in_axes=in_axes,
            params_only=params_only,
            return_jit=return_jit, # type: ignore
        )
```

src/hssm/distribution_utils/onnx_utils/model.py
```diff
@@ -0,0 +1,76 @@
...
+        )
+
+    model_path = Path(model)
+    if model_path.exists():
+        return onnx.load(model_path)
+
+    if isinstance(model, PathLike):
+        raise FileNotFoundError(
+            f"The specified ONNX model file does not exist: {model}"
+        )
```
cpaniaguam, line 68: Why not let `onnx.load` handle this?

src/hssm/distribution_utils/dist.py
digicosmos86: I like this! No need to know what's inside the tuple here

src/hssm/distribution_utils/dist.py
digicosmos86: I see where you come from, but this is just basically two function calls, so maybe not worth a separate function

src/hssm/distribution_utils/func_utils.py
digicosmos86: This is already "type-stable" I think. The overloaded function definition directly associates the return type with the value of `return_jit`, so there is really no ambiguity here now

src/hssm/distribution_utils/jax.py
digicosmos86: If we get rid of this silliness then the type-checker won't know the the exact return type in relation to the value of `return_jit`, so this has to be left-in to make typing clear

src/hssm/distribution_utils/func_utils.py
cpaniaguam: Aren't three objects being returned in line 74 but two are returned in line 76? That would be two different types.

src/hssm/distribution_utils/onnx_utils/model.py
digicosmos86: Here there is still a case of when `model` is a string, and in this case, if the model does not exist locally, the function will also look for the model on Huggingface. This handles the case when the user specifically provided a `PathLike`, which means the user wants to load a local file

src/hssm/distribution_utils/dist.py
cpaniaguam: The calls can be reduced to a one liner.
```python
def _make_jax_logp_from_callable(loglik, params_only=None):
    """Helper to create JAX logp ops from a callable."""
    logp_funcs = make_jax_logp_funcs_from_callable(
        loglik,
        vmap=False,
        params_only=False if params_only is None else params_only,
    )
    lan_logp_jax = make_jax_logp_ops(*logp_funcs)
    return lan_logp_jax
```
Then lines 695-701 and 712-719 reduce to `return _make_jax_logp_from_callable(loglik, params_only=params_only)`.

src/hssm/distribution_utils/func_utils.py
digicosmos86: But there is no ambiguity which type is returned due to overloading, controlled by `return_jit`

src/hssm/distribution_utils/dist.py
digicosmos86: But you have to define the function, write docstrings for it, and after that, there might not be many lines saved 😀

src/hssm/distribution_utils/onnx.py
```diff
@@ -215,9 +148,7 @@ def make_pytensor_logp(
         The logp function that applies the ONNX model to data and returns the element-
         wise log-likelihoods.
     """
-    model_onnx: onnx.ModelProto = (
-        onnx.load(str(model)) if isinstance(model, (str, PathLike)) else model
-    )
+    model_onnx = load_onnx_model(model)
```
AlexanderFengler, line 151: I think we should rename the outer function here to `make_pytensor_logp_from_onnx`

src/hssm/distribution_utils/onnx.py
```diff
@@ -246,3 +177,45 @@ def logp(data: np.ndarray | None, *dist_params) -> ArrayLike:
         return pt.squeeze(pt_interpret_onnx(model_onnx.graph, inputs)[0])
 
     return logp
+
+
+def make_simple_jax_logp_funcs_from_onnx(
```
AlexanderFengler: @digicosmos86 don't fully understand the necessity of this function. 
It's just a subset of `make_jax_logp_funcs_from_onnx()`? 
If we keep it around, maybe we can change the name (and docstring) to explain better what the idea is and when to use it?

src/hssm/likelihoods/rldm.py
```diff
@@ -6,40 +6,17 @@
...
+from ..distribution_utils.jax import make_jax_logp_ops
+from ..distribution_utils.onnx import make_simple_jax_logp_funcs_from_onnx
 
 # Obtain the angle log-likelihood function from an ONNX model.
-angle_logp_jax_func, _ = make_jax_logp_funcs_from_onnx(
-    model=angle_onnx,
-    params_is_reg=[True] * 5,
-    params_only=False,
-    return_jit=False,
+angle_logp_jax_func = make_simple_jax_logp_funcs_from_onnx(
```
AlexanderFengler: let's rename `make_simple_jax_logp_fujncs_from_onnx()` to clarify the use case.

src/hssm/likelihoods/rldm.py
```diff
@@ -249,8 +181,8 @@ def make_rldm_logp_op(n_participants: int, n_trials: int) -> Callable:
     callable
         A function that computes the log likelihood for the RLDM model.
     """
-    logp = make_logp_func(n_participants, n_trials)
-    vjp_logp = make_vjp_logp_func(logp)
+    logp = make_rldm_logp_func(n_participants, n_trials)
+    vjp_logp = make_vjp_func(logp, params_only=False)
```
AlexanderFengler: possibly drop `params_only=False` since it's the default?

tests/distribution_utils/test_distribution_utils.py
```diff
@@ -248,3 +249,44 @@ def test_ensure_positive_ndt():
...
+            backend="jax",
+            params_only=False,
+        )
+
+        mock_make_jax_logp_funcs.assert_called_once_with(
+            mock_func,
+            vmap=True,
+            params_only=False,
+            params_is_reg=[False, False, False, False],
+        )
```
AlexanderFengler, line 292: two comments on this block:

1. slightly confused that no returns are stored for `distribution_utils.make_likelihood_callable()` 
2. do you need to enter the same context twice here?

tests/distribution_utils/test_func_utils.py
```diff
@@ -0,0 +1,72 @@
...
+    # The gradient of x*y w.r.t x is y, w.r.t y is x
+    assert np.allclose(grads[0], y * gz)
+    assert np.allclose(grads[1], x * gz)
+
+    def logp(x, y):
+        return x * y
+
+    # Test with params_only=True (should return all outputs)
+    vjp_func_params = make_vjp_func(logp, params_only=True)
+    grads_params = vjp_func_params(x, y, gz=gz)
```
AlexanderFengler, line 58: two things: 

1. defining two function with name `logp` is prob weird.
2. why are you using `vjp_func` above and not you are using `vjp_func_params` ?

src/hssm/distribution_utils/dist.py
```diff
@@ -694,7 +692,13 @@ def make_likelihood_callable(
         if loglik_kind == "analytical":
             if backend is None or backend == "pytensor":
                 return loglik
-            return make_blackbox_op(loglik)
+            logp_funcs = make_jax_logp_funcs_from_callable(
```
AlexanderFengler, line 695: Let's put on the agenda for Tuesday's team meeting to go through the branching logic of this. It was confusing before and it's still a bit hairy I find.

Would be great if we could simplify / make more readable.

AlexanderFengler (commented):
Thanks @digicosmos86, left a few comments.

src/hssm/distribution_utils/onnx.py
digicosmos86: Sounds good to me

src/hssm/distribution_utils/onnx.py
digicosmos86: This function exists so that this wrapper [here](https://github.com/lnccbrown/HSSM/blob/5fd70c14ab426ee6a9898946690fce52fe2b0de1/src/hssm/likelihoods/rldm.py#L24) is no longer needed. Basically, `make_jax_logp_funcs_from_onnx()` assumes that the resulting functions will be used directly as a likelihood function, so those functions will have the `logp(data, *dist_params)` signature. However, if we want to use LANs with RL likelihoods, then these functions can simply accept a matrix, just as what the LANs need, without wrappers or any data wrangling

src/hssm/likelihoods/rldm.py
digicosmos86: open to suggestions

src/hssm/likelihoods/rldm.py
digicosmos86: Might be beneficial to keep it around for clarity?

tests/distribution_utils/test_distribution_utils.py
digicosmos86: 1. The return value is mocked on line 277.
2. It might not be strictly necessary when the assertions are clearly different in this case, but it's a good practice to "reset" test context between different items. For example, it is possible that the second function is not called at all, but in the same context, the second `assert_called_once_with` check might still pass because the first function was called. Separate contexts ensures that the first call does not interfere with the second

tests/distribution_utils/test_func_utils.py
digicosmos86: These are separate tests focusing on making sure that the function works with `params_only` being both` True` and `False` for coverage

src/hssm/distribution_utils/dist.py
digicosmos86: Sounds good! I don't think there is a perfect solution to this because we just don't get enough information from the users to make good branching decisions, but we can at least clarify among ourselves what happens when

src/hssm/distribution_utils/onnx.py
AlexanderFengler: How about `make_jax_matrix_logp_funcs_from_onnx()`
Basically trying to find a name that hints at the actual usecase more than 'simple' (since arguably it's not simpler, just different).

src/hssm/likelihoods/rldm.py
AlexanderFengler: answered above

AlexanderFengler (approved):
ok left small suggestion re naming, otherwise should be good to go. 
approving already, thanks @digicosmos86.

src/hssm/distribution_utils/onnx.py
digicosmos86: Sounds good! Updating...

PR #756, opened 2025-07-09
Optimize RLDM likelihood function

digicosmos86:
This PR includes a few optimizations to the JAX likelihood function of RLDM to achieve a speed-up of ~20%. (266 it/s to 318 it/s in a rough comparison). The following are the optimizations:

1. Rewrote `process_trials` function into a pure function, that is: the old `process_trials` puts an increment index `t` into its `carry`-over parameter and in each iteration, it updates the entries of an external LAN matrix as a side-effect. However, this is not really how `scan` works. We can output `v` directly as a result of the computation in each iteration, and `scan` will collect the result as a new array. This avoids editing the LAN matrix array which saves time.

2. Simplified the computation of `v` through paneling. Since `n_participants` and `n_trials` are known, we can turn the input matrix into a 3-D array of dimension `(n_participants, n_trials, n_columns)`, and we can vmap a function over its first axis. This way we avoid lots of slicing operations in the vmapped function to achieve some speed-ups

3. Compute the likelihoods last. Since the inner loops really focuses only on computing `v`, we can compute the log-likelihoods with LAN once all `v`s are computed, so that we better take advantage of batching.

4. Eliminates two `extra_fields`. After paneling, we don't really need `participant_id`, `trial_id`, which are not really involved in likelihood computation other than providing some information about the shape of the data. We only have `feedback` as an `extra_field` now. In fact, `feedback` feels different from `participant_id` and `trial_id`, and if it's generalizable, maybe we can include it in `data` instead of `extra_fields`? @AlexanderFengler @krishnbera?

Hopefully the optimizations makes the generalization of RL likelihoods further.

@krishnbera quick note that this PR doesn't have your PR merged yet (since it's not yet merged into main)


src/hssm/distribution_utils/onnx.py
```diff
@@ -75,7 +75,7 @@ def make_jax_logp_funcs_from_onnx(
     model_onnx = load_onnx_model(model)
     scalars_only = all(not is_reg for is_reg in params_is_reg)
 
-    def logp(*inputs) -> jnp.ndarray:
+    def logp(*inputs) -> np.ndarray:
```
AlexanderFengler, line 78: is this true?

src/hssm/distribution_utils/onnx.py
```diff
@@ -199,7 +199,7 @@ def make_simple_jax_logp_funcs_from_onnx(
     """
     model_onnx = load_onnx_model(model)
 
-    def logp(input_matrix) -> jnp.ndarray:
+    def logp(input_matrix) -> np.ndarray:
```
AlexanderFengler, line 202: same here not sure if this is really a `np.array`?
It's a jax function no right?

src/hssm/likelihoods/rldm.py
```diff
@@ -17,96 +19,84 @@
...
+        A 2D jnp array containing the RL parameters (rl_alpha, scaler),
+        action (response), and reward (feedback) for the current trial.
+
+    Returns
+    -------
+    tuple
+        A tuple containing the updated q-values and the computed drift rate (v).
+    """
+    rl_alpha, scaler, action, reward = inputs
+    action = jnp.astype(action, jnp.int32)
```
AlexanderFengler, line 51: does it make sense to take care of the conversion somewhere earlier?

src/hssm/likelihoods/rldm.py
```diff
@@ -125,48 +115,53 @@ def make_rldm_logp_func(n_participants: int, n_trials: int) -> Callable:
         A function that computes the log likelihood for the RLDM model.
     """
 
-    def logp(data, *dist_params) -> jnp.ndarray:
+    def logp(data, *dist_params) -> np.ndarray:
```
AlexanderFengler, line 118: again np.ndarry?

src/hssm/likelihoods/rldm.py
```diff
@@ -125,48 +115,53 @@ def make_rldm_logp_func(n_participants: int, n_trials: int) -> Callable:
...
+            - a: boundary separation.
+            - z: starting point.
+            - t: non-decision time.
+            - theta: lapse rate.
+            - feedback: feedback for each trial.
 
         Returns
         -------
-        jnp.ndarray
+        np.ndarray
```
AlexanderFengler, line 138: flagging here too.

src/hssm/likelihoods/rldm.py
```diff
@@ -125,48 +115,53 @@ def make_rldm_logp_func(n_participants: int, n_trials: int) -> Callable:
...
             The log likelihoods for each subject.
         """
-        participant_id = dist_params[6]
-        trial = dist_params[7]
-        feedback = dist_params[8]
+        action = data[:, 1]
+        rl_alpha = dist_params[0]
```
AlexanderFengler, line 142: just to double check the logic. 
we know the ordering here, because we actively control that upstream?

tests/test_rldm_likelihood.py
```diff
@@ -67,18 +67,18 @@ def test_make_rldm_logp_op(fixture_path):
...
-    n_trials = trial.size // len(subj)
+    total_trials = trial.size
 
-    rl_alpha = np.ones(n_trials) * 0.60
-    scaler = np.ones(n_trials) * 3.2
-    a = np.ones(n_trials) * 1.2
-    z = np.ones(n_trials) * 0.1
-    t = np.ones(n_trials) * 0.1
-    theta = np.ones(n_trials) * 0.1
+    rl_alpha = np.ones(total_trials) * 0.60
```
AlexanderFengler, line 72: might have missed it, double checking why the size change to `total_trials` here.

src/hssm/likelihoods/rldm.py
krishnbera: can possibly do this in rlssm dataset sanity checks?

src/hssm/likelihoods/rldm.py
```diff
@@ -17,96 +19,84 @@
...
+    Parameters
+    ----------
+    subj_trials:
+        A jnp array of dimension (n_trials, 4) containing rl_alpha, scaler,
+        action (response), and reward (feedback) for each trial of the subject.
+
+    Returns
+    -------
+    jnp.ndarray
+        The log likelihoods for the RLDM model for the given subject.
```
krishnbera: This function returns the computed `v` and not the likelihoods, right?

krishnbera (commented):
overall looks good and the speed up is impressive. 
There is definitely a trade-off in terms of making the code readable while squeezing the efficiency. I'm thinking about it in the context of a user who is bringing a custom rlssm model for incoporating it into HSSM. I guess we'll have to provide them with the broader guidelines regarding how they should code up their model likelihood.

krishnbera (approved):

src/hssm/distribution_utils/onnx.py
digicosmos86: Yes. Most `jax.numpy` functions actually work on numpy arrays too (they are both just data stored contiguously in memory). Changing `jnp.ndarray` to `np.ndarray` gets around a lot of type checking error

src/hssm/likelihoods/rldm.py
digicosmos86: Unfortunately, when `inputs` is passed in as a 2D array, the dtype needs to be homogenous

src/hssm/likelihoods/rldm.py
digicosmos86: Again, this is just relaxing some type restrictions just to make mypy happy, but practically, this function indeed handles numpy arrays too

src/hssm/likelihoods/rldm.py
digicosmos86: same as above

tests/test_rldm_likelihood.py
digicosmos86: Had to change this because internally how the likelihood function works changed

src/hssm/likelihoods/rldm.py
digicosmos86: Yes, we stack the data into a matrix upstream so we know the ordering here

src/hssm/likelihoods/rldm.py
digicosmos86: Good catch! I'll update this

PR #771, opened 2025-07-13
drop two more notebboks

AlexanderFengler:


AlexanderFengler:
this should be stale given that the carney-workshop branch was merged.

PR #782, opened 2025-08-17
fix bug that mistakenly checks for positive ndts on missing data

AlexanderFengler:
This fixes an issue where we mistakenly check for `ndt - rt > 0` for missing data, which wrongly set the log likelihood to the lower bound in those cases.


src/hssm/distribution_utils/dist.py
```diff
@@ -773,28 +768,40 @@ def make_missing_data_callable(
...
+    #         raise ValueError(
+    #             "You have chosen `jax` as the backend "
+    #             + "for the missing data likelihood. "
+    #             + "However, you have not provided any values to `params_only`."
+    #         )
+
+    if params_is_reg is not None:
+        params_only = True if (not any(params_is_reg)) else False
+    else:
+        params_only = False
```
cpaniaguam:
```suggestion
    params_only = params_is_reg is not None and not any(params_is_reg)
```

src/hssm/data_validator.py
```diff
@@ -178,30 +178,35 @@ def _set_missing_data_and_deadline(
...
-        if not missing_data and not deadline:
-            return MissingDataNetwork.NONE
-
+        network = MissingDataNetwork.NONE
+        if not missing_data:
+            return network
         if missing_data and not deadline:
             network = MissingDataNetwork.CPN
-        elif not missing_data and deadline:
+        elif missing_data and deadline:
```
cpaniaguam, line 190: There is a change in logic here. Is it intentional?

src/hssm/hssm.py
```diff
@@ -1920,29 +1920,34 @@ def _make_model_distribution(self) -> type[pm.Distribution]:
...
                 missing_data_callable = make_missing_data_callable(
-                    self.loglik_missing_data, "jax", params_is_reg, params_only
+                    self.loglik_missing_data, "jax", params_is_reg, None
                 )
             else:
                 missing_data_callable = make_missing_data_callable(
                     self.loglik_missing_data,
                     self.model_config.backend,
+                    params_is_reg,
                     None,
```
cpaniaguam: Why not determine the backend first and keep a single call to `make_missing_data_callable`? It could look like this:
```python
backend = "jax" if self.model_config.backend != "pytensor" else self.model_config.backend
missing_data_callable = make_missing_data_callable(
    self.loglik_missing_data,
    backend,
    params_is_reg,
    None,
)
```

src/hssm/hssm.py
```diff
@@ -1920,29 +1920,34 @@ def _make_model_distribution(self) -> type[pm.Distribution]:
...
+                None,
                 has_deadline=self.deadline,
             )
 
+        if self.missing_data:
+            _logger.info(
+                f"Re-arranging data to put split missing and observed datapoints. "
+                f"Missing data (rt == {self.missing_data_value}) will be on top, "
+                f"observed datapoints follow."
+            )
```
cpaniaguam, line 1981:
```suggestion
            _logger.info(
                "Re-arranging data to put split missing and observed datapoints. "
                "Missing data (rt == %s) will be on top, observed datapoints follow.",
                self.missing_data_value
            )
```

cpaniaguam (changes requested):
Just a few observations to consider for readability and maintainability:
- Possibly refactor some conditionals and logic for clarity
- Could be worth adjusting logging to defer string formatting

src/hssm/param/regression_param.py
```diff
@@ -213,22 +213,28 @@ def make_safe_priors(
...
         """
+        # print("param name: ", self.name)
         safe_priors = {}
         dm = self._get_design_matrices(data, eval_env)
 
         get_prior = get_hddm_default_prior if is_ddm else get_default_prior
         specified_priors = (
             set(self.prior.keys()) if isinstance(self.prior, dict) else set()
         )
+        # print("specified_priors", specified_priors)
```
digicosmos86: We should probably remove these comments since it's done

tests/slow/test_mcmc.py
```diff
@@ -161,7 +105,7 @@ def test_lba_sampling():
 
 
 @pytest.mark.slow
-@pytest.mark.parametrize(parameter_names, parameter_grid)
+@pytest.mark.parametrize(PARAMETER_NAMES, PARAMETER_GRID)
```
digicosmos86, line 108: This is more a stylistic choice, but why we are using all caps for variables that are not constants?

tests/test_data_sanity.py
```diff
@@ -79,9 +79,9 @@ def test_deadline_missing_data_true(data_ddm, cpn):
     # Case 6: if deadline or missing_data is True, data should contain missing values
     with pytest.raises(
         ValueError,
-        match="You have no missing data in your dataset, "
-        + "which is not allowed when `missing_data` or `deadline` is set to "
-        + "True.",
+        match=r"Missing data is provided as True, "
```
digicosmos86: Maybe use `missing_data` to signify that it is an argument?

tests/test_data_sanity.py
```diff
@@ -90,20 +90,33 @@ def test_deadline_missing_data_true(data_ddm, cpn):
...
     data_ddm_missing = data_ddm.copy()
     data_ddm_missing.loc[missing_indices, "rt"] = -999.0
 
-    model = hssm.HSSM(
-        data=data_ddm_missing,
-        model="ddm",
-    )
+    with pytest.raises(
+        ValueError,
+        match="Missing data provided as False. \n"
```
digicosmos86, line 104: Same with this

tests/slow/test_mcmc.py
AlexanderFengler: hm from my perspective these are global constants, defined up top.
weird case.

PR #783, opened 2025-08-27
Refactor rng fn

cpaniaguam:
This pull request refactors and modularizes the logic for handling random sample generation and parameter processing in the `dist.py` utility functions. The main improvements include extracting logic into helper functions, improving code readability, and consolidating validation and preparation steps for simulation parameters.

Refactoring and modularization:

* Extracted size parsing from `args` and `kwargs` into a dedicated helper function `_extract_size` to standardize sample size handling.
* Moved parameter array creation and outlier handling into `_create_arg_arrays` and `_get_p_outlier` helper functions, reducing duplication and clarifying intent.
* Centralized random seed generation in `_get_seed`, improving reproducibility and code clarity.
* Consolidated parameter matrix preparation logic into `_prepare_theta_and_shape`, and sample replication calculation into `_calculate_n_replicas`, making the main function easier to follow and maintain. [[1]](diffhunk://#diff-9cda7cfbef005161ff85e52f17656d7c1331079d76d73ea3c759ab9da66fa07dR125-R261) [[2]](diffhunk://#diff-9cda7cfbef005161ff85e52f17656d7c1331079d76d73ea3c759ab9da66fa07dR415-R425)
* Added `_validate_size` to ensure sample size is compatible with input data, improving error handling and robustness.

These changes collectively improve maintainability, readability, and reliability of the sampling logic in the distribution utilities.

### DIFF
--- a/src/hssm/distribution_utils/dist.py
+++ b/src/hssm/distribution_utils/dist.py
@@ -125,10 +125,14 @@ def ensure_positive_ndt(data, logp, list_params, dist_params):
 def _extract_size(args, kwargs):
     """Extract size from args and kwargs.
 
-    Note
-    ----
-    This function has a side effect on `kwargs`. If "size" is present in `kwargs`,
-    it will be removed (popped) from the dictionary.
+    Returns
+    -------
+    size : int
+        The size of the random sample to generate.
+    args : tuple
+        The original arguments, with size removed if it was present.
+    kwargs : dict
+        The original keyword arguments, with size removed if it was present.
     """
     if "size" in kwargs:
         size = kwargs.pop("size")
@@ -139,7 +143,7 @@ def _extract_size(args, kwargs):
     if size is None:
         size = 1
 
-    return size, args
+    return size, args, kwargs
 
 
 def make_hssm_rv(
--- a/src/hssm/distribution_utils/dist.py
+++ b/src/hssm/distribution_utils/dist.py
@@ -146,6 +146,25 @@ def _extract_size(args, kwargs):
     return size, args, kwargs
 
 
+def _get_p_outlier(list_params, arg_arrays):
+    """Get p_outlier from arg_arrays and update arg_arrays."""
+    p_outlier = None
+    if list_params and list_params[-1] == "p_outlier":
+        p_outlier = arg_arrays.pop(-1)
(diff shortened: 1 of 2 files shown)


src/hssm/distribution_utils/dist.py
```diff
@@ -732,35 +932,37 @@ def make_likelihood_callable(
...
 
     # Other cases, when `loglik` is a string or a PathLike.
-    if loglik_kind != "approx_differentiable":
-        raise ValueError(
-            "You set `loglik_kind` to `approx_differentiable "
-            + "but did not provide a pm.Distribution, an Op, or a callable "
-            + "as `loglik`."
-        )
+    if isinstance(loglik, str) or isinstance(loglik, PathLike):
+        if loglik_kind != "approx_differentiable":
```
AlexanderFengler, line 936: just double checking here (seems like this was wrong before also).
The if statements tests for `loglik_kind != "approx..."`, then the `ValueError` claims that `loglik_kind` was actually set to `approx...`. 

Logic seems wrong.

src/hssm/distribution_utils/dist.py
```diff
@@ -350,20 +554,9 @@ def rng_fn(
...
-                        (*max_shape[:-1], max_shape[-1], obs_dim_int)
-                    )
-                else:
-                    # sims_out = sims_out.reshape(
-                    #     (*max_shape[:-1], max_shape[-1] * size, obs_dim_int)
-                    # )
-                    sims_out = sims_out.reshape(
-                        (*max_shape[:-1], max_shape[-1], n_replicas, obs_dim_int)
-                    )
+                shape_spec = _reshape_sims_out(max_shape, n_replicas, obs_dim_int)
```
AlexanderFengler, line 558: my sense here is that we could handle some of the shape stuff directly in `ssm-simulators` by allowing theta to arrive with different shapes in the original simulators, and then coming out respectively. 

(all the `n_replicas` logic could go in that case I think)

cpaniaguam:
>... but we might want to consider how we can host a bunch of the overall dimensionality logic directly in ssms with a simplifying wrapper that can handle more shapes of theta directly.

I agree. My goal with this PR is to clean this up, get a good understanding of what goes on, and then decide what pieces can be relocated.

PR #789, opened 2025-09-04
Relocate rng fn

cpaniaguam:
Move RV generation functionality to ssm-simulators. See https://github.com/lnccbrown/ssm-simulators/pull/210

Merging the current PR will require merging https://github.com/lnccbrown/ssm-simulators/pull/210 in that repo and a corresponding release.

### DIFF
--- a/tests/distribution_utils/test_distribution_utils.py
+++ b/tests/distribution_utils/test_distribution_utils.py
@@ -11,6 +11,9 @@ from hssm.distribution_utils.dist import (
     apply_param_bounds_to_loglik,
     make_distribution,
     ensure_positive_ndt,
+    _create_arg_arrays,
+    _extract_size,
+    _get_p_outlier,
 )
 from hssm.likelihoods.analytical import logp_ddm, DDM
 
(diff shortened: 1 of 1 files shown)


src/hssm/distribution_utils/dist.py
```diff
@@ -18,11 +17,19 @@
...
+    _calculate_n_replicas,
+    _create_arg_arrays,
+    _extract_size,
+    _get_p_outlier,
+    _get_seed,
+    _get_simulator_fun_internal,
+    _prepare_theta_and_shape,
+    _reshape_sims_out,
+    _validate_simulator_fun,
+)
```
digicosmos86, line 23: By convention it is not recommended to use private methods from another package. Is there a way to package all these methods in `ssm-simulators` and export a public method, instead of importing all these private methods?

digicosmos86 (changes requested):
This seem like a direct move of the same code to another package, but can we build something higher level within `ssm-simulators` instead of importing private methods?

src/hssm/distribution_utils/dist.py
cpaniaguam: I see your point and understand the risk when using private functions from a third-party package, but in this case we own the two packages. I guess we could just rename the functions so they are not "private" but doing this seems a bit silly to me. Another alternative could be to put these functions in a separate helper package.

digicosmos86: I definitely understand, but we shouldn't ignore good software engineering practices just because we manage both packages. However, I guess I can let this go if this takes too much time. My point actually is the latter - can we build a higher-level abstraction combining these functions, so we can export a public function from the ssms package?

cpaniaguam: I am adding a wrapper for `rng_fn` in https://github.com/lnccbrown/ssm-simulators/pull/219.

src/hssm/distribution_utils/dist.py
```diff
@@ -499,84 +174,29 @@ class HSSMRV(RandomVariable):
...
-            sims_out = simulator_fun_internal(
-                theta=theta,
-                random_state=seed,
-                n_replicas=n_replicas,
+            """Generate random variables from this distribution."""
+            sims_out = _rng_fn(
+                cls,
+                rng,
+                simulator_fun_internal,
+                _apply_lapse_model,
```
AlexanderFengler: one thought here @cpaniaguam ( @digicosmos86 for visibility )

I think overall it might make more sense to keep the whole lapse distribution logic out of this. 
That belongs much more naturally at the hssm level. 

The idea is that you can choose (in principle) any lapse distribution and mix it with any simulator upon construction of an HSSM model. I think there is no need to include this logic at the ssms level per se.

We just want the simulators from ssms work well with the shapes, they don't need to also take care of the lapse distribution. 

I am kind of willing to merge this and put this up as a todo for the next round of PRs, but preference would be:

1. separate out the lapse distribution logic here (by e.g. always dummy-passing no lapse_model to `rng_fn` already.
2. take that out of ssms upon the next release and drop here correspondingly.

Just let me know if I misunderstand something, it's possible of course.

AlexanderFengler (changes requested):
left few small comments.

src/hssm/distribution_utils/dist.py
cpaniaguam: I think we might need to clarify this in person but for now I will point out two things.

> 1. separate out the lapse distribution logic here (by e.g. always dummy-passing no lapse_model to rng_fn already.

The `_apply_lapse_model` function is defined in this very file within the HSSM package. The wrapper in ssm-simulator only includes the function as an argument. So in this way the lapse dist logic is not included in ssms.

> 2. take that out of ssms upon the next release and drop here correspondingly.

I think that's sensible! In the wrapper [`apply_lapse_model` is done at the very end](https://github.com/lnccbrown/ssm-simulators/blob/6ec78754d7e3c63d9f9785e33e5ecca03a28b096/ssms/hssm_support.py#L504), so it could be applied conditionally. The argument could have type `bmb.Prior | None` and do nothing if `None` is passed.

src/hssm/distribution_utils/dist.py
```diff
@@ -18,11 +17,13 @@
 import pytensor.tensor as pt
 from bambi.backend.utils import get_distribution_from_prior
 from pytensor.tensor.random.op import RandomVariable
-from ssms.basic_simulators.simulator import simulator
-from ssms.config import model_config as ssms_model_config
+from ssms.hssm_support import (
+    get_simulator_fun_internal,
+    validate_simulator_fun,
+)
+from ssms.hssm_support import rng_fn as _rng_fn
```
AlexanderFengler: this is a small nitpick, but I would rather rename this to something like `ssms_rng_fn`. Using `_rng_fn` is a bit generic / confusing.

src/hssm/distribution_utils/dist.py
```diff
@@ -499,84 +174,37 @@ class HSSMRV(RandomVariable):
...
-                _prepare_theta_and_shape(arg_arrays, size)
-            )
-            n_replicas = _calculate_n_replicas(is_all_args_scalar, size, new_data_size)
-
-            sims_out = simulator_fun_internal(
-                theta=theta,
-                random_state=seed,
-                n_replicas=n_replicas,
+            """Generate random variables from this distribution."""
+            sims_out, p_outlier = _rng_fn(
```
AlexanderFengler: Left the corresponding comment on the `ssms` side as well. 
I think `_rng_fn` shouldn't return `sims_out, p_outlier` but just `sims_out`, 
so `p_outlier` should be processed / split from the other args, before passing to `_rng_fn`. 

If you think what is done currently is necessary could you motivate / explain the choice?

AlexanderFengler (commented):
Thanks @cpaniaguam left some minor comments (and corresponding minor comments on the related newest ssms PR)

PR #797, opened 2025-09-11
Use `jaxonnxruntime` for onnx model conversion

digicosmos86:
`jaxonnxruntime` is a package maintained by Google that converts most onnx `op`s to JAX functions way beyond what we support right now. This PR creates a drop-in replacement to the `interpret_onnx` function that we had for JAX model conversion and all downstream tasks

### DIFF
--- a/src/hssm/distribution_utils/onnx.py
+++ b/src/hssm/distribution_utils/onnx.py
@@ -104,7 +104,7 @@ def make_jax_logp_funcs_from_onnx(
                 param_vector = param_vector.squeeze(axis=-1)
             input_vector = jnp.concatenate((param_vector, data))
 
-        return jax_func(input_vector)[0].squeeze()
+        return jax_func(input_vector)
 
     if params_only and scalars_only:
         logp_vec = lambda *inputs: logp(*inputs).reshape((1,))
@@ -218,6 +218,6 @@ def make_jax_matrix_logp_funcs_from_onnx(
         jnp.ndarray
             The element-wise log-likelihoods.
         """
-        return jax_func(input_matrix)[0].squeeze()
+        return jax_func(input_matrix)
 
     return logp
--- a/src/hssm/distribution_utils/onnx.py
+++ b/src/hssm/distribution_utils/onnx.py
@@ -17,8 +17,8 @@ from numpy.typing import ArrayLike
 from .._types import LogLikeFunc, LogLikeGrad
 from .jax import make_vmap_func
 from .onnx_utils.model import load_onnx_model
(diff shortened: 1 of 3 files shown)


src/hssm/distribution_utils/onnx.py
```diff
@@ -216,6 +218,6 @@ def logp(input_matrix) -> np.ndarray:
         jnp.ndarray
             The element-wise log-likelihoods.
         """
-        return interpret_onnx(model_onnx.graph, input_matrix)[0].squeeze()
+        return jax_func(input_matrix)[0].squeeze()
```
cpaniaguam: Is it sensible/possible to include the slicing + squeezing into `jax_func`? Seems like this is the standard workflow.

digicosmos86: great idea! Let me implement this

src/hssm/distribution_utils/onnx_utils/onnx2jax.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    model_graph = onnx_model.graph
+
+    # Get the input name and shape from the ONNX model to create a dummy input for
+    # initialization.
+    input_name = model_graph.input[0].name
+    input_dims = tuple(
+        dim.dim_value if (dim.dim_value > 0) else 1
+        for dim in model_graph.input[0].type.tensor_type.shape.dim
+    )
```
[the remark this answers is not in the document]
digicosmos86: We can assume that the model has an input

AlexanderFengler (commented):
lgtm, just double checking here. 
you tested this on a few LANs to verify outputs are exactly the same between our previous version and this?
asking because, tests passing might mask something on this front (e.g. if the resulting likelihood aren't exactly the same, MCMC might still sample without complains, not sure if we test for very specific outputs on the likelihoods anywhere in particular, beyond just "format").

AlexanderFengler:
> > lgtm, just double checking here. you tested this on a few LANs to verify outputs are exactly the same between our previous version and this? asking because, tests passing might mask something on this front (e.g. if the resulting likelihood aren't exactly the same, MCMC might still sample without complains, not sure if we test for very specific outputs on the likelihoods anywhere in particular, beyond just "format").
> 
> I did run a few LANs and the output seems correct to me. We do have tests that check that the JAX likelihoods and the Pytensor likelihoods output close enough results and gradients. We can't test on performance, though, so it might be a good idea for you to run a few LAN sampling sessions to compare performance. This package does introduce quite a few abstractions, which might have an impact on performance

ok cool thanks. 
I would be shocked if runtime is much affected to be honest.

PR #820, opened 2025-10-02
Improve setup env

AlexanderFengler:


cpaniaguam (approved):
I think it should be good. Thanks!

PR #824, opened 2025-10-09
Builder functions for assembling RL learning update rule functions and SSM likelihood functions

cpaniaguam:
## Purpose
Bridges RL dynamics with SSM likelihoods by computing trial-wise parameters and feeding them into an SSM log-likelihood in a vectorized, differentiable way.

**Key Abstractions**
- Annotated functions: `AnnotatedFunction` Protocol and `annotate_function` decorator carry `inputs`, `outputs`, and `computed` metadata to build dependency graphs between parameter calculators and likelihoods.
- Column resolution: Helpers (`_get_column_indices`, `_get_column_indices_with_computed`, `_collect_cols_arrays`) resolve sources of inputs from `data`, `list_params`, and `extra_fields`.

**RL Computation**
- Trial-wise update: `compute_param_trial_wise` maintains Q-values and produces the computed parameter using a TD-style rule, fully in JAX.
- Subject-wise vectorization: `compute_param_trial_wise` applies `jax.lax.scan` over trials to generate `v` per subject.

**Likelihood Builder**
- `make_rl_logp_func`: Detects computed dependencies, vmaps the computation across subjects, concatenates computed and direct params, and evaluates the provided SSM logp over participants/trials.

**Backend Integration**
- `make_rl_logp_op`: Wraps the composed logp into a PyTensor `Op`, adds VJP gradients via `make_vjp_func`, and JIT-compiles with JAX using `make_jax_logp_ops`.

**Design Highlights**
- Metadata-driven composition for clarity and extensibility.
- Efficient JAX primitives (`scan`, `vmap`, `jit`) for performance.
- Clean separation of data, model params, and computed params for hybrid RL+SSM models.

### DIFF
--- a/src/hssm/rl/likelihoods/two_armed_bandit.py
+++ b/src/hssm/rl/likelihoods/two_armed_bandit.py
@@ -1,0 +1,80 @@
+"""Two-armed bandit RLDM drift computation functions.
+
+These functions compute trial-wise and subject-wise drift rates and Q-value updates
+for a two-armed bandit reinforcement learning drift-diffusion model (RLDM).
+"""
+
+import jax.numpy as jnp
+from jax.lax import scan
+
+
+# Inner function to compute the drift rate and update q-values for each trial.
+# This function is used with `jax.lax.scan` to process each trial in the RLDM model.
+def compute_v_trial_wise(
+    q_val: jnp.ndarray, inputs: jnp.ndarray
+) -> tuple[jnp.ndarray, jnp.ndarray]:
+    """Compute the drift rate and updates the q-values for each trial.
+
+    This function is used with `jax.lax.scan` to process each trial. It takes the
+    current q-values and the RL parameters (rl_alpha, scaler), action (response),
+    and reward (feedback) for the current trial, computes the drift rate, and
+    updates the q-values. The q_values are updated in each iteration and carried
+    forward to the next one.
+
+    Parameters
+    ----------
+    q_val
+        A length-2 jnp array containing the current q-values for the two alternatives.
+        These values are updated in each iteration and carried forward to the next
+        trial.
+    inputs
+        A 2D jnp array containing the RL parameters (rl_alpha, scaler),
+        action (response), and reward (feedback) for the current trial.
+
+    Returns
+    -------
+    tuple
+        A tuple containing the updated q-values and the computed drift rate (v).
+    """
+    rl_alpha, scaler, action, reward = inputs
+    action = jnp.astype(action, jnp.int32)
+
+    # drift rate on each trial depends on difference in expected rewards for
+    # the two alternatives:
+    # drift rate = (q_up - q_low) * scaler where
+    # the scaler parameter describes the weight to put on the difference in
+    # q-values.
+    computed_v = (q_val[1] - q_val[0]) * scaler
+
+    # compute the reward prediction error
+    delta_RL = reward - q_val[action]
+
+    # update the q-values using the RL learning rule (here, simple TD rule)
+    q_val = q_val.at[action].set(q_val[action] + rl_alpha * delta_RL)
+
+    return q_val, computed_v
+
+
+def compute_v_subject_wise(
+    subj_trials: jnp.ndarray,
+) -> jnp.ndarray:
+    """Compute the drift rates (v) for a given subject.
+
+    Parameters
+    ----------
+    subj_trials:
+        A jnp array of dimension (n_trials, 4) containing rl_alpha, scaler,
+        action (response), and reward (feedback) for each trial of the subject.
+
+    Returns
+    -------
+    jnp.ndarray
+        The computed drift rates (v) for the RLDM model for the given subject.
+    """
+    _, v = scan(
+        compute_v_trial_wise,
+        jnp.ones(2) * 0.5,  # initial q-values for the two alternatives
+        subj_trials,
+    )
+
+    return v


src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,185 @@
...
+        jnp.ones(2) * 0.5,  # initial q-values for the two alternatives
+        subj_trials,
+    )
+
+    return v
+
+
+def make_rldm_logp_func(
+    mapping_function: Callable, n_participants: int, n_trials: int
+) -> Callable:
```
cpaniaguam: Add mapping function to function signature.

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,187 @@
...
+        # Reshape subj_trials into a 3D array of shape (n_participants, n_trials, 4)
+        # so we can act on this object with the vmapped version of the mapping function
+        subj_trials = jnp.stack((rl_alpha, scaler, action, feedback), axis=1).reshape(
+            n_participants, n_trials, -1
+        )
+
+        # Use the compute_v function to get the drift rates (v)
+        v = subject_wise_vmapped(subj_trials).reshape((-1, 1))
+
+        return v
```
cpaniaguam: Scope reduced (to drift rates?).

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,187 @@
...
+
+    return logp
+
+# TODO[CP]: Note really sure how to adapt this function given the changes to make_rldm_logp_func
+def make_rldm_logp_op(
```
cpaniaguam: @digicosmos86 Due to the scope reduction, should the rest of the workflow be added here?

tests/test_rldm_likelihood_abstraction.py
```diff
@@ -0,0 +1,59 @@
...
+            rl_alpha,
+            scaler,
+            a,
+            z,
+            t,
+            theta,
+            feedback,
+        )
+        assert jax_ll.shape[0] == total_trials
+        np.testing.assert_allclose(jax_ll.sum(), 7692.216, rtol=1e-2)
```
cpaniaguam: Expectations updated.

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,188 @@
...
+    _, v = scan(
+        compute_v_trial_wise,
+        jnp.ones(2) * 0.5,  # initial q-values for the two alternatives
+        subj_trials,
+    )
+
+    return v
+
+
+def make_rldm_logp_func(
```
digicosmos86: `RLDM` is a specific RL model. in order to make this more general, we should have a more general name (like `make_rl_logp_func`)

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,188 @@
...
+    mapping_function: Callable, n_participants: int, n_trials: int
```
digicosmos86: `mapping_function` can be slightly confusing (what is the mapping here?). How about `inner_func` or `subject_wise_func`?

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,188 @@
...
+    n_trials : int
+        Number of trials per participant.
+
+    Returns
+    -------
+    Callable
+        A function that computes drift rates (v) for all subjects given their trial data and RLDM parameters.
+    """
+    # Vectorized version of  mapping_function to handle multiple subjects.
+    subject_wise_vmapped = jax.vmap(mapping_function, in_axes=0)
```
digicosmos86: This is great!

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def logp(data, *dist_params) -> np.ndarray:
+        """Compute the log likelihood for the RLDM model.
```
digicosmos86: Again, it's no longer specific just for the RLDM model

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,188 @@
...
+
+        Returns
+        -------
+        np.ndarray
+            The log likelihoods for each subject.
+        """
+        action = data[:, 1]
+        rl_alpha = dist_params[0]
+        scaler = dist_params[1]
+        feedback = dist_params[-1]
```
digicosmos86: Here you'd be just stacking `data` and `dist_params` together to make a matrix

src/hssm/likelihoods/rldm_optimized_abstraction.py
digicosmos86: Yes, but no longer just `v`

src/hssm/likelihoods/rldm_optimized_abstraction.py
digicosmos86: There should be another function in-between. Let's focus on `make_rl_logp_func`

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,188 @@
+"""The log-likelihood function for the RLDM model."""
```
digicosmos86: I know you are probably just testing the water here, but eventually this needs to be in something like `likelihoods/rl/utils.py`

src/hssm/likelihoods/rldm_optimized_abstraction.py
cpaniaguam: Right, that's the next and more delicate change. I wanted to ensure everything was in place before tackling this.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: This will likely be the very last change in the PR.

src/hssm/likelihoods/rldm_optimized_abstraction.py
cpaniaguam: @digicosmos86 It seems like the order in which these are coming in might be important given how they are passed to `jnp.stack`. What's the most intuitive order? Perhaps require `(rl_alpha, scaler, action, feedback)` as in the current implementation?
```python
        subj_trials = jnp.stack((rl_alpha, scaler, action, feedback), axis=1).reshape(
            n_participants, n_trials, -1
        )
```

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,190 @@
...
+
+        Returns
+        -------
+        np.ndarray
+            The log likelihoods for each subject.
+        """
+        rl_alpha = dist_params[0]
+        scaler = dist_params[1]
+        action = data[:, 1]
+        feedback = dist_params[-1]
```
digicosmos86: I have some ideas about parameter look up in `dist_params` for different funcitons. For now, let's work on the most general case: this function just stacks data and every vector in `dist_params` and make a matrix. Is this enough info to unblock you for now?

cpaniaguam:
> @cpaniaguam @AlexanderFengler @krishnbera: about parameter lookup that we discussed today, here's the idea:
> 
> The inner function that `make_rl_logp_func` accepts will have this signature: `logp_subjwise_inner_func(matrix)` where `matrix` is a 2-D matrix with a subject-wise slice of data. We can make a function decorator for the inner function that looks like this:
> 
> ```python
> @rl(inputs=["response", "rl_alpha", "scaler", "feedback"], outputs=["v"])
> def logp_subjwise_inner_func(matrix)
>     ...
> ```
> 
> Then this makes it very explicit to everyone what the inputs and outputs of this function are. Inside the function decorator, we just attach the inputs and outputs arguments to the function itself:
> 
> ```python
> def rl(inner_func, inputs, outputs):
>     inner_func.inputs = inputs
>     inner_func.outputs = outputs
> 
>     return inner_func
> ```
> 
> Then we can access the information inside the `make_rl_logp_func` like this:
> 
> ```python
> def make_rl_logp_func(subject_wise, ...):
>     subjwise_inputs = subject_wise.inputs
>     subjwise_outputs = subject_wise.outputs
>     ...
> ```
> 
> Then we can perform look up based on the information. Potentially, this also enables us to accept multiple inner functions and compose them with the LAN functions, since we know exactly what the outputs are.
> 
> Let me know what you think

I like this idea of attaching the metadata to the inner function. I think it's sensible and elegant.

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,247 @@
...
+        # If data_cols is None but we have required parameters, raise early.
+        if dist_params or extra_fields:
+            raise ValueError("data_cols is None but required columns were provided.")
+        return
+    all_cols = [*dist_params, *extra_fields]
+    missing_cols = set(all_cols) - set(data_cols)
+    if missing_cols:
+        raise ValueError(
+            f"The following columns are missing from data_cols: {missing_cols}"
+        )
```
digicosmos86: I like the idea of validating that the columns exist, but if we are writing `make_rl_logp_func` as a pure function, we won't be able to validate the columns because we don't know what the columns are

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,247 @@
...
+        return
+    all_cols = [*dist_params, *extra_fields]
+    missing_cols = set(all_cols) - set(data_cols)
+    if missing_cols:
+        raise ValueError(
+            f"The following columns are missing from data_cols: {missing_cols}"
+        )
+
+
+def _get_column_indices(
```
digicosmos86, line 136: Please see my next comment. I think this is the right idea, but I think that, rather than just returning the indices, maybe this function should return the columns. The reason is that you are looking across two objects, the data object, and a tuple of `args`

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,247 @@
...
+        A function that computes drift rates (v) for all subjects given their trial data
+        and RLDM parameters.
+    """
+    _validate_columns(data_cols, dist_params, extra_fields)
+    idxs = _get_column_indices(data_cols, dist_params, extra_fields)
+
+    # Vectorized version of  subject_wise_func to handle multiple subjects.
+    subject_wise_vmapped = jax.vmap(subject_wise_func, in_axes=0)
+
+    def logp(data) -> np.ndarray:
```
digicosmos86: We can't change the signature of this function, unless we want to have a RL version of `make_distribution` in `distribution_utils`. Otherwise, like we discussed, this function will have to have `logp(data, *dist_params, *extra_fields)` signature

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,247 @@
...
+    dist_params_idxs = [col2idx[c] for c in (dist_params or [])]
+    extra_fields_idxs = [col2idx[c] for c in (extra_fields or [])]
+    return dist_params_idxs + extra_fields_idxs
+
+
+def make_rl_logp_func(
+    subject_wise_func: Callable,
+    n_participants: int,
+    n_trials: int,
+    data_cols: list[str] | None = None,
```
digicosmos86: Better to have `["rt", "response"] defaults since it's the case most of the time

src/hssm/likelihoods/rldm_optimized_abstraction.py
cpaniaguam: Maybe I'm missing something, but the intent of passing in `data_columns` is to capture what the columns in the data matrix represent.

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,247 @@
...
+    dist_params: list[str] | None = None,
```
digicosmos86: We probably shouldn't have a default as `None` here, since we are not replacing it within the function to have sensible defaults. Might be a good idea to make it required

digicosmos86 (commented):
Great! We are definitely on the right track here, but I still see some lingering confusion with the signature of `logp()` in `make_rl_logp_func`. As we discussed, PyMC requires it to have this signature `logp(data, *dist_params, *extra_fields)`. See [here](https://github.com/lnccbrown/HSSM/blob/e456e45667f83d3c635c86536d772121b4af95d9/src/hssm/distribution_utils/dist.py#L515) for how it is called

src/hssm/likelihoods/rldm_optimized_abstraction.py
digicosmos86: I think the misunderstanding is the idea that `data_cols = dist_params + extra_fields`. Actually, `data_cols` only means the columns in the `data` object, which is the only object that is passed in as a matrix

src/hssm/likelihoods/rldm_optimized_abstraction.py
cpaniaguam: The reason I changed it to this is because when building traces `jax.jit(logp)` will throw a type error if any of the arguments of the function passed are not array data types.

digicosmos86: It is safe to assume that all arguments passed to this function are arrays. Is it because some values passed to the jitted function in the tests are scalars? They should not be scalars

cpaniaguam: I'm thinking that `logp(data: np.array, dist_param:list[str], extra: list[str])`. In the constructor,
```python
def make_rl_logp_func(
    subject_wise_func: Callable,
    n_participants: int,
    n_trials: int,
    data_cols: list[str] | None = None, # column names in the data matrix passed to logp
    dist_params: list[str] | None = None, # dist_param column names
    extra_fields: list[str] | None = None, # extra_fields column names
)
```

With the last three arguments, we can do the lookup for the indices of the names in `dist_params` and `extra_fields` within `data_columns`. With this, one only needs the data matrix in `logp` so its signature can be simplified to `logp(data)` as all the information in `args` would be available in the scope of `make_rl_logp_func`.

digicosmos86: This is where the misunderstanding is. `data_cols` contains only the name of the columns in the `data` matrix (the first argument passed to `logp`, that typically has 'rt, response' columns. It is not to be confused with the compiled matrix that we pass to the inner function

src/hssm/likelihoods/rldm_optimized_abstraction.py
cpaniaguam: The problem there is that one can't follow an argument with a default value with a required argument. Their order would need to be changed.

digicosmos86: The order is not super important for these arguments here, but if we want to keep the order, we can handle the `None` inside the function

cpaniaguam: There is handling of `None` defaults in `_validate_columns`.

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,276 @@
...
+from jax.lax import scan
+from pytensor.graph import Op
+
+from hssm.distribution_utils.func_utils import make_vjp_func
+
+from ..distribution_utils.jax import make_jax_logp_ops
+from ..distribution_utils.onnx import make_jax_matrix_logp_funcs_from_onnx
+
+# Obtain the angle log-likelihood function from an ONNX model.
+angle_logp_jax_func = make_jax_matrix_logp_funcs_from_onnx(
```
krishnbera: loading the angle LAN was hardcoded since the existing likelihood used angle. But we will need to make this generic (based on the decision process defined in the rlssm model_config) and load the correct LAN based on what the user wants.

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,276 @@
...
+        and RLDM parameters.
+    """
+    inputs = subject_wise_func.inputs  # type: ignore[attr-defined]
+    _validate_columns(data_cols, inputs)
+    idxs = _get_column_indices(data_cols, inputs)
+
+    # Vectorized version of  subject_wise_func to handle multiple subjects.
+    subject_wise_vmapped = jax.vmap(subject_wise_func, in_axes=0)
+
+    def logp(data, *args) -> np.ndarray:
```
krishnbera: this function name should be updated since it not computing the logp?

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,276 @@
...
+        The number of participants in the dataset.
+    n_trials : int
+        The number of trials per participant.
+
+    Returns
+    -------
+    Op
+        A pytensor Op that computes the log likelihood for the RLDM model.
+    """
+    logp = make_rl_logp_func(subject_wise_func, n_participants, n_trials)
```
krishnbera: Is this function meant to output the logp or drifts (v) only? I am assuming logp, right?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,276 @@
...
+from ..distribution_utils.jax import make_jax_logp_ops
+from ..distribution_utils.onnx import make_jax_matrix_logp_funcs_from_onnx
+
+# Obtain the angle log-likelihood function from an ONNX model.
+angle_logp_jax_func = make_jax_matrix_logp_funcs_from_onnx(
+    model="angle.onnx",
+)
+
+
+def annotate_function(**kwargs):
```
AlexanderFengler: This has parallels with the `decorate_simulator` decorator? 
Can we reuse functionality (either base that one on this one or vice versa)?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,276 @@
...
+        for key, value in kwargs.items():
+            setattr(wrapper, key, value)
+        return wrapper
+
+    return decorator
+
+
+# Inner function to compute the drift rate and update q-values for each trial.
+# This function is used with `jax.lax.scan` to process each trial in the RLDM model.
+def compute_v_trial_wise(
```
AlexanderFengler: similar to the `angle` in the name, here we shouldn't hardcode `_v_` in the function name. 

The learning rule could be applied to any parameter, we should use a name for it that just reflects the type of updating we are doing, regardless of which parameter it is applied to.

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,276 @@
...
+    q_val = q_val.at[action].set(q_val[action] + rl_alpha * delta_RL)
+
+    return q_val, computed_v
+
+
+# This function computes the drift rates (v) for each subject by processing
+# their trials one by one. It uses `jax.lax.scan` to efficiently iterate over
+# the trials and compute the drift rates based on the RL parameters, actions,
+# and rewards for each trial.
+def compute_v_subject_wise(
```
AlexanderFengler: would take out `_v_` here again in the name, and also instead of `subject_wise` (which insinuates that it is done per subject but FOR EACH subject) rather `_one_subject`.

src/hssm/likelihoods/rldm_optimized_abstraction.py
AlexanderFengler: more or less the following applies: 

`likelihood(data_cols|dist_params;extra_cols)`

So data-cols refer to observables, `dist_params` are estimable parameters (even if we fix them in a given model) and extra_cols is some extra fixed set of covariates that affects the internal computation of the likelihood.

In the rl case `extra_fields` are the feedback vectors which are used in the scan to update `v` (a `dist_param`).

src/hssm/likelihoods/rldm_optimized_abstraction.py
```diff
@@ -0,0 +1,276 @@
...
+def _get_column_indices(
+    data_cols: list[str] | None,
+    cols_to_look_up: list[str] | None = None,
+) -> list[int]:
+    """Return indices for required columns.
+
+    When data_cols is None, return an empty list so that callers can defer
+    indexing until data is available.
+    """
+    if data_cols is None:
```
AlexanderFengler: `data_cols` should never be `None` or am I missing something?

src/hssm/rl/likelihoods/builder.py
AlexanderFengler: @cpaniaguam this is basically just the update rule for the target parameter of the reinforcement learning process (`v` in our example model, but again, doesn't have to be that).

src/hssm/likelihoods/rldm_optimized_abstraction.py
AlexanderFengler: @cpaniaguam / @krishnbera  let's put a meeting on the calendar for monday to discuss this point? ( @digicosmos86 also if you are back).

src/hssm/rl/likelihoods/builder.py
cpaniaguam: @krishnbera @AlexanderFengler Thanks for pointing this out. We intentionally removed a couple of steps from the true logp computation which we intend to reintroduce in a different function.

src/hssm/likelihoods/rldm_optimized_abstraction.py
cpaniaguam: Indeed, we are aware this is computing drift rates only and some renaming of variables is necessary. The main goal of this PR is to build the abstraction and then ingest the drifts to get logp in a separate step.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: The `annotate_function` decorator is more generic than `decorate_atomic_simulator` as the latter attaches specific metadata
```python
    model_name: str,
    choices: list | np.ndarray | None = None,
    obs_dim: int = 2,  # At least for now ssms models all fall under 2 obs dims
```
We could refactor `decorate_atomic_simulator` using a call to `annotate_function`:
```python
def decorate_atomic_simulator(
    model_name: str,
    choices: list | np.ndarray | None = None,
    obs_dim: int = 2,  # At least for now ssms models all fall under 2 obs dims
):
    choices = [-1, 1] if choices is None else choices

    return annotate_function(
        model_name=model_name,
        choices=choices,
        obs_dim=obs_dim,
    )
```

src/hssm/likelihoods/rldm_optimized_abstraction.py
cpaniaguam: Good eye! `_get_column_indices` is called after a validation check on `data_cols `. Thanks for catching that!

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Thanks @krishnbera. This is another element to abstract as well.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: And this change should probably be introduced in a separate PR.

src/hssm/likelihoods/rldm_optimized_abstraction.py
digicosmos86: @AlexanderFengler @krishnbera this is still WIP so none of the code you see is in a final reviewable state. @cpaniaguam can you change this PR to a draft so we can iterate more on it?

tests/rl/test_rl_likelihood_builder.py
```diff
@@ -0,0 +1,256 @@
...
+        data = np.array([1, 2]) * np.ones((10, 2))
+
+        # Create mock parameter arrays: each parameter gets a different constant value
+        # _args[0] = [3, 3, 3, ...] (rl.alpha), _args[1] = [4, 4, 4, ...] (scaler), etc.
+        _args = [
+            np.ones(10) * i for i, _ in enumerate(list_params_extra_fields, start=3)
+        ]
+
+        ssm_logp_func = angle_logp_jax_func
+        ssm_logp_func.inputs = ["v", "a", "z", "t", "theta", "rt", "response"]
```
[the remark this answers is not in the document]
cpaniaguam: This is fine.

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,381 @@
...
+from ..distribution_utils.onnx import make_jax_matrix_logp_funcs_from_onnx
+
+
+class AnnotatedFunction(Protocol):
+    """A protocol for functions annotated with metadata."""
+
+    inputs: list[str]
+    outputs: list[str]
+    computed: dict[str, "AnnotatedFunction"]
+
```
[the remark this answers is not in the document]
cpaniaguam: This is for type-checking.

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,381 @@
...
+        vmapped_func = jax.vmap(ssm_logp_func.computed["v"], in_axes=0)
+        computed_arg = vmapped_func(subj_trials)
+        computed_arg = computed_arg.reshape((-1, 1))
+
+        non_computed_args = _collect_cols_arrays(
+            data, args, ssm_logp_func_colidxs.colidxs
+        )
+        # create parameter arrays to be passed to the likelihood function
+        ddm_params_matrix = jnp.stack(non_computed_args, axis=1)
+        lan_matrix = jnp.concatenate((computed_arg, ddm_params_matrix), axis=1)
```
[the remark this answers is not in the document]
cpaniaguam: This will be addressed later as suggested by the TODO note.

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,381 @@
...
+        # so we can act on this object with the vmapped version of the mapping function
+        # TODO: Generalize to handle every member in computed
+        # TODO: Move to helper function.
+        # Get column indices for SSM logp function
+        ssm_logp_func_colidxs = _get_column_indices_with_computed(
+            ssm_logp_func,
+            data_cols,
+            list_params,
+            extra_fields,
+        )
```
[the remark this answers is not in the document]
cpaniaguam: That's okay for now.

src/hssm/rl/likelihoods/builder.py
digicosmos86, line 1: I think the naming of this file is misleading. This file should only contain general function for RL. Maybe we should create a new RL directory and put all RL related stuff there?

cpaniaguam: I think that's very sensible. Choosing good names is sometimes tricky -- do you have any suggestions?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,491 @@
...
+    inputs: list[str]
+    outputs: list[str]
+    computed: dict[str, "AnnotatedFunction"]
+
+    # Added to satisfy static type checkers
+    def __call__(self, *args: Any, **kwargs: Any) -> Any: ...  # noqa: D102
+
+
+# Obtain the angle log-likelihood function from an ONNX model.
+angle_logp_jax_func = make_jax_matrix_logp_funcs_from_onnx(
```
digicosmos86: This should not be here any more. We should make a rldm.py with the likelihood for RLDM built with the utilities in this file

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,491 @@
...
+        for key, value in kwargs.items():
+            setattr(wrapper, key, value)
+        return wrapper
+
+    return decorator
+
+
+# Inner function to compute the drift rate and update q-values for each trial.
+# This function is used with `jax.lax.scan` to process each trial in the RLDM model.
+def compute_v_trial_wise(
```
digicosmos86: Same with this. Should be in the file that contains the log likelihood of rldm.py

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,491 @@
...
+import numpy as np
+from jax.lax import scan
+from pytensor.graph import Op
+
+from hssm.distribution_utils.func_utils import make_vjp_func
+from hssm.distribution_utils.jax import make_jax_logp_ops
+from hssm.distribution_utils.onnx import make_jax_matrix_logp_funcs_from_onnx
+
+
+class AnnotatedFunction(Protocol):
```
digicosmos86, line 53: Now that we want to be type-safe, we should make protocols for learning process and decision process functions with different metatdata requirements

src/hssm/rl/likelihoods/builder.py
cpaniaguam: I will delete this line now.

> We should make a rldm.py with the likelihood for RLDM built with the utilities in this file

Can we do this in a separate PR?

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Is the file you are referring to `src/hssm/likelihoods/rldm.py`?

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Can this go in a separate PR?

tests/rl/test_rl_likelihood_builder.py
```diff
@@ -0,0 +1,479 @@
...
```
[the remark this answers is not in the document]
cpaniaguam: Removed (no longer needed)

src/hssm/rl/likelihoods/builder.py
digicosmos86: It can be a file in the same directory named rldm.py

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,478 @@
...
+            colidxs[col] = ("args", list_params_extra_fields.index(col))
+        else:
+            raise ValueError(
+                f"Column '{col}' not found in any of `data`, `list_params`, "
+                f"or `extra_fields`."
+            )
+    return colidxs
+
+
+@dataclass
```
digicosmos86, line 201: Is `dataclass` an overkill here? How about a `NamedTuple`?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,478 @@
...
+
+    inputs: list[str]
+    outputs: list[str]
+    computed: dict[str, "AnnotatedFunction"]
+
+    # Added to satisfy static type checkers
+    def __call__(self, *args: Any, **kwargs: Any) -> Any: ...  # noqa: D102
+
+
+def annotate_function(**kwargs):
```
digicosmos86: This function feels more generic and maybe could go to a separate utils file?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,478 @@
...
+        ssm_logp_func_colidxs = _get_column_indices_with_computed(
+            ssm_logp_func,
+            data_cols,
+            list_params,
+            extra_fields,
+        )
+        computed_colidxs1 = _get_column_indices(
+            ssm_logp_func.computed["v"].inputs, data_cols, list_params, extra_fields
+        )
+        computed_colidxs1_data = _collect_cols_arrays(data, args, computed_colidxs1)
```
digicosmos86: can the variable names be a bit more informative?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,478 @@
...
+        subj_trials = jnp.stack(computed_colidxs1_data, axis=1)
+        subj_trials = subj_trials.reshape(n_participants, n_trials, -1)
+        vmapped_func = jax.vmap(ssm_logp_func.computed["v"], in_axes=0)
```
digicosmos86: This should be outside this inner function. Otherwise a `vmapped_func` is created every time this function is called

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,478 @@
...
+        computed_arg = vmapped_func(subj_trials)
+        computed_arg = computed_arg.reshape((-1, 1))
```
digicosmos86: Can these lines be combined?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,478 @@
...
+            data_cols,
+            list_params,
+            extra_fields,
+        )
+        computed_colidxs1 = _get_column_indices(
+            ssm_logp_func.computed["v"].inputs, data_cols, list_params, extra_fields
+        )
+        computed_colidxs1_data = _collect_cols_arrays(data, args, computed_colidxs1)
+        subj_trials = jnp.stack(computed_colidxs1_data, axis=1)
+        subj_trials = subj_trials.reshape(n_participants, n_trials, -1)
```
digicosmos86: Can these lines be combined?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,478 @@
...
+    Factory function that builds a complete log-likelihood for sequential sampling
+    models where some parameters are computed by other models (e.g., drift rates
+    computed from reinforcement learning).
+
+    Parameters
+    ----------
+    ssm_logp_func : AnnotatedFunction
+        Log-likelihood function for the sequential sampling model, decorated with
+        `@annotate_function`. Must have `.inputs`, `.outputs`, and `.computed`
+        attributes specifying parameter dependencies.
```
digicosmos86: Let's also clarify that this function here is a non-jitted JAX function

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,478 @@
...
+    Factory function that builds a complete log-likelihood Op for reinforcement learning
+    models where some parameters are computed by other models (e.g., drift rates
+    computed from reinforcement learning).
+
+    Parameters
+    ----------
+    ssm_logp_func : AnnotatedFunction
+        Log-likelihood function for the sequential sampling model, decorated with
+        `@annotate_function`. It must have `.inputs`, `.outputs`, and `.computed`
+        attributes specifying parameter dependencies.
```
digicosmos86: Let's also clarify that this function here is a non-jitted JAX function

tests/rl/test_rl_likelihood_builder.py
```diff
@@ -0,0 +1,477 @@
...
+        @annotate_function(inputs=["input1", "input2"], outputs=["output1"], other=42)
+        def sample_function():
+            return
+
+        assert sample_function.inputs == ["input1", "input2"]
+        assert sample_function.outputs == ["output1"]
+        assert sample_function.other == 42
+
+
+class TestRldmLikelihoodAbstraction:
```
digicosmos86: For clarity, maybe something like `TestRldmLikelihoodBuilder`? The word `abstraction` here is very abstract :)

src/hssm/rl/likelihoods/builder.py
cpaniaguam: I prefer  `dataclass` over the old way of defining namedtuples for type-hinting purposes. With the newer way, inheriting from `typing.NamedTuple`, I guess it's a matter of taste.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: I moved the `annotate_function` decorator to `hssm.utils`.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: These are the column indices for the first computed parameter and the associated data columns from the data matrix. Can you suggest one or two alternatives?

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Thanks!

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Yes, they can. However, I see two potential issues:

- the line would do more than one thing (JAX vectorizacion and a separate independent reshaping operation)
- it'd be more difficult to debug

I prefer the current approach for its clarity and separation of concerns. Thoughts?

src/hssm/rl/likelihoods/builder.py
cpaniaguam: See previous discussion.

src/hssm/rl/likelihoods/builder.py
digicosmos86: How about ssm_logp_input_indices and ssm_logp_data?

cpaniaguam: I like it!

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Done.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Done.

tests/rl/test_rl_likelihood_builder.py
cpaniaguam: Done!

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,548 @@
...
+    # Convert back to lists maintaining original order
+    found_inputs = [inp for inp in inputs if inp in found_set]
+    computed = [inp for inp in inputs if inp in computed_set]
+
+    colidxs = _get_column_indices(found_inputs, data_cols, list_params, extra_fields)
+
+    return ColumnLookupResult(colidxs=colidxs, computed=computed)
+
+
+def _collect_cols_arrays(data, _args, colidxs):
```
AlexanderFengler: typing info missing

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,548 @@
...
+        """Extract and reshape data for a computation function."""
+        # Get column indices for inputs needed by computation function
+        colidxs = _get_column_indices(
+            compute_func.inputs, data_cols, list_params, extra_fields
+        )
+        # Extract and organize data for this computation
+        cols_data = _collect_cols_arrays(data, args, colidxs)
+        # Reshape into 3D array (n_participants, n_trials, n_inputs)
+        subj_trials = jnp.stack(cols_data, axis=1)
+        return subj_trials.reshape(n_participants, n_trials, -1)
```
AlexanderFengler, line 586: here just double-checking. Are we guaranteed that the reshape is applied as intended here every-time? (as in a priori column order in `subj_trials` guaranteed to make sense for the reshaping logic)

src/hssm/rl/likelihoods/builder.py
```diff
@@ -0,0 +1,548 @@
...
+            # Find its position in non_computed_args based on colidxs order
+            colidxs_keys = list(ssm_logp_func_colidxs.colidxs.keys())
+            idx = colidxs_keys.index(param_name)
+            return non_computed_args[idx].reshape((-1, 1))
+
+        param_arrays = [get_param_array(name) for name in ssm_logp_func.inputs]
+
+        # Stack all parameters into final matrix
+        lan_matrix = jnp.concatenate(param_arrays, axis=1)
+        return ssm_logp_func(lan_matrix)
```
AlexanderFengler, line 676: ok just one structural question here: 

This assumes the inputs to the `ssm_logp_fun` as a matrix, so implies that `ssm_logp_fun` is a LAN (or has that structure). 

We should make sure this also works with our `blackbox` and `analytical` likelihoods, which may demand another version of the outer function here.

pyproject.toml
```diff
@@ -197,6 +197,21 @@ convention = "numpy"
...
+exclude = ["pytensor"]
+follow_imports = "skip"
+namespace_packages = true
+
+[[tool.mypy.overrides]]
+module = "pytensor.*"
+ignore_errors = true
+
+[[tool.mypy.overrides]]
+# TODO [CP]: Tighten types in `src/hssm/plotting/model_cartoon.py` and
```
AlexanderFengler: this should get associated issues?

pyproject.toml
```diff
@@ -197,6 +197,21 @@ convention = "numpy"
 
 [tool.mypy]
 ignore_missing_imports = true
+exclude = ["pytensor"]
```
AlexanderFengler: why do we actually need to exclude "pytensor" here?

cpaniaguam: There was new bug from mypy apparently due to a update. A fix was added to main in #863 so changes to this file are not longer needed.

pyproject.toml
cpaniaguam: Resolved in #863.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: To be addressed in https://github.com/lnccbrown/HSSM/issues/865. Thanks!

src/hssm/rl/likelihoods/builder.py
krishnbera: Yes, better to have a separation of concerns here. Since this function is used to get the likelihoods, we can put this under `src/hssm/likelihoods/rlssm.py`.

src/hssm/rl/likelihoods/builder.py
krishnbera: this function can also be moved to `src/hssm/likelihoods/rlssm.py`.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: It's not totally clear to me how to make this more generic. Moving it to an issue.

cpaniaguam: https://github.com/lnccbrown/HSSM/issues/871

src/hssm/rl/likelihoods/builder.py
cpaniaguam: https://github.com/lnccbrown/HSSM/issues/872

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Yes, the order is deterministic. I added some more comments documenting the behavior.

src/hssm/rl/likelihoods/builder.py
krishnbera: We should move this function to another file - to keep the model-specific functions separate from the general machinery to construct RLSSM likelihood (such as the `AnnotatedFunction` class, etc. `builder.py` could be made clean of all model-specific/likelihood-specific functions?

Issue #801, opened 2025-09-12
New Analytical Models: Poisson Race Model


AlexanderFengler:
Source e.g.:

(Poisson Race model)[https://link.springer.com/article/10.3758/BF03212980]

PR #840, opened 2025-10-28
initial poisson race implementation

hayden-johnson:
## Issue #801

This PR introduces a preliminary implementation of the Poisson race model as described in Appendix A of:

[A comparison of two response time models applied to perceptual matching](https://link.springer.com/article/10.3758/BF03212980)

Key Differences from the original formulation

Stimulus-dependent rate parameters are not implemented. This would require trial-by-trial stimulus identifiers for log-probability evaluation.

Shape parameter generalization: we do not restrict the shape parameter to be an integer. This simplifies the implementation but requires evaluation of the incomplete gamma function using scipy, rather than using factorials.

### DIFF
--- a/src/hssm/likelihoods/analytical.py
+++ b/src/hssm/likelihoods/analytical.py
@@ -631,7 +631,7 @@ def logp_poisson_race(
     log_pdf = (
         k_c * pt.log(r_c_safe)
         + (k_c - 1.0) * pt.log(rt_safe)
-        - r_c * rt
+        - r_c * rt_safe
         - pt.gammaln(k_c)
     )
 
@@ -651,7 +651,6 @@ def logp_poisson_race(
 
 
 # set bounds
-poisson_race_params = ["r1", "r2", "k1", "k2", "t"]
 poisson_race_bounds = {
     "r1": (np.finfo(float).eps, np.inf),
     "r2": (np.finfo(float).eps, np.inf),
@@ -659,6 +658,7 @@ poisson_race_bounds = {
     "k2": (np.finfo(float).eps, np.inf),
     "t": (0.0, np.inf),
 }
+poisson_race_params = list(poisson_race_bounds)
 
 # build distribution
 POISSON_RACE = make_distribution(
--- a/tests/test_likelihoods_poisson_race.py
+++ b/tests/test_likelihoods_poisson_race.py
@@ -1,11 +1,14 @@
 """Unit tests for the Poisson race likelihood."""
 
 import numpy as np
+import pandas as pd
 import pymc as pm
 import pytest
+from ssms.config import model_config
 
 import hssm
 from hssm.likelihoods.analytical import LOGP_LB, logp_poisson_race
+from hssm.simulator import simulate_data
 
 hssm.set_floatX("float32")
 
@@ -33,7 +36,7 @@ def poisson_race_data():
         [
             (0.45, 1.0),
             (0.55, -1.0),
-            (0.65, 0.0),
+            (0.65, -1.0),
             (0.75, 1.0),
         ],
         dtype="float32",
@@ -60,7 +63,7 @@ def test_poisson_race_matches_exponential_case():
         [
             (0.4, 1.0),
             (0.5, -1.0),
-            (0.8, 0.0),
+            (0.8, -1.0),
             (0.9, 1.0),
         ],
         dtype="float32",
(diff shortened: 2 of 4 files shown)


src/hssm/likelihoods/analytical.py
```diff
@@ -14,6 +14,9 @@
 from numpy import inf
 from pymc.distributions.dist_math import check_parameters
 
+# import gammaln and gammaincc from scipy.special
+from scipy.special import gammaincc, gammaln
```
AlexanderFengler: is this actually used?

src/hssm/likelihoods/analytical.py
```diff
@@ -559,3 +562,101 @@ def logp_lba3(
...
+}
+poisson_race_params = ["r1", "r2", "k1", "k2", "t"]
+
+# build distribution
+POISSON_RACE = make_distribution(
+    rv="poisson_race",
+    loglik=logp_poisson_race,
+    list_params=poisson_race_params,
+    bounds=poisson_race_bounds,
+)
```
cpaniaguam: Is this distribution intended to be exposed in the public API? If so, it should also be exported in `src/hssm/likelihoods/__init__.py`.

hayden-johnson: Thanks! Added it now

src/hssm/likelihoods/analytical.py
hayden-johnson: Yes, currently used for the logp evaluation. Is it alright to use scipy here? Or would it be preferable not to introduce the dependecy?

AlexanderFengler: I guess one problem with using `scipy` functions on that front is that we won't be able to autodiff through the logp. 

If there is a pytensor / jax equivalent we would prefer it for that reason. 

If we need to use this, we will have to go the "blackbox" route on the likelihood implementation.

tests/test_likelihoods_poisson_race.py
```diff
@@ -0,0 +1,105 @@
...
+
+    logp = logp_poisson_race(data, **theta).eval()
+
+    expected = []
+    for rt, response in data:
+        winner_rate = theta["r2"] if response > 0 else theta["r1"]
+        loser_rate = theta["r1"] if response > 0 else theta["r2"]
+        log_pdf = np.log(winner_rate) - winner_rate * rt
+        log_survival = -loser_rate * rt
+        expected.append(log_pdf + log_survival)
```
cpaniaguam: When populating a list, consider extracting the computation to a function and use a comprehension.
```suggestion
    def _compute_exponential_logp(rt, response, winner_rate, loser_rate):
        log_pdf = np.log(winner_rate) - winner_rate * rt
        log_survival = -loser_rate * rt
        return log_pdf + log_survival 
    
    expected = [
        _compute_exponential_logp(
            rt,
            response,
            theta["r2"] if response > 0 else theta["r1"],
            theta["r1"] if response > 0 else theta["r2"],
        )
        for rt, response in data
    ]
```

hayden-johnson:
@AlexanderFengler 

Added an implementation of the Poisson race model in `ssm-simulators` (see [PR 246](https://github.com/lnccbrown/ssm-simulators/pull/246)), along with a minimal example notebook in `HSSM/docs/poisson_race.ipynb`.

Lmk what you think! Feel free to suggest additions to the tutorial notebook.

src/hssm/likelihoods/analytical.py
hayden-johnson: This should be resolved now. We now use a PyTensor equivalent rather than scipy.

src/hssm/likelihoods/analytical.py
```diff
@@ -559,3 +559,103 @@ def logp_lba3(
...
+    t: float,
+    epsilon: float = 1e-15,
+) -> np.ndarray:
+    """Compute analytical log-likelihoods for a 2-accumulator Poisson race.
+
+    Each accumulator time follows a Gamma distribution (with continuous
+    shape parameter) with shape parameter k and rate r. The per-trial
+    likelihood decomposes into the density of the winning accumulator
+    evaluated at the observed decision time and the survival function of
+    the losing accumulator at the same time.
```
cpaniaguam: Would it be okay to update the first sentence from `Each accumulator time follows a Gamma distribution (with continuous shape parameter) with shape parameter k and rate r.`  to `Each accumulator time follows a Gamma distribution with continuous shape parameter k and rate r.`?

src/hssm/likelihoods/analytical.py
```diff
@@ -559,3 +559,103 @@ def logp_lba3(
...
+        Shape parameters (> 0) for the two accumulators.
+    t
+        Non-decision time [0, inf).
+    epsilon
+        A small positive number to prevent division by zero or taking ``log(0)``.
+
+    Returns
+    -------
+    pytensor.tensor.TensorVariable
+        Per-trial log-likelihoods compatible with PyTensor graphs.
```
cpaniaguam: The return type in the function's signature says it's `np.ndarray`. Consider revising this.

src/hssm/likelihoods/analytical.py
```diff
@@ -559,3 +559,103 @@ def logp_lba3(
...
+    rt_safe = pt.switch(negative_rt, epsilon, rt)
+    rt_safe = pt.maximum(rt_safe, epsilon)
+
+    r_c = pt.switch(flip, r2, r1)
+    r_l = pt.switch(flip, r1, r2)
+    k_c = pt.switch(flip, k2, k1)
+    k_l = pt.switch(flip, k1, k2)
+
+    log_pdf = (
+        k_c * pt.log(pt.maximum(r_c, epsilon))
```
cpaniaguam: Why not define `r_c_safe` as well?

src/hssm/likelihoods/analytical.py
```diff
@@ -559,3 +559,103 @@ def logp_lba3(
...
+        + (k_c - 1.0) * pt.log(rt_safe)
+        - r_c * rt_safe
```
cpaniaguam, line 634: Just curious: why use safe versions for a standard product for one variable but not the other? It seems like it should be fine to use the _untreated_ versions.

src/hssm/likelihoods/analytical.py
```diff
@@ -559,3 +559,103 @@ def logp_lba3(
...
+
+# set bounds
+poisson_race_bounds = {
+    "r1": (0.0, np.inf),
+    "r2": (0.0, np.inf),
+    "k1": (0.0, np.inf),
+    "k2": (0.0, np.inf),
+    "t": (0.0, np.inf),
+}
+poisson_race_params = ["r1", "r2", "k1", "k2", "t"]
```
cpaniaguam: Because all the bounds are the same consider building the bounds dict with a comprehension from the params name list.

```suggestion
poisson_race_params = ["r1", "r2", "k1", "k2", "t"]
poisson_race_bounds = {param: (0.0, np.inf) for param in poisson_race_params}
```

docs/tutorials/poisson_race.ipynb
cpaniaguam, line 1: Notebooks looks good to me!

src/hssm/likelihoods/analytical.py
hayden-johnson: Resolved, and added some more detail in the docstring.

src/hssm/likelihoods/analytical.py
hayden-johnson: I was originally only using the safe option for problematic computations (i.e., logarithms). For other computations, I was using unsafe variables; in either case, the evaluation is invalid and should be caught by the parameter checks. 

I was inconsistent with this, however, in that I used rt_safe outside of the log function later in the pdf - I changed this now.

src/hssm/likelihoods/analytical.py
hayden-johnson: True. I meant to only use safe values inside the log function. I changed this for consistency.

AlexanderFengler (dismissed):
This looks good now, thanks @hayden-johnson

src/hssm/modelconfig/poisson_race_config.py
```diff
@@ -0,0 +1,53 @@
...
+    Returns
+    -------
+    DefaultConfig
+        A dict containing the default configuration settings for the Poisson Race Model
+    """
+    return {
+        "response": ["rt", "response"],
+        "list_params": poisson_race_params,
+        "choices": [-1, 1],
+        "description": "The Poisson Race Model",
```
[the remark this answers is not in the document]
hayden-johnson: I will make one last change in this regard, then all set from my perspective ;)

src/hssm/likelihoods/analytical.py
```diff
@@ -559,3 +559,111 @@ def logp_lba3(
...
+
+# set bounds
+poisson_race_params = ["r1", "r2", "k1", "k2", "t"]
+poisson_race_bounds = {
+    "r1": (np.finfo(float).eps, np.inf),
+    "r2": (np.finfo(float).eps, np.inf),
+    "k1": (np.finfo(float).eps, np.inf),
+    "k2": (np.finfo(float).eps, np.inf),
+    "t": (0.0, np.inf),
+}
```
cpaniaguam, line 660: One can always retrieve the `poisson_race_params` object from `poisson_race_bounds` with `list(mydict.keys())`

cpaniaguam (commented):
lgtm

PR #858, opened 2025-12-10
`make_distribution_for_supported_model` for convenient creation of pm.Distribution

digicosmos86:
What's added:
- a `make_distribution_for_supported_model` that creates `pm.Distribution` classes for supported models.
- tests for this function

What's updated:
- `Type` is deprecated in Python 3.9. Updated this in favor of `type`
- Added a section for `make_distribution_for_supported_model` in documentation

### DIFF
--- a/src/hssm/hssm.py
+++ b/src/hssm/hssm.py
@@ -845,7 +845,11 @@ class HSSM(DataValidator):
                 if key_ in [var_.name for var_ in self.pymc_model.deterministics]
             )
         )
-        vars_to_keep_clean = [var_ for var_ in vars_to_keep if "_mean" not in var_]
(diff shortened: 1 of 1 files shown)


tests/distribution_utils/test_distribution_utils.py
```diff
@@ -171,6 +172,22 @@ def fake_logp_function(data, param1, param2):
     )
 
 
+@pytest.mark.slow
```
AlexanderFengler, line 175: worth it to add a basic constructor test for more supported models?

digicosmos86: Not very important to test this one because the it's simply a wrapper of other functions that are tested more comprehensively

Issue #805, opened 2025-09-21
RLSSM model config


krishnbera:
Introduce model configs for RLSSM models in HSSM to store information about the model parameters as well as other model/paradigm info such as choices, extra_fields, etc. In addition, other meta-data fields will need to be introduced to flag certain models as type RLSSM or getting the likelihood function name which would be implemented in some HSSM file such as `rldm.py`. 
 
**Current implementation**
- See example model config below. We can reuse many fields from current SSM model configs but will have to add a few new fields. 
- We can have a separate file that holds all the RLSSM model configs. 
- We can potentially also add RL_params and SSM_params attributes so that we have some metadata to construct likelihoods in a modular fashion. This would also help with downloading the LAN onnx for the corresponding SSM process.

**Example model config**
```
"model_name": "custom_model_1"
"description": "<some text description>",
"type_rlssm": True, 

"list_params": [param1, param2, ... paramN],
"data": ["rt", "response"],
"choices": [0, 1, 2],

"likelihood_function": "<function name from rldm.py>",
"extra_fields": ["ef_1", "ef_2", ... "ef_M"],
```

PR #861, opened 2025-12-12
Add RLSSMConfig class for reinforcement learning models

cpaniaguam:
This PR introduces the `RLSSMConfig` class, a specialized configuration class for Reinforcement Learning Sequential Sampling Models (RLSSM) that inherits from the newly created `BaseModelConfig` abstract base class (#860).

### RLSSMConfig Class
- **Inherits from `BaseModelConfig`**: Leverages the common configuration interface while adding RLSSM-specific functionality
- **RLSSM-specific fields**:
  - `params_default`: List of default parameter values (matches `list_params` order)
  - `decision_process`: Reference to the decision model (e.g., "LAN")
  - `learning_process`: Dictionary of learning functions for computed parameters
  - `n_params`, `n_extra_fields`: Validation counts for parameter consistency
  - `decision_model`, `lan_model`: Additional metadata for RLSSM models

### Methods
- `from_rlssm_dict()`: Factory method to create configs from existing RLSSM dictionary structures
- `validate()`: Comprehensive validation ensuring parameter counts, bounds, and required fields are consistent
- `get_defaults()`: Returns default values and bounds for parameters (returns `float` instead of `ParamSpec`)
- `to_config()`: Converts `RLSSMConfig` to standard `Config` for backward compatibility
- `data` property: Alias for `response` field to support both naming conventions

### Key Differences from Standard Config
- Uses `params_default` (list) instead of `default_priors` (dict) to match RLSSM parameter ordering
- Supports hierarchical structure with separate decision and learning processes
- Returns numeric defaults rather than prior distributions
- Validates consistency between parameter lists and counts

## Testing
- Added 25 comprehensive tests in test_rlssm_config.py
- 100% coverage of `RLSSMConfig` class
- Tests cover creation, validation, defaults, conversion, and data alias functionality

## Usage Example
```python
from hssm.config import RLSSMConfig

# Create from dictionary (matching rlssm_model_config_list structure)
config = RLSSMConfig.from_rlssm_dict("rlwm", {
    "name": "rlwm",
    "description": "RLWM model with collapsing bound decision process",
    "n_params": 4,
    "n_extra_fields": 3,
    "list_params": ["alpha", "beta", "v", "a"],
    "extra_fields": ["participant_id", "trial_id", "feedback"],
    "params_default": [0.5, 0.3, 1.0, 1.5],
    "bounds": {
        "alpha": (0.0, 1.0),
        "beta": (0.0, 1.0),
        "v": (-3.0, 3.0),
        "a": (0.3, 2.5),
    },
    "decision_model": "LAN",
    "LAN": "angle",
    "data": ["rt", "response"],
    "choices": [0, 1],
})

# Validate configuration
config.validate()

# Convert to standard Config for HSSM compatibility
standard_config = config.to_config()
```

## Related
- Depends on PR #860 (BaseModelConfig implementation)
- Part of issue #805: RLSSM model configuration refactoring

### DIFF
--- a/src/hssm/config.py
+++ b/src/hssm/config.py
@@ -303,9 +303,9 @@ class RLSSMConfig(BaseModelConfig):
             Dictionary containing model configuration. Expected keys:
                 - description: Model description (optional)
                 - list_params: List of parameter names (required)
-                - extra_fields: List of extra field names from data (optional)
+                - extra_fields: List of extra field names from data (required)
                 - params_default: Default parameter values (required)
-                - bounds: Parameter bounds (optional)
+                - bounds: Parameter bounds (required)
                 - response: Response column names (required)
                 - choices: Valid choice values (required)
                 - decision_process: Decision process specification (required)


src/hssm/config.py
```diff
@@ -241,6 +244,220 @@ def get_defaults(
...
+
+    # RLSSM-specific: parameter defaults as list (matches list_params order)
+    params_default: list[float] = field(default_factory=list)
+
+    # RLSSM-specific: hierarchical structure
+    decision_process: str | ModelConfig | None = None
+    learning_process: dict[str, Any] = field(default_factory=dict)
+
+    # Number of parameters (for validation)
+    n_params: int | None = None
```
AlexanderFengler: slightly confused about the `n_params` argument. 
we should be able to deduce number of parameters from the `decision_process` + `learning_process` arguments right?
I think this shouldn't be a "free" (independently configurable) attribute at this level.

src/hssm/config.py
```diff
@@ -241,6 +244,220 @@ def get_defaults(
...
+    n_extra_fields: int | None = None
```
AlexanderFengler: same here

src/hssm/config.py
```diff
@@ -241,6 +244,220 @@ def get_defaults(
...
+
+    # Additional metadata for RLSSM models
+    decision_model: str | None = None  # e.g., "LAN"
```
AlexanderFengler: `decision_model` and `lan_model` speaks to a confusion in terminology here. 

There is a distinction between the `decision_model` (ddm, angle etc. etc.) and the `likelihood_kind` (blackbox, approx_differentiable, analytical).

src/hssm/config.py
```diff
@@ -241,6 +244,220 @@ def get_defaults(
...
+    This configuration class is designed for models that combine reinforcement
+    learning processes with sequential sampling decision models (RLSSM).
+    """
+
+    # RLSSM-specific: parameter defaults as list (matches list_params order)
+    params_default: list[float] = field(default_factory=list)
+
+    # RLSSM-specific: hierarchical structure
+    decision_process: str | ModelConfig | None = None
+    learning_process: dict[str, Any] = field(default_factory=dict)
```
AlexanderFengler: Idea is that the `learning_process` here is defined as a dictionary with with keys being the target parameters?

In might not yet understand the logic fully, but this should then deduce what free parameters remain?

src/hssm/config.py
krishnbera: `extra_fields` are tied to the learning rule used so `n_extra_fields` can be derived from there as Alex mentions.

src/hssm/config.py
```diff
@@ -241,6 +244,220 @@ def get_defaults(
...
+        default_priors = {}
+        if self.list_params and self.params_default:
+            default_priors = {
+                param: default
+                for param, default in zip(self.list_params, self.params_default)
+            }
+
+        return Config(
+            model_name=self.model_name,
+            loglik_kind="approx_differentiable",  # RLSSM models use approx_diff
```
krishnbera: While it is true that the current RLSSM models use approx_diff, we would want to relax this assumption. 
As Alex mentioned in one of the comments above, `decision_process` and `loglik_kind` are different aspects where `loglik_kind` can be LANs or analytical.

src/hssm/config.py
cpaniaguam: We'll do `n_params = len(list_params)` for now and see about a more sophisticated way of doing this count using target/computed params later.

src/hssm/config.py
cpaniaguam: > Idea is that the learning_process here is defined as a dictionary with with keys being the target parameters?

Yes, indeed. This is what we are annotating functions with in #824. For example see [this](https://github.com/lnccbrown/HSSM/blob/198c8b042f0318edcac639c3f2b5770eda944337/tests/rl/test_rl_likelihood_builder.py#L365).

src/hssm/config.py
cpaniaguam: I'm moving the `loglik_kind` field to the core config abstract class and making "approx_differentialble" as the default for RL configs.

src/hssm/config.py
```diff
@@ -241,6 +244,217 @@ def get_defaults(
...
+
+        Returns
+        -------
+        RLSSMConfig
+            Configured RLSSM model configuration object.
+        """
+        # Support both 'response' and 'data' keys, with 'data' taking precedence
+        response_value = config_dict.get(
+            "data", config_dict.get("response", ["rt", "response"])
+        )
```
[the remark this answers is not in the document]
cpaniaguam: Sticking to `response` only in alignment with the rest of HSSM.

src/hssm/config.py
```diff
@@ -241,6 +244,217 @@ def get_defaults(
...
+        Parameters
+        ----------
+        model_name
+            The name of the RLSSM model.
+        config_dict
+            Dictionary containing model configuration. Expected keys:
+            - name: Model name
+            - description: Model description
+            - n_params: Number of parameters
+            - n_extra_fields: Number of extra data fields
```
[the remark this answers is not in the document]
cpaniaguam: Removed from the docstring.

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,419 @@
...
+
+
+def test_rlssm_config_basic_creation():
+    """Test basic RLSSMConfig creation."""
+    config = RLSSMConfig(
+        model_name="test_rlssm",
+        description="Test RLSSM model",
+        list_params=["alpha", "beta", "gamma"],
+        params_default=[0.5, 0.3, 0.2],
+        decision_process="LAN",
```
AlexanderFengler: Noticing this here, but this points towards a rough edge in the design. 

The `decision_process` is e.g. `angle` / `ddm` etc..

The `likelihood_kind` can be `approx_differentiable` (which for us atm means LAN) or `analytical` or `blackbox`.

Currently the terminology confuses the idea of a "decision process" (the underlying forward model like the "ddm" etc.) and the type of `likelihood` we use to calculate event probabilities (`approx_differentiable` / `analytical` / `blackbox`)

src/hssm/config.py
```diff
@@ -241,6 +241,194 @@ def get_defaults(
...
+        ----------
+        model_name
+            The name of the RLSSM model.
+        config_dict
+            Dictionary containing model configuration. Expected keys:
+            - name: Model name
+            - description: Model description
+            - list_params: List of parameter names
+            - extra_fields: List of extra field names from data
+            - decision_model: Name of decision model (e.g., "LAN")
```
AlexanderFengler: See longer comment below concerning terminology.
The individual "LAN" attribute should also go here afaic.

tests/test_rlssm_config.py
cpaniaguam: Thanks for pointing this out @AlexanderFengler. There may be fields that likely we won't need until after a first implementation of the RLSSM is built. I will remove the `decision_model` field for the moment and revisit this later.

src/hssm/config.py
```diff
@@ -241,6 +241,191 @@ def get_defaults(
...
+        RLSSMConfig
+            Configured RLSSM model configuration object.
+        """
+        return cls(
+            model_name=model_name,
+            description=config_dict.get("description"),
+            list_params=config_dict.get("list_params"),
+            extra_fields=config_dict.get("extra_fields"),
+            params_default=config_dict.get("params_default", []),
+            decision_process=config_dict.get("decision_model"),
```
AlexanderFengler: Let me try to express the logic here one more time in detail: 

The components we have are:

1. `learning_process`
2. `decision_process`
3. `decision_process_loglik_kind`
4. `learning_process_kind` (we can leave this for future work in principle, since for now we are only considering differentiable learning processes so jax/pytensor)

Downstream we then stitch together the pieces. 

`lan_model` should be dropped as separate entry to the config. Which network to load (and whether to load a network at all) should be discoverable from the information above.

AlexanderFengler: If you can just apply this throughout ( @digicosmos86 in case you have additional thoughts on this ) here, then the PR is ready for me.

AlexanderFengler (changes requested):
Sorry to double down on annoyance but it's what I did :).
Once we agree on the path it should be 10-20min to fix and we can merge. Thanks @cpaniaguam

src/hssm/config.py
cpaniaguam: Thanks for the insight, @AlexanderFengler. It helps me further understand the inner workings of these models :).

We can drop the `lan_model` field. And we can always add/drop others later. We're just getting this off the ground! Thanks again!

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
```
AlexanderFengler: `decision_process="LAN"` should go.
It should be `decision_process="ddm"` e.g.

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+        response=["rt", "response"],
+        choices=[0, 1],
+        extra_fields=["feedback", "trial_id"],
+    )
+
+    assert config.model_name == "test_rlssm"
+    assert config.description == "Test RLSSM model"
+    assert config.list_params == ["alpha", "beta", "gamma"]
+    assert config.params_default == [0.5, 0.3, 0.2]
+    assert config.decision_process == "LAN"
```
AlexanderFengler: same here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+
+
+def test_rlssm_config_from_rlssm_dict():
+    """Test creating RLSSMConfig from dictionary."""
+    config_dict = {
+        "name": "rlwm",
+        "description": "Reinforcement Learning Working Memory model",
+        "list_params": ["alpha", "beta", "gamma", "v", "a"],
+        "extra_fields": ["feedback", "trial_id", "block"],
+        "decision_model": "LAN",
```
AlexanderFengler: same here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+
+    config = RLSSMConfig.from_rlssm_dict("rlwm", config_dict)
+
+    assert config.model_name == "rlwm"
+    assert config.description == "Reinforcement Learning Working Memory model"
+    assert config.n_params == 5
+    assert config.n_extra_fields == 3
+    assert config.list_params == ["alpha", "beta", "gamma", "v", "a"]
+    assert config.extra_fields == ["feedback", "trial_id", "block"]
+    assert config.decision_process == "LAN"
```
AlexanderFengler: and here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+    assert config.learning_process == {"v": "subject_wise_function"}
+
+
+def test_rlssm_config_from_rlssm_dict_with_defaults():
+    """Test creating RLSSMConfig with default values."""
+    config_dict = {
+        "name": "minimal_rlssm",
+        "description": "Minimal RLSSM model",
+        "list_params": ["alpha", "beta"],
+        "decision_model": "LAN",
```
AlexanderFengler: here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+    assert config.learning_process == {}  # Default value
+
+
+def test_rlssm_config_validate_success():
+    """Test successful validation."""
+    config = RLSSMConfig(
+        model_name="test_model",
+        list_params=["alpha", "beta"],
+        params_default=[0.5, 0.3],
+        decision_process="LAN",
```
AlexanderFengler: here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+    # Should not raise
+    config.validate()
+
+
+def test_rlssm_config_validate_missing_response():
+    """Test validation fails when response is missing."""
+    config = RLSSMConfig(
+        model_name="test_model",
+        list_params=["alpha"],
+        decision_process="LAN",
```
AlexanderFengler: here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+
+    with pytest.raises(ValueError, match="Please provide `response` columns"):
+        config.validate()
+
+
+def test_rlssm_config_validate_missing_list_params():
+    """Test validation fails when list_params is missing."""
+    config = RLSSMConfig(
+        model_name="test_model",
+        decision_process="LAN",
```
AlexanderFengler: here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+    with pytest.raises(ValueError, match="Please provide `list_params`"):
+        config.validate()
+
+
+def test_rlssm_config_validate_missing_choices():
+    """Test validation fails when choices is missing."""
+    config = RLSSMConfig(
+        model_name="test_model",
+        list_params=["alpha"],
+        decision_process="LAN",
```
AlexanderFengler: here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+    config = RLSSMConfig(
+        model_name="test_model",
+        list_params=["alpha", "beta", "gamma"],
+        params_default=[0.5, 0.3, 0.2],
+        bounds={
+            "alpha": (0.0, 1.0),
+            "beta": (0.0, 1.0),
+            "gamma": (0.0, 1.0),
+        },
+        decision_process="LAN",
```
AlexanderFengler: we get the idea :D

AlexanderFengler (changes requested):
small naming corrections to tests remain otherwise good afaic.
Thanks @cpaniaguam

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+
+def test_rlssm_config_from_rlssm_dict():
+    """Test creating RLSSMConfig from dictionary."""
+    config_dict = {
+        "name": "rlwm",
+        "description": "Reinforcement Learning Working Memory model",
+        "list_params": ["alpha", "beta", "gamma", "v", "a"],
+        "extra_fields": ["feedback", "trial_id", "block"],
+        "decision_model": "ddm",
+        "LAN": "angle",
```
krishnbera: I'm assuming this needs to be updated based on Alex's comment? 
The config will have `learning_process`, `decision_process`, `decision_process_loglik_kind` and `learning_process_loglik_kind` (optional for now). And we will be dropping `LAN` attribute here.

src/hssm/config.py
```diff
@@ -241,6 +241,186 @@ def get_defaults(
...
+            - description: Model description
+            - list_params: List of parameter names
+            - extra_fields: List of extra field names from data
+            - LAN: Specific LAN model variant
+            - params_default (optional): Default parameter values
+            - bounds (optional): Parameter bounds
+            - response (optional): Response column names
+            - choices (optional): Valid choice values
+            - learning_process (optional): Learning process functions
+            - loglik_kind (optional): Type of likelihood computation
```
krishnbera: A lot of attributes here are marked optional, is this desirable? Or am I missing something?

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,417 @@
...
+
+import hssm
+from hssm.config import Config, RLSSMConfig
+
+hssm.set_floatX("float32")
+
+
+def test_rlssm_config_basic_creation():
+    """Test basic RLSSMConfig creation."""
+    config = RLSSMConfig(
```
krishnbera: i'm assuming this will be updated to contain `learning_process` and other attributes?

src/hssm/config.py
cpaniaguam: They are marked as optional because default values are provided if the user doesn't specify them. You can see how this is handled in lines 308-314:
```python
            params_default=config_dict.get("params_default", []),
            decision_process=config_dict.get("decision_model"),
            learning_process=config_dict.get("learning_process", {}),
            bounds=config_dict.get("bounds", {}),
            response=config_dict.get("response", ["rt", "response"]),
            choices=config_dict.get("choices", [0, 1]),
            loglik_kind=config_dict.get("loglik_kind", "approx_differentiable"),
```
Should they be required?

tests/test_rlssm_config.py
cpaniaguam: I'll drop it, thanks!

tests/test_rlssm_config.py
cpaniaguam: I added `learning_process` to the tests. The fields `decision_process_loglik_kind` and `learning_process_loglik_kind` were missing in the config and are now added. I also updated the tests accordingly. Thanks!

src/hssm/config.py
AlexanderFengler: I would also say as a starting point here we should default to `required` for all these. 
Later we can try to be smart about it. ( @cpaniaguam @krishnbera )

krishnbera: ok, let's just keep everything as required. that should also help us make minimal assumptions. 
also, we will have to flush out the logic of the `LAN` attribute better here since we already have `decision_process_loglik_kind` and `decision_process` attributes. 

if the user is relying on one of the existing models in HSSM, then we can directly get the corresponding LAN from the two attributes. However, we also want to handle the case where user brings their own `decision_process` which is not currently residing in HSSM. there should be a way of just directly specifying that here (via path of LAN onnx file or something).

src/hssm/config.py
```diff
@@ -241,6 +244,196 @@ def get_defaults(
...
+            description=config_dict.get("description"),
+            list_params=config_dict.get("list_params"),
+            extra_fields=config_dict.get("extra_fields"),
+            params_default=config_dict.get("params_default", []),
+            decision_process=config_dict.get("decision_model"),
+            learning_process=config_dict.get("learning_process", {}),
+            bounds=config_dict.get("bounds", {}),
+            response=config_dict.get("response", DEFAULT_RLSSM_RESPONSE),
+            choices=config_dict.get("choices", DEFAULT_RLSSM_CHOICES),
+            loglik_kind=config_dict.get("loglik_kind", "approx_differentiable"),
```
krishnbera: `loglik_kind` is now redundant since we have `decision_process_loglike_kind` and `learning_process_loglik_kind`?

src/hssm/config.py
```diff
@@ -241,6 +244,196 @@ def get_defaults(
...
+        """
+        if "decision_process_loglik_kind" not in config_dict:
+            raise ValueError("decision_process_loglik_kind must be provided")
+        return cls(
+            model_name=model_name,
+            description=config_dict.get("description"),
+            list_params=config_dict.get("list_params"),
+            extra_fields=config_dict.get("extra_fields"),
+            params_default=config_dict.get("params_default", []),
+            decision_process=config_dict.get("decision_model"),
```
krishnbera: We can use `decision_process` as dict key for uniformity.

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,392 @@
...
+        "model_name, config_dict, expected_model_name, expected_params_default, expected_bounds, expected_response, expected_choices, expected_learning_process",
+        [
+            (
+                "rlwm",
+                {
+                    "name": "rlwm",
+                    "description": "Reinforcement Learning Working Memory model",
+                    "list_params": ["alpha", "beta", "gamma", "v", "a"],
+                    "extra_fields": ["feedback", "trial_id", "block"],
+                    "decision_model": "ddm",
```
krishnbera: same here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,392 @@
...
+                [0, 1],
+                {"v": "subject_wise_function"},
+            ),
+            (
+                "minimal_rlssm",
+                {
+                    "name": "minimal_rlssm",
+                    "description": "Minimal RLSSM model",
+                    "list_params": ["alpha", "beta"],
+                    "decision_model": "ddm",
```
krishnbera: same here

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,392 @@
...
+                model_name="test_model",
+                list_params=["alpha"],
+                decision_process="ddm",
+                response=["rt", "response"],
+                choices=[0, 1],
+            )
+        config_dict = {
+            "name": "test_model",
+            "list_params": ["alpha"],
+            "decision_model": "ddm",
```
krishnbera: same here

src/hssm/config.py
krishnbera: one option could be to just drop the `LAN` attribute here for the first pass. We assume that `decision_process` is already registered in HSSM. We can create an issue to address this later after the first pass.

src/hssm/config.py
cpaniaguam: Done.

src/hssm/config.py
cpaniaguam: Dropped `loglik_kind` in favor of `decision_process_loglike_kind`.

tests/test_rlssm_config.py
```diff
@@ -0,0 +1,485 @@
...
+def valid_rlssmconfig_kwargs():
+    return dict(
+        model_name="test_model",
+        list_params=["alpha", "beta"],
+        params_default=[0.5, 0.3],
+        decision_process="ddm",
+        response=["rt", "response"],
+        choices=[0, 1],
+        extra_fields=["feedback"],
+        decision_process_loglik_kind="analytical",
```
AlexanderFengler, line 62: Question on this one: 

Do we actually allow mixing `analytical` with `blackbox` atm?

AlexanderFengler: This combination is used pre-dominantly below. 

Point being, if you have `analytical` + `blackbox` the overall likelihood will become `blackbox`. 

Do we specify this kind of logic anywhere?

AlexanderFengler: Reflecting on it, might depend on where the corresponding logic should go. Might be ok to ignore the current downstream incompatibility at this level for now.

src/hssm/config.py
```diff
@@ -20,8 +20,27 @@
 if TYPE_CHECKING:
     from pytensor.tensor.random.op import RandomVariable
 
+# ====== Centralized RLSSM defaults =====
+DEFAULT_RLSSM_RESPONSE = ["rt", "response"]
```
AlexanderFengler: Rename to `DEFAULT_RLSSM_OBSERVED_DATA`.
add `DEFAULT_SSM_OBSERVED_DATA` and use that one for `Config(BasicConfig)` definition.

src/hssm/config.py
```diff
@@ -20,8 +20,27 @@
...
+DEFAULT_RLSSM_CHOICES = [0, 1]
```
AlexanderFengler: same idea here

src/hssm/config.py
```diff
@@ -159,7 +178,7 @@ def from_defaults(
             return Config(
                 model_name=model_name,
                 loglik_kind=loglik_kind,
-                response=["rt", "response"],
+                response=DEFAULT_RLSSM_RESPONSE,
```
AlexanderFengler: as per comment above, I think it's semantically a bit confusing to have the RLSSM default show up here.

Also sidenote that I just notice now. 

We are doing a redefinition of `reponse` --> `['rt', 'response']` here, which is a bit confusing. 

Maybe renaming this argument to `observed_data_vars` makes more sense globally. (This can be either `["response"]` or `["rts", "response"]` or even something entirely different later).

cpaniaguam: Thanks for this @AlexanderFengler. I think @krishnbera was using the name `data` for this. However, in HSSM model configs `response` was being used already. In the interest of being more consistent I sort of defaulted to

```python
response = ["rt", "response"]
```
for this config class. We can use `observed_data_vars` as you suggest. Again, we can make changes to this later as we actually start to use it to build models. Thoughts?

src/hssm/config.py
```diff
@@ -20,8 +20,28 @@
 if TYPE_CHECKING:
     from pytensor.tensor.random.op import RandomVariable
 
+# ====== Centralized RLSSM defaults =====
+DEFAULT_SSM_OBSERVED_DATA = ("rt", "response")
+DEFAULT_RLSSM_OBSERVED_DATA = ("rt", "response")
```
krishnbera, line 25: We deferred the discussion around what constitutes as "data" in RLSSM or regression case, but I think this assumption can hold for now given that `extra_fields` is a required field.

src/hssm/config.py
```diff
@@ -241,13 +261,199 @@ def get_defaults(
...
+
+        Parameters
+        ----------
+        model_name : str
+            The name of the RLSSM model.
+        config_dict : dict[str, Any]
+            Dictionary containing model configuration. Expected keys:
+                - description: Model description (optional)
+                - list_params: List of parameter names (required)
+                - extra_fields: List of extra field names from data (optional)
```
krishnbera: minor: typo in the docstring? `extra_fields` is required as per `RLSSM_REQUIRED_FIELDS`?

src/hssm/config.py
```diff
@@ -241,13 +261,199 @@ def get_defaults(
...
+                - params_default: Default parameter values (required)
+                - bounds: Parameter bounds (optional)
```
krishnbera: minor: typo again re- optional?

krishnbera (approved):
Looks good to me!

PR #864, opened 2025-12-15
Add-validation-rl-builder

cpaniaguam:
Adds comprehensive input validation to the RL likelihood builder and documents the column ordering contract between data_cols, list_params, and extra_fields. This prevents silent errors from malformed inputs and makes the data/parameter structure explicit.

What Changed
Validation System (builder.py)

Added 5 validation functions checking data shape, args length, trial uniformity, and array shapes
Called automatically in make_rl_logp_func to catch errors early with clear messages
Column Ordering Contract

Module-level docstring documenting the strict ordering requirements:
data[:, i] → data_cols[i]
args[i] → list_params[i]
args[len(list_params) + i] → extra_fields[i]
Updated function docstrings with concrete examples
Test Suite (test_rl_likelihood_builder.py)

39 tests covering validation, integration, and edge cases
Fixed bug where tests passed full dataframes (6 cols) instead of just data_cols (2 cols)
100% coverage of builder.py

### DIFF
--- a/src/hssm/rl/likelihoods/builder.py
+++ b/src/hssm/rl/likelihoods/builder.py
@@ -414,7 +414,7 @@ def _validate_data_shape(
     if data.ndim != 2:
         raise ValueError(
             f"Data array must be 2D, but got shape {data.shape} with "
-            f"{'dimension' if data.ndim == 1 else 'dimensions'}."
+            f"{data.ndim} {'dimension' if data.ndim == 1 else 'dimensions'}."
         )
 
     if data.shape[1] != len(data_cols):


tests/rl/test_rl_likelihood_builder.py
```diff
@@ -168,7 +173,7 @@ def rldm_setup(rldm_data, model_config, param_arrays, annotated_ssm_logp_func):
 
     return RLDMSetup(
         data=rldm_data.data,
-        values=rldm_data.data.values,
```
AlexanderFengler, line 171: don't fully understand why this adjustment is necessary. 

the `rldm_data.data` attribute should have the right columns?

src/hssm/rl/likelihoods/builder.py
```diff
@@ -342,6 +393,180 @@ def _validate_computed_parameters(
...
+    expected_n_args = len(list_params) + len(extra_fields)
+    if len(args) != expected_n_args:
+        raise ValueError(
+            f"Expected {expected_n_args} argument arrays "
+            f"({len(list_params)} list_params + {len(extra_fields)} extra_fields), "
+            f"but got {len(args)}."
+        )
+
+
+def _validate_uniform_trials(
```
AlexanderFengler, line 457: @krishnbera / @cpaniaguam 

Uniform number of trials I guess needs to happen by the time we reach the actual likelihood ops, but upstream we might allow for a jagged setup and pad. 

I remember @krishnbera had padding logic in his code already, are we inheriting this?

krishnbera: @AlexanderFengler The plan is to expect a uniform length array across participants. The user will already specify a padded dataset along with a `trial_likelihood_mask` extra field.  So also ok if we perform this check here - we can probably update the error message to say that supply `trial_likelihood_mask` extra field if using padded datasets.

src/hssm/rl/likelihoods/builder.py
```diff
@@ -342,6 +393,180 @@ def _validate_computed_parameters(
...
+    """
+    total_trials = data.shape[0]
+    expected_trials = n_participants * n_trials
+
+    if total_trials != expected_trials:
+        raise ValueError(
+            f"Data has {total_trials} total trials, but with {n_participants} "
+            f"participants and {n_trials} trials per participant, expected "
+            f"{expected_trials} trials. All participants must have the same "
+            "number of trials."
```
krishnbera, line 486: Can append the error message with the following - 
"... If padded dataset is supplied to ensure uniform number of trials, specify an additional extra fields column (e.g. `trial_likelihood_mask`) that provides a binary mask for ignoring the trial-level likelihood for the padded trials. Note that the masking operation needs to be specified within the likelihood definition."

krishnbera (approved):

tests/rl/test_rl_likelihood_builder.py
cpaniaguam: This is needed here because the object `rldm_data.data`, which is loaded from a file in line 101, is a pandas dataframe and has additional columns.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Thanks for pointing this out!

I see the padding logic in @krishnbera's original code, but it's specific to a particular model.

If padding support is needed, we could approach it two ways:

Option 1: User-side preprocessing
Provide a separate utility tool for users to pad/mask their data before passing it to the model. This keeps the core codebase simple while giving users control over how padding is handled for their specific research context.

Option 2: Built-in support (separate PR)
Add padding as a core feature, which would require:

- General masking API across RL models (not model-specific)
- Updated validation to conditionally allow non-uniform trials when masks are provided
- Documentation on RL state propagation implications
- Comprehensive tests

Recommendation: Option 1 avoids adding complexity to the core codebase while still supporting use cases that need padding.

src/hssm/rl/likelihoods/builder.py
cpaniaguam: Such a change would need to be added in a separate PR adding support for padding. See https://github.com/lnccbrown/HSSM/pull/864/changes#r2665485469.

PR #866, opened 2025-12-17
Update bambi dependency to 0.16 and added support for Python 3.13

digicosmos86:
The following changes were made in this PR:

1. Updated bambi dependency to 0.16
2. Updated `model.sample()` method to be consistent with the `inference_method` argument of `bmb.Model.fit()` method
3. Updated tests and documentation to reflect this change
4. Now that bambi supports Python 3.13, HSSM now supports Python 3.13
5. Updated ci workflows to support Python 3.13
6. Default workflow Python version is now pinned to Python 3.13 for best performance

### DIFF
--- a/src/hssm/hssm.py
+++ b/src/hssm/hssm.py
@@ -627,15 +627,15 @@ class HSSM(DataValidator):
         if sampler in _new_sampler_mapping:
             _logger.warning(
                 f"Sampler '{sampler}' is deprecated. "
-                + "Please use the new sampler names: "
-                + "'pymc', 'numpyro', 'blackjax', 'nutpie', or 'laplace'."
+                "Please use the new sampler names: "
+                "'pymc', 'numpyro', 'blackjax', 'nutpie', or 'laplace'."
             )
             sampler = _new_sampler_mapping[sampler]  # type: ignore
 
         if sampler == "vi":
             raise ValueError(
                 "VI is not supported via the sample() method. "
-                + "Please use the vi() method instead."
+                "Please use the vi() method instead."
             )
 
         if initvals is not None:


src/hssm/hssm.py
```diff
@@ -609,11 +619,25 @@ def sample(
...
+                f"Sampler '{sampler}' is deprecated. "
+                + "Please use the new sampler names: "
+                + "'pymc', 'numpyro', 'blackjax', 'nutpie', or 'laplace'."
+            )
+            sampler = _new_sampler_mapping[sampler]  # type: ignore
+
+        if sampler == "vi":
+            raise ValueError(
+                "VI is not supported via the sample() method. "
+                + "Please use the vi() method instead."
```
cpaniaguam:
```suggestion
                "VI is not supported via the sample() method. "
                "Please use the vi() method instead."
```

src/hssm/hssm.py
```diff
@@ -609,11 +619,25 @@ def sample(
...
+            instance if `sampler` is `"pymc"` (default), `"numpyro"`,
+            `"blackjax"` or "`laplace".
         """
         # If initvals are None (default)
         # we skip processing initvals here.
+        if sampler in _new_sampler_mapping:
+            _logger.warning(
+                f"Sampler '{sampler}' is deprecated. "
+                + "Please use the new sampler names: "
+                + "'pymc', 'numpyro', 'blackjax', 'nutpie', or 'laplace'."
```
cpaniaguam:
```suggestion
                f"Sampler '{sampler}' is deprecated. "
                "Please use the new sampler names: "
                "'pymc', 'numpyro', 'blackjax', 'nutpie', or 'laplace'."
```

src/hssm/hssm.py
```diff
@@ -61,6 +61,13 @@
 
 _logger = logging.getLogger("hssm")
 
+# NOTE: Temporary mapping from old sampler names to new ones in bambi 0.16.0
+_new_sampler_mapping: dict[str, Literal["pymc", "numpyro", "blackjax"]] = {
+    "mcmc": "pymc",
+    "nuts_numpyro": "numpyro",
+    "nuts_blackjax": "blackjax",
+}
+
```
cpaniaguam, line 70: What's the criteria for removing this in the future? When would it be no longer necessary?

src/hssm/hssm.py
```diff
@@ -568,9 +575,8 @@ def find_MAP(self, **kwargs):
 
     def sample(
         self,
-        sampler: (
-            Literal["mcmc", "nuts_numpyro", "nuts_blackjax", "laplace", "vi"] | None
-        ) = None,
+        sampler: Literal["pymc", "numpyro", "blackjax", "nutpie", "laplace"]
```
cpaniaguam, line 578: Would it be good to define this type in the header?

digicosmos86: Programmatically yes, but then this definition won't show up in the documentation

src/hssm/hssm.py
digicosmos86: I think in general we support an old feature for 6 months to a year and then completely deprecate it

PR #876, opened 2026-01-14
874 hssmbase implement parameter initialization and config creation

cpaniaguam:
Add hssmbase.py module from hssm.py with refactorings and tests from test_hssm.py.

### DIFF
--- a/src/hssm/base.py
+++ b/src/hssm/base.py
@@ -269,6 +269,8 @@ class HSSMBase(DataValidatorMixin):
         The jitter value for the initial values.
     """
 
+    config_class = Config
+
     def __init__(
         self,
         data: pd.DataFrame,
@@ -482,8 +484,9 @@ class HSSMBase(DataValidatorMixin):
         )
         _logger.info("Model initialized successfully.")
 
-    @staticmethod
+    @classmethod
     def _build_model_config(
+        cls,
         model: SupportedModels | str,
         loglik_kind: LoglikKind | None,
         model_config: ModelConfig | dict | None,
@@ -508,7 +511,7 @@ class HSSMBase(DataValidatorMixin):
             A complete ModelConfig object with choices and other settings applied.
         """
         # Start with defaults
-        config = Config.from_defaults(model, loglik_kind)
+        config = cls.config_class.from_defaults(model, loglik_kind)
 
         # Handle user-provided model_config
         if model_config is not None:


src/hssm/hssmbase.py
```diff
@@ -0,0 +1,2357 @@
...
+    """
+
+    def __init__(self, fget):
+        self.fget = fget
+
+    def __get__(self, instance, owner):  # noqa: D105
+        return self.fget(owner)
+
+
+class HSSM(DataValidator):
```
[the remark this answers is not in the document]
cpaniaguam: Renaming to Mixin for clarity.

src/hssm/base.py
digicosmos86, line 1: I'd rather not use`hssmbase.py`. `base.py` is more succinct

src/hssm/hssmbase.py
```diff
@@ -0,0 +1,2374 @@
...
+        -------
+        dict[str, Any]
+            A dictionary containing all initialization arguments, excluding 'self'.
+        """
+        init_args = {
+            k: v for k, v in locals_dict.items() if k not in ["self", "kwargs"]
+        }
+        if kwargs:
+            init_args.update(kwargs)
+        return init_args
```
digicosmos86: We likely won't need this. This basically returns a dictionary that stores the initial values of the arguments for serialization purposes. However, since we are not serialize models, we should not need this in the future

src/hssm/hssmbase.py
```diff
@@ -0,0 +1,2374 @@
...
+        initval_jitter: float = INITVAL_JITTER_SETTINGS["jitter_epsilon"],
+        **kwargs,
+    ):
+        # Attach arguments to the instance
+        # so that we can easily define some
+        # methods that need to access these
+        # arguments (context: pickling / save - load).
+        # assert False, "Temporary prevent instantiation of HSSM class."
+        # Define a dict with all call arguments:
+        self._init_args = self._get_init_args(locals(), kwargs)
```
digicosmos86: No need for this any more

src/hssm/base.py
```diff
@@ -0,0 +1,2374 @@
...
+        choices : list[int] | None
+            User-provided choices list.
+
+        Returns
+        -------
+        ModelConfig
+            A complete ModelConfig object with choices and other settings applied.
+        """
+        # Start with defaults
+        config = Config.from_defaults(model, loglik_kind)
```
digicosmos86: `Config` here should be a class variable referring to the actual type of the config. That way, we can swap out the class variable in HSSM and RLSSM classes without overriding this method

src/hssm/hssmbase.py
```diff
@@ -0,0 +1,2374 @@
...
+    """
+
+    def __init__(self, fget):
+        self.fget = fget
+
+    def __get__(self, instance, owner):  # noqa: D105
+        return self.fget(owner)
+
+
+class HSSM(DataValidatorMixin):
```
digicosmos86: The name should be `HSSMBase` right?

digicosmos86 (changes requested):
Looks good! This PR actually does two things: 1: update the name of `DataValidator` to `DataValidatorMixin`, which I really like. 2. Beginning of the implementation of `HSSMBase` class. This needs to be made clear in the description of the PR. Or maybe the changes to `DataValidator` should be in its own PR

A few other suggestions:
1. Maybe delete the rest of the `HSSMBase` class so people can focus on what is changed. There is no diff since it's a new file. Maybe that's easier
2. Instead of merging to `main`, create a copy of the `main` branch and merge to that branch. Tests of this class should be added incrementally

src/hssm/hssmbase.py
cpaniaguam: removed

src/hssm/hssmbase.py
cpaniaguam: removed

src/hssm/base.py
```diff
@@ -0,0 +1,2344 @@
+"""HSSM: Hierarchical Sequential Sampling Models.
```
AlexanderFengler, line 1: this string may need some adjustment?

src/hssm/base.py
```diff
@@ -0,0 +1,2344 @@
...
+    """
+
+    def __init__(self, fget):
+        self.fget = fget
+
+    def __get__(self, instance, owner):  # noqa: D105
+        return self.fget(owner)
+
+
+class HSSMBase(DataValidatorMixin):
```
AlexanderFengler, line 99: possible should add here something like: 

"This is not to be used independently [...]"

AlexanderFengler:
> > I am slightly confused about the intent of `HSSMBase` here. Reading through it it seems too much of a direct copy of the `HSSM` class, I was expecting a few more `Not implemented` statements that get specified only at the level of the inheriting class.
> > Could we clarify this a bit? ( @cpaniaguam , @digicosmos86 )
> 
> Thank you for the feedback, @AlexanderFengler . You're right that `HSSMBase` currently closely mirrors `HSSM`. This is intentional at this stage; we're taking an incremental refactoring approach:
> 
> 1. Current PR scope: Extract the base class with concrete implementations to establish the architecture without breaking existing functionality
> 2. Future work: As we develop additional model types (e.g., RLSSM variants), we'll identify which methods truly need to be abstract based on actual diverging requirements, rather than speculating prematurely
> 
> Starting with mostly concrete methods allows us to:
> 
> * Maintain test coverage during refactoring
> * Avoid premature abstraction
> * Make incremental, reviewable changes
> 
> We'll add `@abstractmethod` declarations as subclass-specific needs become clear through implementation. For now, I think only `__init__`, `_make_model_distribution`, and the configs are candidates for abstraction since we know already these will differ across model types.

Ok thanks @cpaniaguam I understand the strategy now. Then I think we are ok here.

AlexanderFengler (approved):

PR #878, opened 2026-01-14
877 update datavalidator to a mixin

cpaniaguam:
This pull request refactors the behavioral data validation logic for HSSM models by replacing the `DataValidator` class with a more flexible `DataValidatorMixin`. This change improves code reusability and clarity by making data validation functionality available as a mixin, and updates all references and tests accordingly. Minor improvements to method checks and property handling are also included.

**Refactoring and code structure:**

* Replaced the `DataValidator` class with `DataValidatorMixin`, updating its documentation to clarify that it is a mixin and expects certain attributes to be defined in subclasses. All references in the codebase and tests have been updated to use `DataValidatorMixin` instead of `DataValidator`. (`src/hssm/data_validator.py`, `src/hssm/hssm.py`, `tests/test_data_validator.py`, `tests/test_data_sanity.py`) [[1]](diffhunk://#diff-a413a2dab5c466fd6fc9d808d809de6876d77d26044d329b525d36d7d228c003L14-R44) [[2]](diffhunk://#diff-886b4f455536b5c662d101c272845eaf0d7ea4acc0475bc11473cbf11147546eL35-R35) [[3]](diffhunk://#diff-886b4f455536b5c662d101c272845eaf0d7ea4acc0475bc11473cbf11147546eL100-R100) [[4]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L7-R7) [[5]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L51-R66) [[6]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L112-R112) [[7]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L136-R136) [[8]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L147-R147) [[9]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L157-R157) [[10]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L167-R167) [[11]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L192-R209) [[12]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L222-R230) [[13]](diffhunk://#diff-9d11fe60d48fad4746362a0ffea6be7e337accafc731e52b31c19b3134c77e85L240-R240) [[14]](diffhunk://#diff-599d988e92fbbe94680aef502dec6c71b29a6fc18d255d23a3d9462e78f8e8b8L16-R16)

**API and logic improvements:**

* Improved handling of optional arguments in `_update_extra_fields` and related methods, including more explicit `None` checks and clarifying comments about expected subclass attributes. (`src/hssm/data_validator.py`)
* Updated static method calls to reference `DataValidatorMixin` instead of `DataValidator` for field checking. (`src/hssm/data_validator.py`)
* Enhanced property methods in `HSSM` to handle cases where `self.response` may be `None`, returning empty or default values as appropriate. (`src/hssm/hssm.py`)

### DIFF
--- a/tests/test_data_validator.py
+++ b/tests/test_data_validator.py
@@ -4,7 +4,7 @@ from typing import Callable
 import pytest
 import pandas as pd
 import numpy as np
-from hssm.data_validator import DataValidator
+from hssm.data_validator import DataValidatorMixin
 from hssm.defaults import MissingDataNetwork
 
 
@@ -48,8 +48,8 @@ def base_data_nan_missing():
 
 def dv_instance(
     data_factory: Callable = _base_data, deadline: bool = True
-) -> DataValidator:
-    return DataValidator(
+) -> DataValidatorMixin:
+    return DataValidatorMixin(
         data=data_factory(),
         extra_fields=["extra"],
         deadline=deadline,
@@ -57,13 +57,13 @@ def dv_instance(
 
 
 def test_constructor(base_data):
-    dv = DataValidator(
+    dv = DataValidatorMixin(
         data=base_data,
         extra_fields=["extra"],
         deadline=True,
     )
 
-    assert isinstance(dv, DataValidator)
+    assert isinstance(dv, DataValidatorMixin)
     assert dv.data.equals(_base_data())
     assert dv.response == ["rt", "response"]
     assert dv.choices == [0, 1]
@@ -109,7 +109,7 @@ def test_post_check_data_sanity_valid(base_data):
     ):
         dv_instance_no_missing._post_check_data_sanity()
 
-    dv_instance_no_missing = DataValidator(
+    dv_instance_no_missing = DataValidatorMixin(
         data=base_data,
         deadline=False,
         missing_data=False,
@@ -133,7 +133,7 @@ def test_post_check_data_sanity_valid(base_data):
 def test_handle_missing_data_and_deadline_deadline_column_missing(base_data):
     # Should raise ValueError if deadline is True but deadline_name column is missing
     data = base_data.drop(columns=["deadline"])
-    dv = DataValidator(
+    dv = DataValidatorMixin(
         data=data,
         deadline=True,
     )
@@ -144,7 +144,7 @@ def test_handle_missing_data_and_deadline_deadline_column_missing(base_data):
 def test_handle_missing_data_and_deadline_deadline_applied(base_data):
     # Should set rt to -999.0 where rt >= deadline
     base_data.loc[0, "rt"] = 2.0  # Exceeds deadline
-    dv = DataValidator(
+    dv = DataValidatorMixin(
         data=base_data,
         deadline=True,
     )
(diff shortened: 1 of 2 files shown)


tests/test_data_validator.py
```diff
@@ -219,15 +219,15 @@ def test_set_missing_data_and_deadline_all_missing():
         match="`missing_data` is set to True, but you have no valid data in your "
         "dataset.",
     ):
-        DataValidator._set_missing_data_and_deadline(True, False, data)
+        DataValidatorMixin._set_missing_data_and_deadline(True, False, data)
```
AlexanderFengler, line 222: can we name the arguments explicitly here?
(counts for above and below too)

It's easier to read the code that way.

src/hssm/data_validator.py
```diff
@@ -11,21 +11,37 @@
 _logger = logging.getLogger("hssm")
 
 
-class DataValidator:
-    """Class for validating and preprocessing behavioral data for HSSM models."""
+class DataValidatorMixin:
```
AlexanderFengler, line 14: I like the mix-in approach here, however I suspect that we will need different types of Mix-ins for different types of HSSM subclasses, which then may defeat the purpose of the pattern?

Just a thought. Maybe also @digicosmos86 has a gut feeling here.

AlexanderFengler (changes requested):
Small comments. Thanks @cpaniaguam

src/hssm/data_validator.py
cpaniaguam: > however I suspect that we will need different types of Mix-ins for different types of HSSM subclasses

Actually, this pattern provides the flexibility we need for that specific use case you mention.

```python
# Base validation mixin (current)
class DataValidatorMixin:
    def _check_extra_fields(self, data): ...
    def _pre_check_data_sanity(self): ...

# Alternative validation for a different model type
class RLSSMValidatorMixin(DataValidatorMixin):
    def _pre_check_data_sanity(self):
        # Override with RL-specific checks
        super()._pre_check_data_sanity()
        # Add RL-specific validation
        
# Or completely different validation
class CustomValidatorMixin:
    # Different validation workflow entirely
    ...

# Usage
class HSSM(DataValidatorMixin, ...): ...
class RLSSM(RLSSMValidatorMixin, ...): ...
class CustomModel(CustomValidatorMixin, ...): ...
```

So we can use different validation strategies while maintaining code reuse for common patterns.

PR #891, opened 2026-02-10
fix dimensionality errors which were introduced with Bambi 0.17.0

AlexanderFengler:


src/hssm/data_validator.py
```diff
@@ -29,7 +29,7 @@ class DataValidatorMixin:
     def __init__(
         self,
         data: pd.DataFrame,
-        response: list[str] | None = ["rt", "response"],
+        response: tuple[str, ...] | None = ("rt", "response"),
```
digicosmos86: Why do we want a `tuple` here? Since we are later extending this list, that means `response` is modified, and therefore should be a `list`?

pyproject.toml
```diff
@@ -26,9 +26,9 @@ classifiers = [
 dependencies = [
     "absl-py>=2.3.1",
     "arviz>=0.22.0",
-    "bambi>=0.16.0,<0.17.0",
+    "bambi>=0.17.2", #>=0.16.0,<0.17.0",
     "cloudpickle>=3.0.0",
-    "formulae<0.6.0",
+    "formulae>=0.6.0", #<0.6.0",
```
digicosmos86: This is a temporary fix and should be removed since the bambi version is fixed. We might run into other version shenanigans by leaving it here

digicosmos86 (approved):
Looks good to me! Minor comment about the decision to use `tuple` over `list` in one place

src/hssm/data_validator.py
AlexanderFengler: There were some test failures because we had an inconsistency on that front between Configs and some expectations somewhere inside `hssm.py`. 

Aligned it no tuple because the config spit that out, but can try to align both on list and see if tests pass.

pyproject.toml
AlexanderFengler: Sorry can you say what exactly I should remove? Are you simply talking about the comments, or we should actually fix the bambi version?

PR #905, opened 2026-02-25
Removed constructor from DataValidatorMixin

digicosmos86:
Note: I am trying to separate #903 into several smaller PRs so it's easier to review and track the changes

This PR removes the constructor from DataValidatorMixin. As a mixin, its internal states should come from the HSSM class. Currently, its constructor is called from within HSSM class with default values, and then static methods are used, which can lead to inconsistencies.

This PR only fixes this class. Will need to modify HSSM class itself in future PRs

### DIFF
--- a/src/hssm/config.py
+++ b/src/hssm/config.py
@@ -260,6 +260,16 @@ class Config(BaseModelConfig):
         """
         return self.default_priors.get(param), self.bounds.get(param)
 
+    @property
+    def is_choice_only(self) -> bool:
+        """Check if the model is a choice-only model."""
+        # Treat both None and an empty list as invalid configurations.
+        if not self.response:
+            raise ValueError(
+                "Please provide at least one `response` column in the configuration."
+            )
+        return len(self.response) == 1
+
 
 @dataclass
 class RLSSMConfig(BaseModelConfig):
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -18,6 +18,7 @@ def test_from_defaults():
     assert config1.loglik is not None
     assert "t" in config1.default_priors
     assert "v" in config1.bounds
+    assert not config1.is_choice_only
 
     # Case 2: Model supported, but no default prior
     config2 = Config.from_defaults("angle", "analytical")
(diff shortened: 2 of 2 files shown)


src/hssm/data_validator.py
```diff
@@ -131,6 +104,11 @@ def _post_check_data_sanity(self):
     def _handle_missing_data_and_deadline(self):
         """Handle missing data and deadline."""
         if not self.missing_data and not self.deadline:
+            # In the case of choice only model, we don't need to do anything with the
```
AlexanderFengler, line 107: probably can't merge this way? 
What is needed here, slight test rewrite?

tests/test_data_validator.py
```diff
@@ -164,24 +187,24 @@ def test_update_extra_fields(monkeypatch):
...
         extra_fields=["extra", "extra2"],
     )
 
     # Mock the model_distribution attribute
     class DummyModelDist:
-        pass
+        extra_fields: list
 
-    dv.model_distribution = DummyModelDist()
+    dv.model_distribution = DummyModelDist()  # type: ignore[assignment]
```
AlexanderFengler, line 199: Why not something like:

`dv.model_distribution = cast(ExpectedModelDistributionType, DummyModelDist())`

src/hssm/data_validator.py
digicosmos86: There are some new rules about to be introduced to the HSSM class. Basically, since there is no `rt` fields in choice-only models, some of these checks are either no longer needed or can be largely skipped. Another PR will be added to address this. The merge conflicts can get really crazy since we have so many outstanding PRs. I will only introduce the new PR after we can merge the current ones

tests/test_data_validator.py
digicosmos86: `cast` is really weird too and is only done to please the type checker. it's a bit redundant in tests

digicosmos86: After checking, I don't think this is a change worth implementing. You need to explicitly declare a type derived from `pm.Distribution` that has `extra_field` and other attributes specifically for this test. As a unit test this is a bit of an overkill

PR #909, opened 2026-02-26
Added softmax likelihood for choice-only models

digicosmos86:
Note: the test only tests that the output shape is correct (a 1-d array of length n where n is the length of data)

### DIFF
--- a/src/hssm/likelihoods/analytical.py
+++ b/src/hssm/likelihoods/analytical.py
@@ -567,7 +567,7 @@ def softmax_inv_temperature(data: np.ndarray, beta: np.ndarray, *logits):
     data
         1D array of responses (choices).
     beta
-        A scaler for the softmax temperature (0, inf).
+        A scalar for the softmax temperature (0, inf).
     *logits
         Logits for each choice excluding logit0.
 


src/hssm/likelihoods/analytical.py
```diff
@@ -395,14 +393,14 @@ def logp_ddm_sdv(
 ddm_params = ["v", "a", "z", "t"]
 ddm_sdv_params = ddm_params + ["sv"]
 
-DDM: Type[pm.Distribution] = make_distribution(
+DDM: type[pm.Distribution] = make_distribution(
```
AlexanderFengler, line 396: this is interesting.
Should this (is this) be (en) done on the entire repo?

tests/test_likelihoods_choice_only.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+@pytest.mark.parametrize(
+    "beta", [_SCALAR_BETA, _VECTOR_BETA], ids=["scalar_beta", "vector_beta"]
+)
+@pytest.mark.parametrize(
+    "logit", [_SCALAR_LOGIT, _VECTOR_LOGIT], ids=["scalar_logit", "vector_logit"]
+)
+def test_softmax_inv_temperature_shape_2choice(beta, logit):
+    result = softmax_inv_temperature(_DATA_BINARY, beta, logit)
```
AlexanderFengler, line 28: this is ok just a note (may be addressed in the other PR already, I am going to look at that next). What is the consistent approach on the pytensor level? Defining every base likelihood as a "likelihood" or a "log-likelihood"? From memory I though we consistently define log likelihoods.

AlexanderFengler (commented):
Small comments, both concerned with consistency.

src/hssm/likelihoods/analytical.py
digicosmos86: This is a python version issue, as `Type` is deprecated in newer versions of Python. I am only doing this for this file. If we want to do this for the entire repo, we should start a new issue

tests/test_likelihoods_choice_only.py
digicosmos86: My math is really rusty these days. Yes, all likelihoods are log likelihood. I'll update the likelihood function

AlexanderFengler (approved):
I think this one is good to go. Thanks @digicosmos86

PR #911, opened 2026-02-26
Added config files for softmax

### LINKED ISSUE #910: Add config files for `softmax_inv_temperature_2` and `softmax_inv_temperature_3`

digicosmos86:
Closes #910

### DIFF
--- a/src/hssm/modelconfig/softmax_inv_temperature_3_config.py
+++ b/src/hssm/modelconfig/softmax_inv_temperature_3_config.py
@@ -15,4 +15,4 @@ def get_softmax_inv_temperature_3_config() -> DefaultConfig:
         Inverse Softmax Temperature Model with 3 logits, including response variables,
         model parameters, choices, description, and likelihood specifications.
     """
-    return softmax_inv_temperature_config(n_logits=3)
+    return softmax_inv_temperature_config(n_choices=3)
--- a/src/hssm/modelconfig/softmax_inv_temperature_2_config.py
+++ b/src/hssm/modelconfig/softmax_inv_temperature_2_config.py
@@ -15,4 +15,4 @@ def get_softmax_inv_temperature_2_config() -> DefaultConfig:
         Inverse Softmax Temperature Model with 2 logits, including response variables,
         model parameters, choices, description, and likelihood specifications.
     """
-    return softmax_inv_temperature_config(n_logits=2)
+    return softmax_inv_temperature_config(n_choices=2)


src/hssm/modelconfig/softmax_inv_temperature_3_config.py
cpaniaguam, line 1: These derived functions are relatively simple. Why not put them in the same module?

tests/test_modelconfig.py
```diff
@@ -108,3 +112,74 @@ def test_load_all_supported_model_configs(model):
...
+    # Test bounds
+    assert lk_analytical["bounds"]["beta"] == (0.0, np.inf)
+    assert lk_analytical["bounds"]["logit1"] == (-np.inf, np.inf)
+
+    # Test default priors
+    assert lk_analytical["default_priors"]["beta"] == {
+        "name": "HalfNormal",
+        "mu": 0.0,
+        "sigma": 1.0,
+    }
```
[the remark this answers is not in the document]
digicosmos86: Already updated

src/hssm/modelconfig/softmax_inv_temperature_3_config.py
digicosmos86: Looks like this https://github.com/lnccbrown/HSSM/blob/8cc94786ecb903707edf8ee5994e82d1ed6b65cb/src/hssm/modelconfig/__init__.py#L38 determines how default model configs are loaded?

cpaniaguam: I mean the three files `_softmax_inv_temperature_config.py`, `softmax_inv_temperature_2_config.py`, and `softmax_inv_temperature_3_config.py` could go into a single one, `softmax_inv_temperature` say. I am suggesting this because they each have one single function. But again, it's totally optional.

digicosmos86: Totally, but the `get_default_model_config` function dynamically imports `get_{model_name}_config` from `{model_name} config` module, so I was following this setup. Might be clearer this way?

cpaniaguam: Oh, I see. I had forgotten about that :sweat_smile:

src/hssm/modelconfig/_softmax_inv_temperature_config.py
```diff
@@ -0,0 +1,49 @@
+"""The default configuration for the Inverse Softmax Temperature Model with 2 logits."""
```
krishnbera, line 1: This should be "Inverse Temperature Softmax Model" or even "Softmax Inv. Temperature" works. 
Ofc, this comment line is fine.

src/hssm/modelconfig/_softmax_inv_temperature_config.py
```diff
@@ -0,0 +1,49 @@
...
+        Inverse Softmax Temperature Model, including response variables, model
+        parameters, choices, description, and likelihood specifications.
+    """
+    if n_logits < 2:
+        raise ValueError("n_logits must be at least 2.")
+
+    bounds = {"beta": (0.0, np.inf)}
+    bounds.update({f"logit{i}": (-np.inf, np.inf) for i in range(1, n_logits)})
+    default_priors: dict[str, ParamSpec] = {
+        "beta": {"name": "HalfNormal", "mu": 0.0, "sigma": 1.0},
```
krishnbera: We can update the default priors here @AlexanderFengler ? Since close to zero would only make it random. 
"beta": {"name": "HalfNormal", "mu": 1.0, "sigma": 2.0},

src/hssm/modelconfig/_softmax_inv_temperature_config.py
```diff
@@ -0,0 +1,49 @@
...
+    }
+    default_priors.update(
+        {
+            f"logit{i}": {"name": "Normal", "mu": 0.0, "sigma": 1.0}
```
krishnbera, line 40: this might need an update - pending what @AlexanderFengler thinks about the suggested prior above?

src/hssm/modelconfig/_softmax_inv_temperature_config.py
```diff
@@ -0,0 +1,49 @@
+"""The default configuration for the Inverse Softmax Temperature Model with 2 logits."""
+
+import numpy as np
+
+from .._types import DefaultConfig, ParamSpec
+from ..likelihoods.analytical import softmax_inv_temperature
+
+
+def softmax_inv_temperature_config(n_logits: int = 2) -> DefaultConfig:
```
krishnbera: maybe `n_choices` is a better parameter name here rather than `n_logits` because the free params we estimate is the logits and number of logits = `n_choice - 1`.

src/hssm/modelconfig/_softmax_inv_temperature_config.py
```diff
@@ -0,0 +1,49 @@
...
+    Get the default config for the Inverse Softmax Temperature Model.
+
+    Returns
+    -------
+    DefaultConfig
+        A dictionary containing the default configuration settings for the
+        Inverse Softmax Temperature Model, including response variables, model
+        parameters, choices, description, and likelihood specifications.
+    """
+    if n_logits < 2:
```
AlexanderFengler: as per @krishnbera 's comments above, this should be `n_logits == 0`, but this seems to refer to `n_choices`?

src/hssm/modelconfig/_softmax_inv_temperature_config.py
AlexanderFengler: I think this one is ok for now, but we can check what the variance setting de facto implies about the expected choice distributions.

src/hssm/modelconfig/_softmax_inv_temperature_config.py
AlexanderFengler: I think `HalfNormal` doesn't actually have "mu" parameter? (See also a comment in the tests to the same effect).

Maybe something like `Gamma(alpha=2, beta=0.5)` could work here.

@krishnbera https://distribution-explorer.github.io/continuous/gamma.html

could you take a look if you agree that those settings are ok for prior on temperature? Alternatively, can replicate HDDM on this front as well (albeit I don't know if particularly much through was put into on the HDDM side, to justify copying without second thought).

AlexanderFengler (changes requested):
This is 99%. Thanks @digicosmos86

src/hssm/modelconfig/_softmax_inv_temperature_config.py
krishnbera: yeah, right - TruncatedNormal not HalfNormal with mu=1, sigma=2, lower=0, upper=10.
But this gamma parameterization works as well.

PR #920, opened 2026-03-06
Update HSSM class for choice-only models

digicosmos86:
This PR updates the HSSM class to be compatible with choice-only models. Changes are:

1. updated `dist.py` with `is_choice_only` toggle to `make_distribution` and `make_hssm_rv` to account for choice-only models
2. updated `DataValidator` to skip certain checks if the model is choice-only
3. modified one helper function for choice-only regressions in `regression_param.py`
4. updated simple tests to choice-only models

EDIT: I had to update package dependencies to streamline CI workflows based on how `uv` works. `pyproject.toml` was a bit messy:
1. `dev` dependencies (installed with `uv add package-name --dev`) are automatically installed with `uv sync`. These should include all dev dependencies required for tests.
2. Other dependencies in dependency groups (installed with `uv add package-name --group group_name`) are not automatically installed via `uv sync` but can be requested with `--group` flag. These should include specific dependencies required for some workflows. For example, `notebook` dependency group should only be specifically requested when notebooks need to be executed
3. Optional dependencies (installed with `uv add package-name --extra extra_name`) are not installed via `uv sync` either, but can be requested with `--extra` flag. These are extra dependencies that can be requested via pip in square brackets (e.g. `pip install hssm[cuda12]`). These are more user facing and should not be installed on ci unless specific tests require them

### DIFF
--- a/src/hssm/hssm.py
+++ b/src/hssm/hssm.py
@@ -489,8 +489,6 @@ class HSSM(DataValidatorMixin):
         # Process lapse distribution
         self.has_lapse = p_outlier is not None and p_outlier != 0
         self._check_lapse(lapse)
-        if self.has_lapse and self.list_params[-1] != "p_outlier":
-            self.list_params.append("p_outlier")
 
         # Process all parameters
         self.params = Params.from_user_specs(
@@ -517,7 +515,6 @@ class HSSM(DataValidatorMixin):
 
         # Set p_outlier and lapse
         self.p_outlier = self.params.get("p_outlier")
-        self.lapse = lapse if self.has_lapse else None
 
         self._post_check_data_sanity()
 
@@ -1999,12 +1996,13 @@ class HSSM(DataValidatorMixin):
     def _check_lapse(self, lapse):
         """Determine if p_outlier and lapse is specified correctly."""
         # Basically, avoid situations where only one of them is specified.
-        if "p_outlier" in self.list_params and self.list_params[-1] != "p_outlier":
-            raise ValueError(
-                "Please do not include 'p_outlier' in `list_params`. "
-                + "We automatically append it to `list_params` when `p_outlier` "
-                + "parameter is not None"
-            )
+        if self.list_params[-1] != "p_outlier":
+            if "p_outlier" in self.list_params:
+                raise ValueError(
+                    "Please do not include 'p_outlier' in `list_params`. "
+                    + "We automatically append it to `list_params` when `p_outlier` "
+                    + "parameter is not None"
+                )
         if self.has_lapse:
             if lapse is None:
                 if self.is_choice_only:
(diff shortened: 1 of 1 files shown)


src/hssm/distribution_utils/dist.py
```diff
@@ -202,6 +203,8 @@ def make_hssm_rv(
         A list of str of all parameters for this `RandomVariable`.
     lapse : optional
         A bmb.Prior object representing the lapse distribution.
+    is_choice_only : bool
+        Whether the model is a choice-only model. This parameter overrides
```
cpaniaguam: The last sentence here seems incomplete. What does the parameter override?

src/hssm/distribution_utils/dist.py
```diff
@@ -561,12 +572,15 @@ def logp(data, *dist_params):  # pylint: disable=E0213
                         "lapse_func is not defined. "
                         "Make sure lapse is properly initialized."
                     )
-                lapse_logp = lapse_func(data[:, 0].eval())
+                data_for_lapse = data if is_choice_only else data[:, 0]
+                lapse_logp = lapse_func(data_for_lapse.eval())
```
AlexanderFengler, line 579: We should be careful here. 

The lapse function logic we have is based around `rt` lapses, so somethine like `Uniform(0, 20)`, which then evaluates to `1/20` for each rt, makes sense. 

If a model is `choice_only` we want a different default lapse distribution, I would suggest simply `1/n_choices`.

src/hssm/hssm.py
```diff
@@ -396,50 +397,67 @@ def __init__(
         self.model_config.validate()
 
         # Set up shortcuts so old code will work
-        self.response = self.model_config.response
+        # TODO: add to HSSMBase
+        self.response = self.model_config.response[:]
```
AlexanderFengler, line 401: you seem to assign `self.response` twice

digicosmos86: Not sure what you mean?

digicosmos86: Ah I see. The second assignment is simply a cast

src/hssm/distribution_utils/dist.py
digicosmos86: @AlexanderFengler I have made the updates to fix this. Can you take another look?

BTW: this is an important implementation detail. In the future, can this type of details be communicated to us early? It's much easier to plan for these details early

AlexanderFengler: @digicosmos86 the truth is that I didn't think about lapse distributions in this context before that's why I didn't mention it earlier.

When I saw it I realized this needs to be figured out correctly.

AlexanderFengler (approved):
This seems ok now. Eventually we might want to get out of the if-else doom with a refactor that focuses the execution path between `choice_only` and `choice,rt` models, but can do that later. Thanks @digicosmos86

PR #926, opened 2026-03-10
fix: safe negative-RT handling and JAX backend for RDM3

krishnbera:
- Replace element-wise jax_check_parameters for rt > 0 with jnp.where + LOGP_LB floor in logp_rdm3. This clamps negative decision times to a small epsilon before passing to the PDF/CDF, preventing NaN in the likelihood and providing a finite gradient floor per trial.
- Set backend to "jax" and add default HalfNormal priors for A and t in both racing_diffusion_3_config and rdm3_config.

Made-with: Cursor

### DIFF
--- a/src/hssm/likelihoods/analytical.py
+++ b/src/hssm/likelihoods/analytical.py
@@ -464,11 +464,11 @@ def _jax_rdm_tcdf(t, b, v, A, eps_t: float = 1e-8, eps_v: float = 1e-6):
     return F
 
 
-def _jax_rdm3_ll(t, ch, A, b, v0, v1, v2):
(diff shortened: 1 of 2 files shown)


src/hssm/likelihoods/analytical.py
```diff
@@ -497,9 +498,13 @@ def logp_rdm3(
     response = data_reshaped[:, 1]
     response_int = response.astype(jnp.int_)
 
-    logp = _jax_rdm3_ll(rt, response_int, A, b, v0, v1, v2).squeeze()
+    negative_rt = rt <= 0.0
```
cpaniaguam: May I suggest the name  `is_negative_rt`?

cpaniaguam (commented):
lgtm. Just one little suggestion.

PR #893, opened 2026-02-11
HSSM base  + RLSSM classes

cpaniaguam:
- Base model class `HSSMBase`
- Subclass `HSSMBase` for `HSSM`
- Subclass `HSSMBase` for new `RLSSM` class
- For the new `RLSSM` class, introduce new configs and likelihoods
- Associated tests

PR #953, opened 2026-05-04
Plan for aDDM HSSM integration

AndrewZhang599:
Pull request with initial plan (in Markdown) for aDDM integration (Claude-aided). The general idea is to first copy over the JAX-implemented likelihood algorithm and Gaussian-Legendre constants from the following repository: [(https://github.com/RiverFlowsInYou98/efficient-fpt)](url)

Then, mirror the RLSSM config setup, including parameters (reward values, order of drift arrays, saccade array, and dimension of saccade array) as extra_fields. 

One thing I'm not entirely certain how to incorporate: including nondecision time in some form for the parameter list.


AlexanderFengler (changes requested):
One high level comment: 

I don't understand why so many files are touched here, seems kind of by accident. 

Can you take out the plan `.md` file and maybe turn that into a clean PR that doesn't touch anything else? 

I'll still leave my feedback here for now.

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+
+## Context
+
+The attentional drift diffusion model (aDDM; Krajbich et al.) extends the standard DDM by modulating the drift rate based on **which option the subject is currently fixating**. A fast, differentiable JAX likelihood for the aDDM has been prototyped in the sibling repo [efficient-fpt](data/azhang/efficient-fpt) — specifically `get_addm_fptd_jax_fast` in [src/efficient_fpt_jax/multi_stage.py](data/azhang/efficient-fpt/src/efficient_fpt_jax/multi_stage.py).
+
+**Integration approach: vendor, do not depend.** Rather than add `efficient-fpt` as a dependency (it is not on PyPI, it ships compiled Cython for the simulator path that HSSM does not need, and bringing it in pulls a heavy build chain into HSSM's install), we **copy the relevant pure-JAX modules into HSSM** and own them going forward. HSSM already depends on `jax`/`jaxlib`, so the vendored code adds zero new transitive dependencies. The simulator (Cython) and NumPy/CPU paths from efficient-fpt are *not* vendored; only the `efficient_fpt_jax` subpackage's likelihood code.
+
+The goal is to wire the aDDM into HSSM so that users can write:
+
+```python
```
AlexanderFengler, line 11: here, arguably, we would go for an ADDM (or even ASSM) class instead. 
This would follow the RLSSM pattern.
@krishnbera we might want to keep RLASSM future capability in mind some of the decisions here?

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+model = hssm.HSSM(model="addm", data=addm_trial_df, include=[...])
+model.sample()
+```
+
+where `addm_trial_df` contains standard columns (`rt`, `response`) plus **aDDM-specific per-trial arrays** (item values, fixation onsets, fixation counts, first-fixation flag). The aDDM needs per-trial covariates that are *not* themselves sampled parameters — exactly the pattern RLSSM already solves in HSSM. We therefore follow the RLSSM design so that aDDM lives alongside it rather than carving a new architectural lane.
+
+The intended outcome is a working `aDDM(...)` class (and matching `aDDMConfig`) inside HSSM that (a) validates aDDM-specific trial data, (b) composes the vendored JAX FPT likelihood with sampled parameters `{eta, kappa, sigma, a, b, x0, t}` (non-decision time optional), (c) exposes the standard HSSM hierarchical regression and sampling machinery, and (d) ships with a tutorial notebook and unit tests.
```
AlexanderFengler, line 18: That's more like it, seems to contradict the earlier statement?!

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+
+> **Architecture update (post-rebase, commit `b4228d1b` — "HSSM base + RLSSM classes" #893).** The HSSM base was refactored: `HSSMBase` is now an abstract base ([src/hssm/base.py](data/azhang/HSSM/src/hssm/base.py)) and both `HSSM` ([src/hssm/hssm.py](data/azhang/HSSM/src/hssm/hssm.py)) and `RLSSM` ([src/hssm/rl/rlssm.py](data/azhang/HSSM/src/hssm/rl/rlssm.py)) are concrete subclasses. `RLSSMConfig` was moved out of `hssm.config` into [src/hssm/rl/config.py](data/azhang/HSSM/src/hssm/rl/config.py) and **no longer exposes a `to_config()` method** — `HSSMBase` accepts any `BaseModelConfig` directly, and the family-specific subclass is responsible for building its own likelihood `Op` and stamping it onto the config via `dataclasses.replace(...)` before calling `super().__init__()`. The new `RLSSMConfig` also adds an `ssm_logp_func` field (an `@annotate_function`-decorated JAX function) and renamed `learning_process_loglik_kind` → `learning_process_kind`.
+
+Given this new architecture, the plan creates **both**:
+
+1. An `aDDM(HSSMBase)` concrete subclass — a peer of `HSSM` and `RLSSM`, exported from `hssm/__init__.py` as `hssm.aDDM`.
```
AlexanderFengler, line 26: might want to name ADDM for camelcase. Slightly confusing why the LLM didn't suggest that out of the box.

High level comment: 
Which model are you using here?

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+   ```python
+   from .multi_stage import get_addm_fptd_jax_fast, pad_sacc_array_safely
+   from .single_stage import fptd_single_jax, q_single_jax
+   from .utils import GAUSS_LEGENDRE_30_X, GAUSS_LEGENDRE_30_W
+   ```
+4. Add a header comment to each vendored file recording the upstream commit hash from `efficient-fpt` so future maintainers can diff against upstream when bug fixes land there.
+5. **Do not** vendor `efficient_fpt_jax/batch.py` (not used by the per-trial likelihood we wrap), nor anything from the `efficient_fpt` (Cython/NumPy) subpackage — that path is the simulator and is not part of inference.
+6. **Do not** add `efficient-fpt` to `pyproject.toml`. HSSM already depends on `jax`/`jaxlib`, so the vendored code introduces no new dependencies.
+
+**License/attribution:** efficient-fpt ships under a permissive license (see `efficient-fpt/LICENSE`); copy that license text into `src/hssm/addm/likelihoods/jax/LICENSE` (or `NOTICE`) so attribution travels with the code. If HSSM and efficient-fpt share authors/license, a brief `# Adapted from efficient-fpt (commit <sha>)` header is sufficient.
```
AlexanderFengler, line 61: @frankmj note this licensing stuff. 
Didn't consider this when working on the addm paper code itself. Any advice on how to handle this?

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+
+**Rationale:** efficient-fpt is not on PyPI; its Cython compile chain is heavyweight and irrelevant to HSSM (HSSM only needs the inference likelihood, not the simulator); and pinning to a remote git dep would couple HSSM's CI to an unstable upstream. Vendoring lets HSSM ship a frozen, audited copy that evolves on HSSM's release cadence.
```
AlexanderFengler, line 63: @AndrewZhang599 here we might want to inject some context around parallel ssm-simulator work.

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+
+**Drift management:** efficient-fpt continues to be the research home for the likelihood. When upstream changes, the vendored copy can be re-synced by re-copying the three files and rebuilding tests. The upstream-commit header in step 4 makes "what version are we on?" trivially answerable. (This is the same pattern HSSM already uses for `bayesflow` from a dev branch, just snapshotted instead of git-tracked.)
```
AlexanderFengler, line 65: probably not necessary tbh

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+      multi_stage.py
+      single_stage.py
+      utils.py
+```
+
+**Rationale:** aDDM is conceptually a two-stage model (attention process → SSM likelihood) just like RLSSM (learning process → SSM likelihood). After the rebase, RLSSM is a real `HSSMBase` subclass with its own subpackage; the aDDM subpackage now mirrors that exactly — same files (`config.py`, `<model>.py`, `utils.py`, `likelihoods/builder.py`), same conventions. The vendored JAX code lives in its own `jax/` subdirectory so it stays clearly identifiable as upstream-derived, isolated from HSSM-original code in `builder.py` and `addm_jax.py`.
+
+### Step 3 — Add `aDDMConfig` dataclass in `src/hssm/addm/config.py`
+
+**Critical file:** [src/hssm/addm/config.py](data/azhang/HSSM/src/hssm/addm/config.py) (new file) — peer of [src/hssm/rl/config.py](data/azhang/HSSM/src/hssm/rl/config.py).
```
AlexanderFengler, line 93: genercal comment: These file-paths should not be personalized, so that the plan isn't machine specific.

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+        ...
+
+    def get_defaults(self, param):
+        return None, self.bounds.get(param)
+```
+
+**Key design decisions (post-rebase):**
+
+- **No `to_config()` method.** The new architecture has `HSSMBase` accept any `BaseModelConfig`; family-specific subclasses build the `loglik` `Op` themselves and stamp it onto the config via `dataclasses.replace(...)`. `RLSSMConfig` no longer has `to_config()` and neither will `aDDMConfig`.
+- **No `from_defaults` registration.** Like `RLSSMConfig`, `aDDMConfig` raises `NotImplementedError` from `from_defaults`. Users construct it explicitly (or via a `from_addm_dict` classmethod, optional). Therefore **`aDDM` is *not* registered through the `default_model_config` / `register_model` pipeline** that `HSSM(model="ddm", ...)` uses — instead, users instantiate `hssm.aDDM(...)` directly.
```
AlexanderFengler, line 142: I think this is something that we want to change about RLSSM class as well (@krishnbera). 

In fact, iiuc, we DO want the ability to go `.from_defaults()` and have a pre-registered `aDDM`.

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+- `extra_fields` defaults to the five aDDM-specific columns the JAX likelihood needs; these flow through the existing extra-fields machinery (data validator → `Op` `*args`) the same way they do for RLSSM.
+- `attention_process` is a pluggable hook (default `"standard_alternating"`) that maps `(r1, r2, flag, eta, kappa) → mu_array_padded` per trial. This mirrors `RLSSMConfig.learning_process` semantically (declarative documentation; the actual callable is resolved by the builder).
```
AlexanderFengler, line 144: this is important, and relates to above.

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+        # and feed the pre-built Op directly into make_distribution.
+        ...
+```
+
+**Key design decisions:**
+
+- **`aDDM` is a peer of `HSSM` and `RLSSM`**, all three inheriting from `HSSMBase`. It is exported from `hssm/__init__.py` as `hssm.aDDM`.
+- **`_make_model_distribution` is overridden** (same as `RLSSM`) because the aDDM `Op` already encapsulates the attention process + per-trial vmap; the standard `make_likelihood_callable` dispatch on `loglik_kind` should be bypassed.
+- **No `participant_col`-style panel reshape** — unlike RLSSM (which reshapes rows into `(n_participants, n_trials, ...)` because the RL learning rule is per-subject), aDDM's likelihood is per-trial. The vmap inside the JAX `Op` handles the trial dimension; participants flow through bambi/HSSM hierarchical regression as usual.
+- **`missing_data` / `deadline` are rejected up front**, same as RLSSM, because rearranging rows would break the strict trial→`sacc_array`-row correspondence.
```
AlexanderFengler, line 235: This is fine as v0, eventually we need to develop a strategy around it. 
We need to add an issue to address this later.

AlexanderFengler (commented):
The plan is already quite reasonable. 

One more high level comment:
When you address the feedback also please make sure you re-organize the plan around `commits` for the next iteration. Steps / Stages is the right idea, but we should really force the plan to operate on clean commit boundaries.

addm_andrew_dev /addm_hssm.md
krishnbera: RLASSM might be better served if ASSM exists as a model under HSSM class? Ofc, internally ASSM will have a separate branch to handle ASSM-specific things. This would avoid bloating the HSSM package with various interoperability layers (RL <-> SSM, RL <-> ASSM, etc.).

krishnbera: And conceptually speaking, ASSM is just a decision process with a specific kind of covariate after all.

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+
+Given this new architecture, the plan creates **both**:
+
+1. An `aDDM(HSSMBase)` concrete subclass — a peer of `HSSM` and `RLSSM`, exported from `hssm/__init__.py` as `hssm.aDDM`.
+2. An `aDDMConfig(BaseModelConfig)` dataclass living in `src/hssm/addm/config.py`, peer of `RLSSMConfig` in `src/hssm/rl/config.py`.
```
krishnbera, line 27: what would a canonical aDDM config look like? Giving/generating an example here would help the coding agents/docs/human implementation.

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+
+    @classmethod
+    def from_defaults(cls, model_name, loglik_kind):
+        raise NotImplementedError(
+            "aDDMConfig does not support from_defaults(). "
+            "Use the aDDM(...) constructor directly, or pass an aDDMConfig "
+            "instance built explicitly."
+        )
+
+    def validate(self) -> None:
```
krishnbera, line 130: we can be explicit about the validation checks we need here.

addm_andrew_dev /addm_hssm.md
krishnbera: yeah, this registration/name-based referencing will be implemented soon for RLSSM. so it naturally extends here as well.

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+
+Mirror [hssm/rl/likelihoods/builder.py](data/azhang/HSSM/src/hssm/rl/likelihoods/builder.py) (`make_rl_logp_func`, `make_rl_logp_op`).
+
+Two functions:
+
+1. **`make_addm_logp_func(attention_process)`** — returns a callable `logp(data, *args)` where:
+   - `data[:, 0]` = rt, `data[:, 1]` = response
+   - `args` are sampled parameters in `list_params` order: `eta, kappa, sigma, a, b, x0`
+   - extra fields `r1, r2, sacc_array, d, flag` are appended to `args` by the HSSM extra-field machinery (exactly as RLSSM does; see [data_validator.py:156](data/azhang/HSSM/src/hssm/data_validator.py#L156)).
+   - Internally: call the attention process to build `mu_array_padded`, then call `get_addm_fptd_jax_fast(t=rt, d=d, mu_array=mu_array, sacc_array=sacc_array, sigma, a, b, x0)` vmapped over trials.
```
krishnbera, line 249: @AlexanderFengler @AndrewZhang599 
I am not too familiar with aDDM, but would be worthwhile to think about the fundamental dimension for vmap here. Is it possible to do better than trial dim?

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
...
+from .utils import validate_addm_panel
+
+
+class aDDM(HSSMBase):
+    def __init__(
+        self,
+        data: pd.DataFrame,
+        model_config: aDDMConfig,
+        include: list | None = None,
+        p_outlier: float | dict | bmb.Prior | None = 0.05,
```
krishnbera, line 177: i assume outliers must be a critical aspect of aDDM (i believe there is a measurement component to it unlike the feedback component in RLSSM). are there basic validation checks we can make on this front? for v0, it is ok to ignore this aspect.

krishnbera (changes requested):

addm_andrew_dev /addm_hssm.md
```diff
@@ -0,0 +1,405 @@
+# Plan: Integrating aDDM into HSSM
+
+## Context
+
+The attentional drift diffusion model (aDDM; Krajbich et al.) extends the standard DDM by modulating the drift rate based on **which option the subject is currently fixating**. A fast, differentiable JAX likelihood for the aDDM has been prototyped in the sibling repo [efficient-fpt](data/azhang/efficient-fpt) — specifically `get_addm_fptd_jax_fast` in [src/efficient_fpt_jax/multi_stage.py](data/azhang/efficient-fpt/src/efficient_fpt_jax/multi_stage.py).
+
+**Integration approach: vendor, do not depend.** Rather than add `efficient-fpt` as a dependency (it is not on PyPI, it ships compiled Cython for the simulator path that HSSM does not need, and bringing it in pulls a heavy build chain into HSSM's install), we **copy the relevant pure-JAX modules into HSSM** and own them going forward. HSSM already depends on `jax`/`jaxlib`, so the vendored code adds zero new transitive dependencies. The simulator (Cython) and NumPy/CPU paths from efficient-fpt are *not* vendored; only the `efficient_fpt_jax` subpackage's likelihood code.
```
cpaniaguam, line 7: Who owns/maintains the `efficient-fpt` package? I think it'd be sensible to split it into two parts, the simulators which are Cython dependent and the rest which is what's needed for this use case. This would avoid passing on the heavy Cython dependency to downstream projects unnecessarily. The new lightweight package could serve as a common dependency for both `hssm` and `efficient_fpt`.

addm_andrew_dev /addm_hssm.md
cpaniaguam: I think using the relevant parts as a dependency would solve this(?). See https://github.com/lnccbrown/HSSM/pull/953/changes#r3210200016

addm_andrew_dev /addm_hssm.md
cpaniaguam: Why not add this package (and the proposed lightweight dependency) to PyPI?

addm_andrew_dev /addm_hssm.md
AlexanderFengler: We prob won't keep `efficient-fpt` around (it's a repo attached to a methods paper), just extract the core likelihoods into HSSM.


The forward simulation part of `efficient-fpt` will go into `ssm-simulators`.

addm_andrew_dev /addm_hssm.md
digicosmos86: If there is not much conceptual difference from RLSSM, not adding a new class is probably the best idea, so the focus can just be on writing the decision process function

addm_andrew_dev /addm_hssm.md
digicosmos86: Yes, this is confusing

AlexanderFengler:
Superseded by #967 (`addm-integration-0`), which now carries the full first-class aDDM implementation. The RLSSM base refactor this PR proposed already landed on `main` via #893; this branch carries no aDDM source. Closing to remove ambiguity between the two addm-integration* branches.

PR #955, opened 2026-05-06
Cpaniaguam/rlssm simplified interface

cpaniaguam:
```mermaid
flowchart TD
    subgraph SSM_Registry
      A1["_SSM_REGISTRY (dict)"]
      A2["_SSM_LOGP_CACHE (dict)"]
    end
    subgraph RLSSM_Registry
      B1["_RLSSM_REGISTRY (dict)"]
    end
    subgraph User_API
      C1[register_ssm]
      C2[register_rlssm_model]
      C3[get_rlssm_model_config]
    end
    C1-->|adds entry|A1
    C1-->|adds entry|A2
    C2-->|adds entry|B1
    C3-->|reads entry|B1
    C3-->|reads entry|A1
    C3-->|calls _get_ssm_logp|A2
    A2-->|lazy build if needed|A1
    C3-->|returns RLSSMConfig|D1["RLSSMConfig"]
    style D1 fill:#1e7a1e,stroke:#fff,stroke-width:2px,color:#fff
    style A1 fill:#1e3a7a,stroke:#fff,stroke-width:2px,color:#fff
    style A2 fill:#1e3a7a,stroke:#fff,stroke-width:2px,color:#fff
    style B1 fill:#7a1e1e,stroke:#fff,stroke-width:2px,color:#fff
    style C1 fill:#7a7a1e,stroke:#fff,stroke-width:2px,color:#fff
    style C2 fill:#7a7a1e,stroke:#fff,stroke-width:2px,color:#fff
    style C3 fill:#7a7a1e,stroke:#fff,stroke-width:2px,color:#fff

### CODE REVIEW

File: tests/rl/test_rlssm.py [hunk 1]
```diff
@@ -327,3 +327,95 @@ def test_rlssm_pickle_round_trip(
...
+        name="rldm_custom_test",
+        decision_process="angle",
+        learning_process={"v": _compute_v_annotated},
+        rl_params=["rl_alpha", "scaler"],
+        rl_bounds={"rl_alpha": (0.0, 1.0), "scaler": (0.0, 10.0)},
+        rl_params_default=[0.1, 1.0],
+        extra_fields=["feedback"],
+        choices=[0, 1],
+    )
+    model = RLSSM(data=rldm_data, model="rldm_custom_test")
```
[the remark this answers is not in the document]
cpaniaguam: Name is unique already.

src/hssm/rl/__init__.py
```diff
@@ -5,9 +5,12 @@
...
 
-- ``RLSSM``: the RL + SSM model class implemented in :mod:`hssm.rl.rlssm`.
-- ``RLSSMConfig``: the config class for RL + SSM models, implemented in
-  :mod:`hssm.rl.config`.
+- ``RLSSM``: the public RL + SSM model class in :mod:`hssm.rl.rlssm`.
+- ``_RLSSM``: the internal base class that requires a fully built config.
+- ``RLSSMConfig``: the config class for RL + SSM models in :mod:`hssm.rl.config`.
+- ``get_rlssm_model_config``: factory that builds a config from a named model.
+- ``register_rlssm_model``: register a custom named RLSSM model.
+- ``register_ssm``: register a custom SSM base logp function.
```
AlexanderFengler, line 12: would we want/need a `register_learning_process` equivalent here? ( @krishnbera )

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,447 @@
...
+def _make_angle_base_logp() -> Any:
+    """Build the annotated angle SSM base logp function from its ONNX file."""
+    _raw = make_jax_matrix_logp_funcs_from_onnx(model="angle.onnx")
+    return annotate_function(
+        inputs=["v", "a", "z", "t", "theta", "rt", "response"],
+        outputs=["logp"],
+    )(_raw)
+
+
+_SSM_REGISTRY: dict[str, dict[str, Any]] = {
```
AlexanderFengler: This patterns is kind of interesting. On some level seems like duplicate work on the SSM side, since we have these registries already?

1. Can we refer to the original ssm registry with this?
2. If we do this, we should, as a starter pack, for sure include a few more models than just the angle model ("ddm", "weibull" e.g.)

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,447 @@
...
+        "rl_params_default": list(rl_params_default),
+        "extra_fields": list(extra_fields) if extra_fields is not None else [],
+        "choices": list(choices) if choices is not None else [0, 1],
+        "description": description,
+        "decision_process_loglik_kind": decision_process_loglik_kind,
+        "learning_process_kind": learning_process_kind,
+    }
+
+
+def register_ssm(
```
AlexanderFengler, line 617: here also slightly confused about code location. 
`register_ssm` is modular and could be sitting in a different folder?

AlexanderFengler (commented):
Left a few comments, mostly looks good already. 
@krishnbera can you take a look too?

cpaniaguam:
```mermaid
flowchart TD
    USER["User: RLSSM(model='rldm')"]

    RLSSM_REG["_RLSSM_REGISTRY\nnamed models\ne.g. 'rldm'"]

    SSM_REG["_SSM_REGISTRY\ncustom SSMs only\n(empty by default)"]

    MODELCONFIG["hssm.modelconfig\nbuilt-in SSMs\nddm · angle · weibull · ornstein · …"]

    CACHE["_SSM_LOGP_CACHE\nlazy ONNX → JAX fn"]

    OUTPUT["RLSSMConfig"]

    USER --> RLSSM_REG
    RLSSM_REG -- decision_process name --> LOOKUP

    LOOKUP{"registered\ncustom SSM?"}
    LOOKUP -- yes --> SSM_REG
    LOOKUP -- no  --> MODELCONFIG

    SSM_REG & MODELCONFIG --> CACHE
    CACHE -- annotated JAX fn --> OUTPUT
    RLSSM_REG -- rl params / bounds / learning process --> OUTPUT
```

src/hssm/rl/registry.py
cpaniaguam: `register_ssm` needs to stay here because it manages the module-level `SSM_REGISTRY` and `SSM_LOGP_CACHE`, which hold the global state for `SSMs`. Keeping it local ensures consistent state and avoids cross-module issues.

src/hssm/rl/__init__.py
krishnbera: yes we should have a way of registering the learning process separately.

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,597 @@
...
+
+def _make_ssm_base_logp_from_onnx(
+    onnx_file: str,
+    list_params_ssm: list[str],
+    response: list[str],
+) -> Any:
+    """Build and annotate a JAX log-likelihood function from an ONNX model file."""
+    _raw = make_jax_matrix_logp_funcs_from_onnx(model=onnx_file)
+    return annotate_function(
+        inputs=list_params_ssm + response,
```
krishnbera, line 180: is concatenation of response correct here? this should also have rt?

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,597 @@
...
+    return _SSM_LOGP_CACHE[name]
+
+
+# ---------------------------------------------------------------------------
+# RLSSM named model registry
+# ---------------------------------------------------------------------------
+# Each entry provides:
+#   decision_process            - key into _SSM_REGISTRY
+#   learning_process            - {param: annotated_func}
+#   rl_params                   - ordered list of sampled RL parameter names
```
krishnbera: quick nomenclature update: let's call this 'learning_process_params' or 'lp_params' for short.

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,597 @@
...
+#   rl_bounds                   - {param: (lo, hi)} for RL params
```
krishnbera: same here — 'lp_bounds'

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,597 @@
...
+#   rl_params_default           - default values aligned with rl_params
```
krishnbera: ditto

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,597 @@
...
+            "Pass the raw (base) annotated function without .computed instead."
+        )
+    return annotate_function(
+        inputs=ssm_base_logp_func.inputs,
+        outputs=ssm_base_logp_func.outputs,
+        computed=learning_process,
+    )(ssm_base_logp_func)
+
+
+def _derive_rl_params(
```
krishnbera: could be renamed to `_derive_lp_params`

krishnbera (dismissed):
looks good overall. added minor comments.

src/hssm/rl/registry.py
cpaniaguam: `rt` is accounted for within the `response` variable: `response = ["rt", "response"]`.

src/hssm/rl/__init__.py
```diff
@@ -17,11 +20,20 @@
...
+    get_rlssm_model_config,
+    register_rlssm_model,
+    register_ssm,
+)
+from .rlssm import _RLSSM, RLSSM
 from .utils import validate_balanced_panel
 
 __all__ = [
     "RLSSM",
+    "_RLSSM",
```
digicosmos86: The name "_RLSSM" suggests that this class is "private" and should not directly be accessed by the user, so it should not be exposed in `__all__`

src/hssm/rl/rlssm.py
```diff
@@ -262,3 +282,211 @@ def _make_model_distribution(self) -> type[pm.Distribution]:
...
+            return obj.__dict__.get(self._storage_key, False)
+        raise NotImplementedError(self._message)
+
+    def __set__(self, obj: Any, value: Any) -> None:  # noqa: D105
+        # Store the value so internal reads during __init__ work correctly.
+        obj.__dict__[self._storage_key] = value
+
+
+class RLSSM(_RLSSM):
+    """Reinforcement Learning Sequential Sampling Model — simplified public API.
```
digicosmos86: This class is public facing - the `simplified public API` may cause confusion for the user - they don't know what the "not simplified API" is

src/hssm/rl/rlssm.py
```diff
@@ -262,3 +282,211 @@ def _make_model_distribution(self) -> type[pm.Distribution]:
...
+
+    This class wraps :class:`_RLSSM` with a user-friendly constructor that
+    accepts a *model* name string (looked up in the named-model registry) and
+    optional overrides for *learning_process*, *decision_process*, and
+    *choices*.  Advanced users can bypass the registry entirely by supplying a
+    pre-built *model_config*.
```
digicosmos86: Again, this is public-facing API documentation, and this does not mean anything to the user if they are not already familiar with `_RLSSM`

src/hssm/rl/rlssm.py
```diff
@@ -262,3 +282,211 @@ def _make_model_distribution(self) -> type[pm.Distribution]:
...
+
+        # Mark initialisation complete — after this point _BlockedAttribute
+        # raises NotImplementedError on any read of missing_data / deadline /
+        # loglik_missing_data.
+        # IMPORTANT: This MUST be the last statement in __init__.  Any code
+        # added after this line cannot read self.missing_data / self.deadline.
+        self.__dict__["_rlssm_fully_initialized"] = True
+
+    @classproperty
+    def list_models(cls) -> dict[str, str | None]:
```
digicosmos86, line 488: This function has a different name than its counterpart in `HSSM`. Instead of making it a `classproperty`, it doesn't really depend on the class. How about making it and its counterpart in `HSSM` a method accessible through `hssm` and `hssm.rl` respectively?

digicosmos86 (dismissed):
Generally it looks good. The use of `_RLSSM` as a separate class is an interesting choice, but unless someone is aware of how this separation came into being, it's difficult to tease out what functions each class serve from their names alone. It might make it harder to find where things are in the future, and it creates separate sets of documentation that can easily get out of sync. I would merge these two classes

src/hssm/rl/rlssm.py
cpaniaguam: Good point! Addressed in #971.

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,598 @@
...
+            "To add a custom model, use register_rlssm_model() "
+            "(and register_ssm() for custom decision processes), "
+            "or pass 'model_config=' directly to RLSSM()."
+        )
+
+    # Shallow-copy so overrides don't mutate the registry entry.
+    entry = dict(_RLSSM_REGISTRY[model])
+
+    if learning_process is not None:
+        entry["learning_process"] = learning_process
```
krishnbera: get_rlssm_model_config() accepts a learning_process override but still uses the registry’s learning_process_params, learning_process_bounds, and learning_process_params_default unless those keys are absent (None).

Impact: RLSSM(data, model="2AB_RescorlaWagner_DDM", learning_process=new_lp) can build a config where ssm_logp_func.computed reflects the new LP, but list_params / bounds / params_default still describe the old one. Depending on the mismatch, you get either a late validate() failure or a model that constructs but samples the wrong parameters.

Fix direction: When learning_process is overridden, re-derive LP params (and ideally bounds/defaults) unless the caller also explicitly overrides those fields.

Missing test: No test for learning_process= or decision_process= overrides on the public RLSSM() constructor.

src/hssm/rl/registry.py
```diff
@@ -0,0 +1,598 @@
...
+#                               - default values aligned with learning_process_params
+#   extra_fields                - extra data column names required by LP
+#   choices                     - response choice values
+#   description                 - human-readable description
+#   decision_process_loglik_kind
+#   learning_process_kind
+
+_RLSSM_REGISTRY: dict[str, dict[str, Any]] = {
+    "2AB_RescorlaWagner_DDM": {
+        "decision_process": _get_decision_process_spec("ddm"),
```
krishnbera: [Minor]
Starter-pack models store a fully resolved decision-process dict at module import. Later register_ssm("ddm", ...) will not affect 2AB_RescorlaWagner_DDM unless the user passes decision_process="ddm" as an override (which re-resolves via _SSM_REGISTRY).

krishnbera (commented):
lgtm -- just one major comment!!

src/hssm/rl/rlssm.py
```diff
@@ -262,3 +282,212 @@ def _make_model_distribution(self) -> type[pm.Distribution]:
...
+    loglik_missing_data = _BlockedAttribute(  # type: ignore[assignment]
+        "loglik_missing_data",
+        "RLSSM does not support 'loglik_missing_data'. "
+        "Missing-data network assembly (OPN / CPN) is not implemented for RLSSM.",
+    )
+
+    def __init__(
+        self,
+        data: pd.DataFrame,
+        model: str | None = None,
```
digicosmos86: Should this default to a `str` instead of `None`?

src/hssm/rl/rlssm.py
```diff
@@ -262,3 +282,212 @@ def _make_model_distribution(self) -> type[pm.Distribution]:
...
+    decision model in a single likelihood. In the common case, you choose a
+    named RLSSM model with ``model`` and optionally override its
+    ``learning_process``, ``decision_process``, or ``choices`` settings. Use
+    :attr:`RLSSM.list_models` to inspect the named models available in HSSM.
+
+    If you already have a fully built :class:`RLSSMConfig`, you can pass it as
+    ``model_config`` instead of selecting a named model.
+
+    RLSSM currently requires balanced panel data and does not support
+    ``missing_data``, ``deadline``, or ``loglik_missing_data`` handling.
```
digicosmos86, line 348: rst-style markups do not work with mkdocs which relies on Markdown. The docs are rendered like this:

Image

BTW we should add RLSSM to docs. Maybe a new PR

digicosmos86 (dismissed):
LGTM now. @AlexanderFengler @krishnbera what's still blocking this from being merged?

src/hssm/rl/rlssm.py
cpaniaguam: Thanks!

src/hssm/rl/registry.py
cpaniaguam: Addressed by moving the learning process metadata into a dedicated `LearningProcessMetadata` class instance and making `get_rlssm_model_config()` resolve overrides through that object instead of reusing flat registry fields.

When `learning_process=` is overridden, we now rederive the sampled LP params from the new annotated functions and only carry forward bounds/defaults for params that still match. We fail early instead of constructing an inconsistent config if the override introduces sampled params without corresponding metadata, 

I also added focused tests for the public `RLSSM(...)` path covering both `learning_process=` and `decision_process=` overrides.

krishnbera (approved):

PR #960, opened 2026-05-12
[hot fix] restore choice only changes

cpaniaguam:
Restore choice-only safeguards.


digicosmos86 (approved):
LGTM! Thanks @cpaniaguam!

PR #962, opened 2026-05-15
tests: add xfail tests for float lapse_dist bugs in _apply_lapse_model

cpaniaguam:
@AlexanderFengler @digicosmos86 Would it be sensible to address this?

**Summary**

Adds two `xfail(strict=True)` tests that document two latent bugs in `_apply_lapse_model` (in dist.py). The tests are currently expected to fail and will flip to passing once a fix is added.

---

**Background**

For choice-only models (e.g. `softmax_inv_temperature_2`), `_check_lapse` sets `self.lapse = 1 / n_choices` — a plain `float`. This float is stored as `HSSMRV._lapse` and passed to `_apply_lapse_model(lapse_dist=...)` during prior/posterior predictive sampling. The existing code assumes `lapse_dist` is always a `bmb.Prior` object, so two things go wrong.

---

**Bug 1 — `AttributeError` on `lapse_dist.args`**

`_apply_lapse_model` always executes:

```python
lapse_rt = pm.draw(
    get_distribution_from_prior(lapse_dist).dist(**lapse_dist.args), ...
)
```

When `lapse_dist` is `0.5` (a `float`), `lapse_dist.args` raises:

```
AttributeError: 'float' object has no attribute 'args'
```

**Trigger:** any call to `model.sample_prior_predictive()` or `model.sample_posterior_predictive()` on a choice-only model that has `p_outlier != 0`.

---

**Bug 2 — 2-column replace mask for a 1-column output**

Even if Bug 1 were somehow bypassed, the replacement mask is always built as:

```python
replace_mask = np.stack([replace, replace], axis=-1)  # always 2 columns
```

Choice-only simulation output has only **1 column** (response; no RT). The 2-column mask either silently corrupts the array via `np.putmask` broadcasting or raises a shape error depending on input dimensions.

**Trigger:** same as Bug 1.

---

**Current status**

Both bugs are latent on the `add-choice-only-rlssm-model-to-registry` branch. For RLSSM models `ssms_rng_fn` raises before reaching `_apply_lapse_model` (ssm_simulators does not support `2AB_RescorlaWagner_Softmax`), so neither bug is reachable through the RLSSM path today. They would be reachable for any other choice-only model supported by ssm_simulators (e.g. `softmax_inv_temperature_2`) the moment a user passes `p_outlier != 0`.

---

**Fix (tracked separately)**

Add an `isinstance(lapse_dist, float)` branch in `_apply_lapse_model` that:
- skips RT sampling (there is no RT column)
- samples replacement responses uniformly over `choices`
- handles both 1-D `(n_obs,)` and 2-D `(n_obs, 1)` `sims_out`

The type annotation of `make_hssm_rv`'s `lapse` parameter should also be updated from `bmb.Prior | None` to `bmb.Prior | float | None`.

### DIFF
--- a/tests/distribution_utils/test_distribution_utils.py
+++ b/tests/distribution_utils/test_distribution_utils.py
@@ -12,6 +12,7 @@ from hssm.distribution_utils.dist import (
     ensure_positive_ndt,
     make_distribution,
     make_distribution_for_supported_model,
+    _apply_lapse_model,
     _create_arg_arrays,
     _extract_size,
     _get_p_outlier,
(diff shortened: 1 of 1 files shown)


AlexanderFengler (approved):
@cpaniaguam lgtm. 

Definitely need to prioritize fixing this, but understood that's taken care of via a different PR.

PR #966, opened 2026-05-19
fix: warn on disconnected RVs from centered/noncentered prior drift

AlexanderFengler:
## Summary

When a user supplies group-specific priors with **nested hyperpriors** (e.g. a Normal prior on `1|participant_id` whose `mu` is itself a `Prior`), HSSM/bambi can silently produce a PyMC graph with **idle, disconnected nodes**. Under the default `noncentered=True`, bambi reparameterizes a group-specific Normal term as `offset * sigma` and never references `mu` — so a user-supplied `mu` hyperprior is created as a free RV in the graph but never wired into the likelihood. Until now this happened silently.

This PR adds two validation hooks and a tutorial.

- **Targeted check** (`src/hssm/param/parameterization_check.py::check_user_priors_against_parameterization`): scoped to user-supplied prior keys, flags Normal group-specific priors with a `mu` hyperprior when the effective parameterization is non-centered, and emits an actionable warning ("either pass `noncentered=False`, or move the location prior onto the common `Intercept`…").
- **General safety net** (`find_disconnected_free_rvs`): walks the PyMC graph after `model.build()` and warns about any free RV that is not an ancestor of an observed RV — catches future name-convention drift, not just the known case.
- Wired into `src/hssm/base.py` around the existing `bmb.Model(...)` / `self.model.build()` block. `noncentered` remains a `**kwargs` passthrough.
- New tutorial `docs/tutorials/centered_vs_noncentered_basic_logic.ipynb` walks through both parameterizations, the node-naming difference, the disconnected-node footgun, and two fixes.

Severity: warn (not raise) — model still builds. The default `noncentered=True` and HSSM behavior are otherwise unchanged.

## Test plan

- [x] `uv run pytest tests/test_prior.py -v` — 8/8 pass (5 new cases covering targeted warning, centered no-warning, defaults-skip, hand-built orphan detection, unit-level check)
- [x] `uv run pytest tests/ --ignore=tests/slow -x` — 467/467 pass; no regressions from the new hook in `HSSMBase.__init__`
- [x] `uv run ruff check` clean on touched files (remaining ruff errors are pre-existing in `tests/test_prior.py`)
- [x] `uv run mypy src/hssm/param/parameterization_check.py` — clean
- [ ] Manual: build the tutorial notebook end-to-end and confirm graphviz output for each model

## Summary by CodeRabbit

* **New Features**
  * Added a new tutorial explaining centered vs. non-centered parameterizations and how to interpret related model graphs.

* **Bug Fixes**
  * Added warnings for potentially problematic prior setups, including disconnected effects and non-identifiable combinations with a common intercept.
  * Improved detection of hidden model issues so they can be surfaced earlier during setup.

* **Documentation**
  * Added the tutorial to the docs navigation.


docs/tutorials/centered_vs_noncentered_basic_logic.ipynb
```diff
@@ -0,0 +1,1332 @@
+{
```
krishnbera, line 1: <p>minor comment: expand upon the description of location prior a bit? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/lnccbrown/HSSM/pull/966/discussion/#comment-3282498959'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='22'/>

src/hssm/param/parameterization_check.py
```diff
@@ -0,0 +1,291 @@
...
+                component_name=param_name,
+                prior_noncentered=getattr(prior, "noncentered", None),
+            )
+            if not effective_nc:
+                continue
+            mismatches.append(
+                PriorMismatch(
+                    parameter=param_name,
+                    term=term_name,
+                    reason=(
```
krishnbera: we can explicitly suggest adding a prior on offset here?

krishnbera (dismissed):
lgtm!!

Issue #16, opened 2022-11-29
[New Features] Summary statistics and graphing


digicosmos86:
We are using this issue to track the features that needed to be customized to HSSM vs. those already offered by arviz (even though we might want wrappers around them).

PR #968, opened 2026-05-25
Plan hssm hmm class

### LINKED ISSUE #957: HSSM-HMM Class
We have a basic Tutorial to illustrate how we can use our HSSM random variables to construct basic HMMs with DDM emissions [here](https://lnccbrown.github.io/HSSM/tutorials/hmm_ddm_regime_switching/#part-2-simulate-regime-switching-ddm-data).

We received quite a bit of demand for applied projects to make use of this type of functionality, since the HMM framework allow a very different approach to deal with lapse trials than our basic mixture model approach.

The goal of this issue to design a HMM class that we can use to construct these types of HSSM-HMM models automatically. At best we can pivot patterns we developed in the HSSM and RLSSM classes, subclass and keep additional implementation needs minimal. 

The first deliverable here is a rigorous (and companion high level) planning document with the design of this class.

fmuia:
Closes #957

First version ready for discussion/iteration. @AlexanderFengler


docs/design/hssm_hmm_overview.md
```diff
@@ -0,0 +1,122 @@
...
+- **Model construction.** Built directly in PyMC. bambi — used by `HSSM`/`RLSSM` — cannot represent the HMM's regime-indexed latents, so it is deferred until per-regime regressions are scoped.
+- **Lapse handling.** v1 rejects the `p_outlier` mixture; lapses are modeled structurally, as a regime.
+- **Emission backends.** Analytical and LAN (`approx_differentiable`) are both validated; the LAN runs on the `jax` (default) and `pytensor` backends.
+- **Scaling.** Multi-participant panels run as one batched recursion, not a per-subject loop (which does not scale).
+
+See Section 10 of the rigorous doc for the full list (nine decisions) and the rationale behind each.
+
+### Still open
+
+- **Default sampler.** Tutorial uses `nuts_sampler="numpyro"`; proposal is to keep PyMC NUTS as the code default (`HSSM_HMM` overrides `sample()` to call `pm.sample` directly) and document the numpyro recommendation in the tutorial.
```
AlexanderFengler, line 118: we can try numba backend to see if this makes the scan function much faster without relying on JAX, but from what I saw, JAX is literally so much faster for the `.scan` operation, that we essentially need "numpyro" backend for this. 

So should def be the default.

AlexanderFengler (commented):
Not a proper review yet just getting an idea.

docs/design/hssm_hmm_overview.md
```diff
@@ -0,0 +1,122 @@
...
+**Issue:** [#957](https://github.com/lnccbrown/HSSM/issues/957)
+**Companions:** [hssm_hmm.md](./hssm_hmm.md) (full rigorous design); [hssm_hmm_math.md](./hssm_hmm_math.md) (full math spec — v1 model in §14; planned Phase 6.1 hierarchical extension in §5–13)
+
+This is a short companion to the rigorous design document. Audience: applied users, project stakeholders, anyone deciding whether the proposed feature meets their needs without reading the math.
+
+---
+
+## What we're building
+
+A new top-level class `HSSM_HMM` that fits **regime-switching sequential sampling models**. The model assumes:
```
AlexanderFengler, line 12: Underscore in class names is probably not good. 
It's a little long on the capital letters now, but probably need to go for HMMSSM, of HSSMHMM (or if there are better ideas also glad to take those).

docs/design/hssm_hmm_overview.md
```diff
@@ -0,0 +1,122 @@
...
+---
+
+## What you'll be able to do (v1)
+
+Fit a regime-switching DDM (or any SSM HSSM supports) in three lines:
+
+```python
+model = hssm.HSSM_HMM(
+    data=df,
+    decision_process="ddm",
```
AlexanderFengler, line 39: here we face some subtlety. 

Currently we have the `HSSM`, and `RLSSM` classes, which have slightly different semantics. 

`HSSM` class uses `model` as the argument to pick the cognitive process, while `RLSSM` (for reinforement learning / sequential sampling models) actually picks the `decision_process` (underlying cognitive process model), and the `learning_process` (operating on one or more of the parameters of the cognitive process model). 

In the spirit of `v0` we probably want to shirk the complexity added by the `RL` process part and use the `model` argument here for now. 

Eventually some global renaming will occur to fix this tension.

docs/design/hssm_hmm_overview.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+What you get back:
+- Posterior distributions over the SSM parameters (per-regime for switching ones, shared otherwise) and the transition matrix.
+- Per-trial posterior probabilities of each regime, recovered post-hoc.
+- Standard post-fit tools — `summary`, trace plots, `find_MAP` — work as inherited. A few methods are HMM-specific (`infer_regimes`, per-trial log-likelihood) or out of v1 scope (posterior-predictive simulation, which a regime-switching model needs a bespoke routine for); see Section 6.3 of the rigorous doc.
+
+---
+
+## v1 boundaries
```
AlexanderFengler, line 54: upstairs used `v0` for what is actually called `v1` according to terminology here, my bad.

docs/design/hssm_hmm_overview.md
```diff
@@ -0,0 +1,122 @@
...
+
+We've drawn a tight boundary for v1 so the first version ships quickly and matches the existing tutorial's behavior exactly.
+
+**Included in v1:**
+- Any number of regimes `K >= 2`.
+- Any subset of SSM parameters can switch by regime.
+- Both analytical (DDM) and LAN-based (approximate differentiable) emission backends.
+- Balanced panels of participants — every participant must have the same number of trials.
```
AlexanderFengler, line 62: Don't get this part as a restriction, because so far this doesn't have hierarchy --> fitting "participant wise" anyways?

docs/design/hssm_hmm_overview.md
```diff
@@ -0,0 +1,122 @@
...
+- Sticky-Dirichlet prior on the transition matrix (the standard "regimes persist" prior).
+- Automatic handling of label-switching (no manual ordering constraint needed).
+- Forward-Filter Backward-Sample (FFBS) regime recovery as a method.
+
+**Deferred to later versions** (all with named extension hooks in the architecture):
+- Hierarchical pooling of regime-specific or transition parameters across participants.
+- Transition matrices that depend on trial-level covariates.
+- Per-regime priors or regressions for switching parameters.
+- Estimable initial-state distribution.
```
AlexanderFengler, line 71: Is this one adding a lot of complexity? Seems like this could fit into `v1`.

docs/design/hssm_hmm_overview.md
```diff
@@ -0,0 +1,122 @@
...
+- Mixed-emission models (different regimes use different SSMs).
+- Semi-Markov regime durations.
+- The per-trial lapse mixture (`p_outlier` / `lapse`). v1 rejects these inherited arguments outright — the HMM handles lapses *structurally*, as a regime (see "Why this matters" above), so the iid mixture would double-model the same effect.
```
AlexanderFengler, line 74: I would consider this as part of `v1` as well. E.g. the regime could be across `p_outlier` parameter shifts (one regime has low `p_outlier`, one regime has high `p_outlier`). 

A priori I think this shouldn't make things much harder, and it would help with forcing through the pre-existing handlers for lapse distributions etc., so it likely forces the design patterns to be closer to pre-existing ones.

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+
+**The model.** Each trial belongs to one of `K` hidden regimes. The regime sequence evolves as a Markov chain (`K×K` transition matrix `P`, initial distribution `pi0`). Within a regime, the trial's `(rt, response)` is generated by a standard SSM (DDM, LBA, ...) with regime-specific values for the *switching* parameters and shared values for the rest. Participants are independent in v1.
+
+**The core challenge.** NUTS only differentiates continuous parameters, so the discrete regimes are marginalised out via the forward algorithm. This produces a single *scalar* log-likelihood per evaluation — which **bypasses** `make_distribution`'s per-row contract that HSSM assumes everywhere else (see §2.2, §3.4). The plan reconciles this by reconstructing per-trial values post-hoc, after sampling.
+
+**The architecture in one picture.**
+
+```
+SAMPLE-TIME  (each NUTS leapfrog step):
+  one batched pytensor.scan over T       (participant axis vectorised in tensors)
```
AlexanderFengler, line 21: ok seeing now where the "equal number of trials" aspect came from. 

For the RLSSM class we are doing something like this, and they key here should be that we have some `no ops` so that unequal trial numbers can work.

If that turns out to be really complicated, we can skip for `v1`.

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+    -> scalar L(theta)
+    -> pm.Potential("hmm_loglik", L)
+    -> idata.posterior(theta_switching, theta_shared, P, pi0)
+
+POST-HOC  (called once by the user, eager NumPy):
```
AlexanderFengler, line 26: If it's possible at all, maybe we can drop all the NumPy stuff.

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+
+**User-facing API.**
+
+```python
+from hssm import HSSM_HMM
+
+model = HSSM_HMM(
+    data=df,                          # long-format, balanced panel, sorted (participant, trial)
+    K=2,
+    decision_process="ddm",
```
AlexanderFengler, line 65: left longer comment on this in the `overview.md` doc.

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+The first version targets the **simplest useful case** so that we can:
+- Ship something usable to applied users quickly.
+- Validate the architectural choices against a concrete implementation.
+- Match the tutorial's numerical results as a regression test.
+
+**v1 includes:**
+- Arbitrary number of regimes `K >= 2`.
+- A subset of SSM parameters that switch by regime (declared as a flat list).
+- All non-switching SSM parameters shared across regimes; standard HSSM priors / regressions for them.
+- Balanced panel of participants (every participant has the same number of trials). v1 has **no per-participant parameter structure**: switching and shared parameters are global (the switching ones are length-K vectors indexed by regime, the shared ones are scalars). The only per-participant quantity is the latent regime sequence `s_{n,1:T}`. Hierarchical pooling of regime-specific parameters across participants is a deferred Phase 6.1 extension (§8).
```
AlexanderFengler, line 100: Wondering here if full-pooling vs. per participant parameters should both work (neither with hierarchy yet)?

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+log_alpha_new[n, k] = logsumexp_j ( log_alpha_prev[n, j] + log_P[j, k] )
+                     + log_emission[n, t, k]
+```
+
+This is one `pytensor.scan` over `T` (not N) — a single JAX-JIT compile unit
+overall — for both the analytical and LAN emission backends (decision 10.1.6).
+
+The joint marginal `sum_n L_n` is then `pt.sum(pt.logsumexp(log_alpha_T, axis=1))` — one scalar contributed to the model as a single `pm.Potential`.
+
+**Why not a Python for-loop over participants.** The feasibility prototype tried the obvious "loop over n in Python, build one `pytensor.scan` per subject" pattern. At `N = 1` it works fine. At `N = 5` it samples in ~2.5 min. At `N = 20` it never finished a 400-draw run in 40+ minutes — because each per-subject scan becomes its own JIT-compile unit and per-subject compilation dominates. The batched single-scan pattern above scales linearly in N for the leaf compute while keeping compilation roughly constant in N.
```
AlexanderFengler, line 243: While I anyways agree with the `one scan` approach, I don't fully follow the logic here as to why `N = 20` dies completely.

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+
+### 3.5 Per-participant batching — one scan, not N
+
+v1 runs **one** batched `pytensor.scan` over `T` and vectorises over participants — *not* `N` separate scans. The feasibility prototype showed the Python-loop-over-subjects pattern doesn't scale (~2.5 min at N=5, didn't finish a 400-draw run at N=20 in 40+ min); details below.
+
+For balanced panels of `N` participants and `T` trials each, the builder reshapes the input `(N*T, ...)` arrays to `(N, T, ...)` internally and runs a single forward recursion whose hidden-state tensor carries a leading participant axis. The recursion proceeds along the *trial* axis, and at every step its scan step processes all N participants in parallel.
```
AlexanderFengler, line 229: This makes sense, but also I think we can solve some of the issues around unbalanced trials simply via data-preprocessing. 

The RLSSM class should actually have something like that already implemented. (basically shape alignment + `no_op` definition)

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+class HSSM_HMM(HSSMBase):
+    def __init__(
+        self,
+        data: pd.DataFrame,
+        decision_process: str | BaseModelConfig | None = None,
+        K: int | None = None,
+        switching_params: list[str] | None = None,
+        *,
+        model_config: HMMConfig | None = None,
+        transition_prior: TransitionPriorSpec | dict | None = None,
```
AlexanderFengler, line 531: here we might have to iterate a bit. 
We have e.g. a way to define priors via `Bambi` utils, at least we probably want to find a way to make the signatures as close as possible so users get as harmonized of an experience across HSSM Classes as possible.

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+        initial_distribution: InitialDistributionSpec | None = None,
+        decision_process_loglik_kind: LoglikKind | None = None,
+        participant_col: str = "participant_id",
+        ordering: OrderingSpec | None = None,
+        # ---- inherited HSSM kwargs ----
+        include: list[...] | None = None,
+        p_outlier: float | bmb.Prior | None = None,   # v1: rejected — decision 10.1.9
```
AlexanderFengler, line 538: In the other doc I left a comment re, whether this should or should not be part of v1

docs/design/hssm_hmm.md
```diff
@@ -0,0 +1,1082 @@
...
+import numpy as np
+import hssm
+from hssm.hmm import (
+    HMMConfig,
+    DirichletConcentration,
+    FixedInitialDistribution,
+    OrderByParam,
+)
+
+cfg = HMMConfig(
```
AlexanderFengler, line 730: this makes sense, just question re naming conventinon on `decision_process` as per my comment on the `_overview.md` doc.

AlexanderFengler (changes requested):
Few high level comments: 

1. Plan should include some tutorials that showcase the new class (sorry if I overlooked that)
2. We might want to do one round of investigation as to what nice class-proprietary visualizations could be here (that could be a separate PR in the end)


Overall this already solid most of my comments are small refinements. I think we can start implementing phase very soon here.  Thanks @fmuia

docs/design/hssm_hmm_overview.md
fmuia: Done.

docs/design/hssm_hmm_overview.md
fmuia: Renamed to `RSSSM` (Regime-Switching SSM, paralleling `RLSSM`). Updated across both design docs. Let me know if you are happy with the name.

docs/design/hssm_hmm_overview.md
fmuia: Tightened the section opener to make the boundary explicit: v1 = first shipped version, everything else is post-v1.

docs/design/hssm_hmm.md
fmuia: Checked RLSSM through Claude scan and couldn't find a no_op preprocessing there. Would you mind pointing me in the right direction?

If it turns out it's not in main yet, three options for v1:

**A. Keep balanced-only.** Matches what RLSSM ships with today; document end-padding + emission-mask as a clean post-v1 extension hook.

**B. Fast-follow PR.** v1 ships balanced-only; the *very next* PR adds unbalanced support (~50 LOC: preprocessing pads each participant to `T_max` at the end with placeholder rows, scan body adds one `pt.switch(mask, log_emission, 0.0)`, plus tests). Scope is written down, not deferred indefinitely.

**C. Inside v1.** Same ~50 LOC + tests land in the v1 architectural PR(s).

Math sketch for B/C: end-padding + zeroed log_emission makes the masked steps "idle" the Markov chain through unobserved trials, so `logsumexp_k log_alpha_{T_max}[n, k] = logsumexp_k log_alpha_{T_n}[n, k]` exactly. Mathematically identical to fitting each participant on their real `T_n` trials.

Happy with any of A/B/C, lean A unless you have a strong reason for B.

docs/design/hssm_hmm.md
fmuia: Replied with detail on the parent comment (line 229) — three options for handling unbalanced panels, asking you to point at the RLSSM `no_op`.

docs/design/hssm_hmm_overview.md
fmuia: If I understand correctly your objection: RLSSM uses `decision_process` specifically to disambiguate from `learning_process`, but RSSSM has only one process (no learning rule), so that disambiguation isn't needed. Aligning with HSSM's `model` is more natural for users.

Renamed `decision_process` -> `model` and `decision_process_loglik_kind` -> `loglik_kind` across both design docs. Closes #16 too.

docs/design/hssm_hmm.md
fmuia: Renamed to `model` per your comment on the overview doc.

docs/design/hssm_hmm.md
fmuia: Reworded throughout the doc: we interrupted the N=20 loop run after 40+ min because it was too slow to be practical, did not wait to confirm the specific failure mode.

docs/design/hssm_hmm.md
fmuia: Do you mean we should drop the post-hoc layer entirely (no FFBS regime recovery, no per-trial logp for `arviz.loo`), or just reimplement it without NumPy?

The post-hoc layer covers two things: (1) regime recovery via FFBS, since the forward algorithm marginalises the latent regimes out at sample-time so they need to be reconstructed after fitting; (2) per-trial logp for `arviz.loo`/`waic`, since v1 contributes a single scalar via `pm.Potential` and bypasses `make_distribution`'s per-trial contract.

We went with NumPy on the post-hoc path because it works uniformly across all three emission backends (analytical, LAN-jax, LAN-pytensor) without a per-backend recompile, and it runs eagerly on saved posterior draws (no graph to JIT). The forward step itself is light: the heavy emission evaluation is reused via the existing `pytensor.function` compile.

In principle we could lift the recursion to pytensor. The win would be vectorising across posterior draws inside one scan; the cost is the JIT-compile-per-backend coupling we deliberately broke. Post-hoc isn't on the sampling hot path so the speedup probably doesn't pay for the coupling. Happy to revisit if profiling says otherwise.

Let me know what you think.

docs/design/hssm_hmm_overview.md
fmuia: Two readings of your comment, want to pick the right one before updating the design.

**Option A (fixed list).** User passes per-regime constants directly, e.g. `p_outlier=[0.02, 0.40]`. The standard HSSM lapse mixture is applied per regime, with each regime k using its own fixed `p_outlier_k`. Natural list-extension of HSSM's `p_outlier=0.05`.

**Option B (inferred prior).** `p_outlier` becomes a switching parameter with a per-regime Beta prior, e.g. `switching_params=["v", "p_outlier"]`. Each regime's lapse rate is inferred from data; K additional RVs to identify.

Both options reuse HSSM's existing `lapse`-distribution machinery (default `Uniform(0, 20)`, precompiled `lapse_func` from [dist.py:498-506](src/hssm/distribution_utils/dist.py#L498-L506)), and both wrap L2's emission with the 4-line mixture formula from [dist.py:587-591](src/hssm/distribution_utils/dist.py#L587-L591). The only real difference is whether per-regime lapse rates are user-specified constants or inferred from data.

Which one did you mean? Your example ("one regime has low, one has high") fits both.

docs/design/hssm_hmm.md
fmuia: Asked you to clarify on the overview comment (line 74). Two API options on the table; deciding doc updates once you weigh in.

docs/design/hssm_hmm_overview.md
AlexanderFengler: Can roll with that name for now. Might have to change later, in acknowledgment of what is a more prominent / recognizable lingo in the field, but have to double check that myself.

docs/design/hssm_hmm_overview.md
AlexanderFengler: yep for now let's do that. 

Later we might wholesale rename to make this homogenous across all classes, but while it's not homogenous, it's logically more coherent here to match `HSSM` class than `RLSSM` class on the argument name.

Global renaming should be it's own PR either way.

docs/design/hssm_hmm_overview.md
AlexanderFengler: Option B is the one I had in mind when writing the comment.

I read option A as a basic extension of setting any parameter actively to a constant. Arguably in the HMM case, every parameter that one wants to set to a constant, should be allowed to be set to a constant per regime?

docs/design/hssm_hmm.md
AlexanderFengler: Since this is light-weight would go with C here.

AlexanderFengler: only wrinkly is if you run into any trouble with the `pt.switch` statement across backends (but I think there is no problem).

docs/design/hssm_hmm.md
AlexanderFengler: shouldn't we be able to run this participant wise as well (basically disconnected participants, independent parameters for each), while using the block structure for fast computations nevertheless?

docs/design/hssm_hmm.md
AlexanderFengler: To be fair, here I might need to go into the details of the existing code to understand better how general the FFBS function is. 

My worry is that we start carrying custom numpy code that won't generalize across many formulations. If that is not a real risk, then willing to forget about it for now.

docs/design/hssm_hmm_overview.md
fmuia: Sounds good. Will revisit naming with you when there's a global rename PR.

docs/design/hssm_hmm_overview.md
fmuia: Agreed, leaving the global rename for its own PR.

docs/design/hssm_hmm.md
fmuia: Ok for C, I added unbalanced panels in v1 via end-padding + emission-mask. We validated this with throwaway scripts across the default, JAX and NUMBA backends, so your pt.switch caveat should be fine. One implementation note: the padding rows still get passed through the emission likelihood before the mask is applied, so they have to hold a valid (rt, response). Otherwise the DDM logp returns NaN/-inf and poisons the gradient even though the mask zeroes those rows. We pad by duplicating each participant's last real trial, which is guaranteed valid.

docs/design/hssm_hmm_overview.md
fmuia: Two things here, now resolved. (1) Balanced-only is dropped: unbalanced panels land in v1 via end-padding + emission-mask. (2) No-pooling is in v1 too: pooling="none" gives independent per-participant params with the block structure kept for speed (P and pi0 global). So the balanced-panel restriction this bullet described is going away. Both validated with throwaway scripts.

docs/design/hssm_hmm.md
fmuia: The generality argument is actually why we went NumPy: it runs eagerly on saved draws and reuses the compiled emission, so one implementation covers analytical, LAN-jax and LAN-pytensor with no per-backend recompile. A pytensor version would be more specialised, not less. We validated the post-hoc path with throwaway scripts; happy to revisit if profiling shows a real cost.

docs/design/hssm_hmm_overview.md
fmuia: Agreed, pulling estimable pi0 into v1 as an opt-in variant (default stays uniform-fixed). We validated it end to end with throwaway scripts. One critical point: pi0 is informed only by first trials, so it is weakly identified at small N and stays global across participants in v1.

docs/design/hssm_hmm_overview.md
fmuia: Confirmed Option B applied. On the broader question, agreed, let's generalize to all params: a scalar stays shared, a length-K list becomes fixed-per-regime, and declaring it in switching_params makes it inferred-per-regime. Three consistent modes that inherit the HSSM constant-vs-prior duality. We validated all three end to end with throwaway scripts, including p_outlier recovered as a switching param with an inferred Beta prior per regime.

docs/design/hssm_hmm.md
fmuia: Resolved on the overview line 74 thread: going with Option B (p_outlier as a switching param, inferred Beta per regime), and the broader generalization to all params is accepted. Validated end to end with throwaway scripts.

docs/design/hssm_hmm.md
fmuia: Agreed, adding no-pooling to v1 with the batched-block design: independent per-participant params, the block structure kept for fast computation, P and pi0 global. We validated this with throwaway scripts (per-participant v recovered, MAE ~0.15, 100% CI coverage; global P recovered; converges). The critical thing to watch is identifiability for participants with few trials, which Phase 2 needs to confirm before merge.

docs/design/hssm_hmm.md
fmuia: Agreed on harmonizing. Users will specify P and pi0 priors the same way as everywhere else in HSSM, the {"name": ...} prior dict or a bmb.Prior (with the usual dict = inferred, value = fixed duality for pi0); the bespoke dataclasses become internal only. Since HSSM has no simplex/matrix-prior convention today, we extend the same dict to Dirichlet for P and pi0. Validated with throwaway scripts: the dict inputs give posteriors identical to the old spec objects.

fmuia:
Re your two high-level review points:

1. **Tutorials.** Phase 5 now enumerates two notebooks in both design docs: a regression-test mirror of `hmm_ddm_regime_switching.ipynb` rewritten to use `RSSSM`, and a multi-participant K=3 showcase exercising the batched scan and FFBS regime recovery.
2. **Class-proprietary visualizations.** Yes, definitely. There are some nice ones we can build here (regime timelines, transition-matrix heatmaps, FFBS marginal-probability plots), also borrowing ideas from finance switching-regime models. We'll scope these as a separate follow-up PR.

AlexanderFengler (approved):
Thanks @fmuia I read through this once more now and I think we are good to start the implementation.

PR #971, opened 2026-06-01
Cpaniaguam/model listing at module level

cpaniaguam:
This PR adds module-level model discovery APIs: [hssm.list_models()](vscode-file://vscode-app/c:/Users/cpaniagu/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for built-in HSSM models and [hssm.rl.list_models()](vscode-file://vscode-app/c:/Users/cpaniagu/AppData/Local/Programs/Microsoft%20VS%20Code/f6cfa2ea24/resources/app/out/vs/code/electron-browser/workbench/workbench.html) for RLSSM models. The existing class-level accessors are kept as backward-compatible delegates, and tests/docs were updated accordingly.

https://github.com/lnccbrown/HSSM/pull/955#discussion_r3312829307

### DIFF
--- a/src/hssm/modelconfig/__init__.py
+++ b/src/hssm/modelconfig/__init__.py
@@ -35,6 +35,17 @@ from typing import get_args
 from .._types import DefaultConfig, SupportedModels
 
 
+def list_models() -> tuple[SupportedModels, ...]:
+    """Return the names of the built-in HSSM models.
+
+    Returns
+    -------
+    tuple[SupportedModels, ...]
+        A tuple containing all built-in HSSM model names.
+    """
+    return get_args(SupportedModels)
+
+
 def get_default_model_config(model_name: SupportedModels) -> DefaultConfig:
     """
     Get the default configuration for a given model name.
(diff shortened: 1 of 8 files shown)


src/hssm/modelconfig/__init__.py
```diff
@@ -35,6 +35,17 @@
 from .._types import DefaultConfig, SupportedModels
 
 
+def list_models() -> tuple[SupportedModels, ...]:
```
digicosmos86, line 38: So this comes from `SupportedModels` while the `list_models` for RL calls the model registry. Should this come from model registry as well? If the user registers a new model, `list_model` would not include the new model right?

digicosmos86 (approved):
LGTM! I would also add a deprecation warning to `supported_models` in the HSSM class and then remove it completely after a while

src/hssm/modelconfig/__init__.py
cpaniaguam: That's another good point! I think this should be added in a separate PR.

cpaniaguam: https://github.com/lnccbrown/HSSM/issues/972

PR #973, opened 2026-06-03
Initial migration to pymc6

digicosmos86:
This is the first of many PRs required to get HSSM to be compatible with PyMC6. All future PRs need to be merged to `migration-pymc6` branch

Changes in this PR:
1. updated dependencies in `pyproject.toml` (still many things need to be decided, such as whether to drop `numpyro` in favor of `nutpie`
2. Updated a few utilities that post-process `InferenceData` object. Now that the `InferenceData` is replace with a `xr.DataTree` object, most functions that deals with the `InferenceData` object need to be updated
3. I dug a little bit into the main tutorial. Many arviz functions need to be updated


cpaniaguam (changes requested):
Consider marking the yet-to-address broken tests as failing.

digicosmos86:
> Consider marking the yet-to-address broken tests as failing.

Through what mechanism? Commenting in code?

cpaniaguam:
@digicosmos86 
> Through what mechanism? Commenting in code?

You can use [pytest.xfail](https://docs.pytest.org/en/stable/reference/reference.html#pytest-xfail).

cpaniaguam:
@digicosmos86 
> Through what mechanism? Commenting in code?

See https://docs.pytest.org/en/stable/how-to/skipping.html#xfail

digicosmos86:
> @digicosmos86
> 
> > Through what mechanism? Commenting in code?
> 
> See https://docs.pytest.org/en/stable/how-to/skipping.html#xfail

I think that's fair only when there's not a whole lot of tests breaking. Otherwise adding the marks itself could be labor intensive. Why don''t we evaluate after merging your recent PRs? Looks like a few broken tests were fixed in them

cpaniaguam:
> Why don''t we evaluate after merging your recent PRs? Looks like a few broken tests were fixed in them

Those were just two of the fast tests. There are many other slow ones that are failing (see https://github.com/lnccbrown/HSSM/pull/982).

> I think that's fair only when there's not a whole lot of tests breaking.

I understand your point. I still think there’s value in marking the failures we already understand, since it keeps CI easier to read and helps distinguish known issues from new regressions. Once a fix is introduced, those tests will show up as `xpassed` in the summary, so we’ll have a clear signal to remove the markers.

src/hssm/base.py
```diff
@@ -795,12 +795,7 @@ def _clean_posterior_group(self, idata: az.InferenceData | None = None):
             for var_ in vars_to_keep
             if isinstance(var_, str) and "_mean" not in var_
         ]
-
-        setattr(
-            idata,
-            "posterior",
-            idata["posterior"][vars_to_keep_clean],
-        )
+        dt["posterior"] = dt["posterior"].to_dataset()[vars_to_keep_clean]
```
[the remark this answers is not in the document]
digicosmos86: `dt` is no longer an `InferenceData` object

src/hssm/utils.py
```diff
@@ -164,34 +165,34 @@ def _compute_log_likelihood(
...
+    # # Populate the posterior in the DataTree object
     # with the likelihood parameters
-    idata = model._compute_likelihood_params(  # pylint: disable=protected-access
-        idata, data, include_group_specific, sample_new_groups
+    dt = model._compute_likelihood_params(  # pylint: disable=protected-access
+        dt, data, include_group_specific, sample_new_groups
     )
 
-    required_kwargs = {"model": model, "posterior": idata["posterior"], "data": data}
+    required_kwargs = {"model": model, "posterior": dt["posterior"], "data": data}
```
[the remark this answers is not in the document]
digicosmos86: not fixing right now

digicosmos86:
@AlexanderFengler @krishnbera can you give this a quick review? This has become a blocker

src/hssm/base.py
```diff
@@ -515,7 +515,7 @@ def sample(
         initvals: str | dict | None = None,
         include_response_params: bool = False,
         **kwargs,
-    ) -> az.InferenceData | pm.Approximation:
+    ) -> xr.DataTree | pm.Approximation:
```
krishnbera, line 518: I think this PR should migrate `sample_posterior_predictive()` to the DataTree.

krishnbera: `save_model()` will break as well?

pyproject.toml
```diff
@@ -14,30 +14,23 @@ readme = "README.md"
 license = { file = "LICENSE" }
 keywords = ["HSSM", "sequential sampling models", "bayesian", "bayes", "mcmc"]
 
-requires-python = ">=3.11,<3.14"
+requires-python = ">=3.12,<3.14"
```
krishnbera, line 17: PR checks are still running and failing on Python 3.11

krishnbera (dismissed):
Thanks for pushing this forward. The basic sampling path  HSSM.sample() now returns an xarray.DataTree. I think we should use this PR to migrate downstream APIs from the old InferenceData? `sample_posterior_predictive()` and `save_model()` will fail currently.

krishnbera (approved):

AlexanderFengler (approved):
lgtm, looks like just small start to the overall refactor.
Thanks @digicosmos86

src/hssm/base.py
digicosmos86: Yep. Will fix in a future PR

pyproject.toml
digicosmos86: Removed Python 3.11 from the matrix

PR #974, opened 2026-06-03
Cpaniaguam/fix predictive idata to dataframe datatree

cpaniaguam:
Add handling of inference data containers (`az.InferenceData` and `xr.DataTree`) in the codebase, increasing compatibility with different versions of ArviZ and supporting both data structures throughout the workflow. The changes include new utility methods for accessing and modifying inference data groups, refactoring of predictive sampling functions to use these utilities, and updates to utility functions and tests for consistency.

**Inference Data Handling Improvements:**

* Added utility methods `_get_idata_groups`, `_get_idata_group_dataset`, and `_set_idata_group_dataset` to abstract group access and modification for both `az.InferenceData` and `xr.DataTree`, ensuring compatibility across ArviZ versions.
* Refactored `_drop_parent_str_from_idata` to use the new group access utilities, and updated its type hints to support both `az.InferenceData` and `xr.DataTree`.

**Predictive Sampling Refactor:**

* Updated `sample_prior_predictive` and `sample_do` methods to use the new group access utilities when renaming dimensions and variables, improving code clarity and compatibility. [[1]](diffhunk://#diff-eaff11a3d060b73385e258bc362b6df7683fad454c6017f8b9e35ed224a700afL1196-R1199) [[2]](diffhunk://#diff-eaff11a3d060b73385e258bc362b6df7683fad454c6017f8b9e35ed224a700afL1122-R1131)
* Modified `sample_posterior_predictive` to use the new `_get_idata_groups` utility for checking group existence.

**Utility Function Updates:**

* Updated `predictive_idata_to_dataframe` to accept both `az.InferenceData` and `xr.DataTree`, and handle group data extraction accordingly. [[1]](diffhunk://#diff-11c83362802db3028664e029ce20f83d181667e3631fe618614ad9672665d383L541-R541) [[2]](diffhunk://#diff-11c83362802db3028664e029ce20f83d181667e3631fe618614ad9672665d383L552-R553) [[3]](diffhunk://#diff-11c83362802db3028664e029ce20f83d181667e3631fe618614ad9672665d383L562-R566)

**Testing:**

* Marked the `test_make_rl_logp_op` test as expected to fail (`xfail`) due to upstream changes in the PyTensor API. See #975 with a fix.


src/hssm/base.py
```diff
@@ -1924,32 +1912,76 @@ def _get_deterministic_var_names(self, idata) -> list[str]:
...
+        """Replace a named inference-data group with a dataset."""
+        if isinstance(idata, xr.DataTree):
+            idata[group] = dataset
+        else:
+            setattr(idata, group, dataset)
+
+    def _drop_parent_str_from_idata(
+        self, idata: az.InferenceData | xr.DataTree | None
+    ) -> az.InferenceData | xr.DataTree:
+        """Drop the parent_str variable from an inference data object.
```
cpaniaguam:
```suggestion
        """Drop the parent_str variable from an (inference) data object.
```

src/hssm/base.py
```diff
@@ -1924,32 +1912,79 @@ def _get_deterministic_var_names(self, idata) -> list[str]:
 
         return var_names
 
-    def _drop_parent_str_from_idata(
-        self, idata: az.InferenceData | None
-    ) -> az.InferenceData:
-        """Drop the parent_str variable from an InferenceData object.
+    def _get_idata_groups(
```
digicosmos86, line 1915: I understand that you are trying to be backward-compatible here. However, there is no need. `Arviz` is no longer backward-compatible, so we are moving forward from `InferenceData` altogether. `az.InferenceData` doesn't really exist any more. I am surprised that we didn't get an error. Let's drop this layer of backward compatibility. This new version of HSSM should not support `InferenceData`

digicosmos86:
@cpaniaguam no need for this to be closed. We still need to make these functions compatible with DataTree. Maybe you'll do this in another PR?

I actually don't really like these functions because they break wrappers and tend to break when breaking changes are introduced. Let's also think of ways to remove these functions. We might lose some functionalities but gain maintainability

PR #975, opened 2026-06-03
Cpaniaguam/fix rl likelihood builder pymc6

### LINKED ISSUE #976: [pymc6 migration] Fix likelihood-related issues with RL due to change of internals

cpaniaguam:
Prevent PyTensor from incorrectly optimizing JAX-backed operations.

**JAX-PyTensor Integration Improvements:**

* Added `do_constant_folding` methods to both `LANLogpOp` and `LANLogpVJPOp` classes in `src/hssm/distribution_utils/jax.py` to prevent PyTensor from attempting to precompute (constant fold) outputs of JAX-backed operations, ensuring correct runtime behavior. [[1]](diffhunk://#diff-5d2a013aa4689c4acaad627f065537a45ad2d88ad60d9a26fd2fc8dbf3739190R49-R52) [[2]](diffhunk://#diff-5d2a013aa4689c4acaad627f065537a45ad2d88ad60d9a26fd2fc8dbf3739190R141-R144)

**Test Suite Updates:**

* Updated `test_make_rl_logp_op` in `tests/rl/test_rl_likelihood_builder.py` to use `pytensor.function` with `mode="FAST_COMPILE"` for gradient evaluation, improving test reliability and compatibility.
* Marked `test_predictive_idata_to_dataframe` in `tests/test_utils.py` with `@pytest.mark.xfail` to indicate it is expected to fail due to recent changes in PyMC, preventing it from causing false negatives in CI. See #974 with a fix.


pyproject.toml
```diff
@@ -26,6 +26,8 @@ dependencies = [
     "absl-py>=2.3.1",
     "bambi>=0.18.0",
     "cloudpickle>=3.0.0",
+    "h5netcdf>=1.6.3",
+    "h5py>=3.14.0",
```
digicosmos86: Why do we need these packages? Are these absolutely necessary to maintain basic functionalities?

src/hssm/distribution_utils/jax.py
```diff
@@ -46,6 +46,10 @@ def make_jax_logp_ops(
     class LANLogpOp(Op):  # pylint: disable=W0223
         """Wraps a JAX function in an pytensor Op."""
 
+        def do_constant_folding(self, fgraph, node):
+            """Keep PyTensor from trying to precompute opaque JAX-backed outputs."""
+            return False
```
digicosmos86, line 51: Wow... this pattern from pytensor is really bad...

pyproject.toml
cpaniaguam: I think so. HSSM still restores .nc traces through `az.from_netcdf(...)`
https://github.com/lnccbrown/HSSM/blob/95b27df961c2bf08c0fd09cce9481df1e2ba8aa8/src/hssm/base.py#L1490
https://github.com/lnccbrown/HSSM/blob/95b27df961c2bf08c0fd09cce9481df1e2ba8aa8/src/hssm/base.py#L1508
https://github.com/lnccbrown/HSSM/blob/95b27df961c2bf08c0fd09cce9481df1e2ba8aa8/src/hssm/base.py#L1665

What seems to have changed is that newer xarray/ArviZ installs no longer seem to guarantee a working NetCDF backend transitively, and CI started failing when pytest tried to load the .nc fixture in

https://github.com/lnccbrown/HSSM/blob/95b27df961c2bf08c0fd09cce9481df1e2ba8aa8/tests/conftest.py#L91

digicosmos86: Yeah it seems that both `xarray` and `arviz` now have them as optional and/or dev dependencies. I think these packages do that for a good reason, and we probably should do the same and then add these packages to dev dependencies. In code, we can probably check if these packages are installed and throw an informative error if not.

I like to keep the dependencies as slim as possible and not add additional packages for things that are occasionally used. @cpaniaguam @AlexanderFengler @krishnbera thoughts?

@cpaniaguam these .nc files might need to be generated again since they came from old `InferenceData` saves

digicosmos86: Additionally, I think it could be a good idea to have an `io` optional group and have `cloudpickle`, `h5py` and `h5netcdf` in that group

digicosmos86:
@cpaniaguam the upstream errors should all be fixed now. Can you change the base to `996-fix-model-cartoon-plot` and see if all tests pass now?

cpaniaguam:
> @cpaniaguam the upstream errors should all be fixed now. Can you change the base to `996-fix-model-cartoon-plot` and see if all tests pass now?

Sure!

cpaniaguam:
At least one test is failing https://github.com/lnccbrown/HSSM/actions/runs/27966221081/job/82760632068
@digicosmos86

cpaniaguam:
@digicosmos86 There still is one function that needs to be updated to handle `xr.DataTree` in `src/hssm/utils.py` -- `predictive_idata_to_dataframe`. The corresponding test is in `tests/test_utils.py::test_predictive_idata_to_dataframe`. I kept the failing marking on it.

digicosmos86:
> @digicosmos86 There still is one function that needs to be updated to handle `xr.DataTree` in `src/hssm/utils.py` -- `predictive_idata_to_dataframe`. The corresponding test is in `tests/test_utils.py::test_predictive_idata_to_dataframe`. I kept the failing marking on it.

This is fixed here https://github.com/lnccbrown/HSSM/pull/993/changes#diff-11c83362802db3028664e029ce20f83d181667e3631fe618614ad9672665d383R558

pyproject.toml
AlexanderFengler: I personally think that easiest user experience should win here over being minimal on dependencies. Instead of `h5py` we could e.g. use `zarr` if that is preferable, but I wouldn't create any hurdles for users specifically with saving and loading models. 

I get the idea of trying to stay minimal, but we have balanced objectives.
Anything we should do should start from the premise of: Without any extra hassle, a user needs to be able to load and save models they are working with.

Arviz and xarray cater to, on average, more advanced users than we do, so they face different trade-offs.

digicosmos86: Yeah I changed my mind too. I think arviz removed those packages because they removed `InferenceData` from the package, so it mainly stays as a viz-focused package. For HSSM users, it might be frustrating to find out that you need additional packages installed after a long sampling session. I'll revert these changes in a new PR

digicosmos86 (approved):
LGTM!

PR #993, opened 2026-06-17
Reapply [pymc6 migration] fixed `sample_posterior_predictive`, `sample_prior_predictive`, and `sample_do`

### LINKED ISSUE #990: [pymc6 migration] Fix sample_posterior_predictive
This method directly modifies the traces object which is a `xr.DataTree` now. Need to fix compatibility issues

digicosmos86:
#991 was merged by accident. This reapplies the changes in that PR. Please see the PR for notes

## Bug Fixes

* Fixed posterior predictive sampling functionality
* Fixed prior predictive sampling
* Fixed sampling with NumPyro sampler using MAP initialization
* Improved handling of inference data management

### DIFF
--- a/src/hssm/base.py
+++ b/src/hssm/base.py
@@ -966,9 +966,8 @@ class HSSMBase(ABC, DataValidatorMixin, MissingDataMixin):
             _logger.info(
                 "dt=None, we use the traces assigned to the HSSM object as datatree."
             )
-
-        if dt is not None:
-            if "posterior_predictive" in dt.groups():
+        else:
+            if inplace and "posterior_predictive" in dt:
                 del dt["posterior_predictive"]
                 _logger.warning(
                     "pre-existing posterior_predictive group deleted from datatree."
@@ -1002,7 +1001,7 @@ class HSSMBase(ABC, DataValidatorMixin, MissingDataMixin):
             and not np.allclose(draws, dt["posterior"].draw.values)
         ):
             # Reassign posterior to sub-sampled version
-            setattr(dt_copy, "posterior", dt["posterior"].isel(draw=draws))
+            dt_copy["posterior"] = dt["posterior"].isel(draw=draws)
 
         if kind == "response":
             # If we run kind == 'response' we actually run the observation RV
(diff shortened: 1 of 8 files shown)


src/hssm/base.py
```diff
@@ -1178,38 +1166,33 @@ def sample_prior_predictive(
...
         del prior_predictive["posterior"]
 
         if self._inference_obj is None:
             self._inference_obj = prior_predictive
         else:
-            self._inference_obj.extend(prior_predictive)
+            for group_name in list(prior_predictive.groups):
+                if group_name == "/":
+                    continue
+                self._inference_obj[group_name] = prior_predictive[group_name]
```
AlexanderFengler, line 1183: don't fully get the logic of this one.

src/hssm/base.py
```diff
@@ -1876,15 +1859,13 @@ def _drop_parent_str_from_datatree(self, dt: DataTree | None) -> DataTree:
         if dt is None:
             raise ValueError("Please provide a DataTree (traces) object.")
         else:
-            for group in dt.groups():
+            for group in dt.groups:
```
AlexanderFengler, line 1862: same here

src/hssm/utils.py
```diff
@@ -535,19 +534,19 @@ def check_data_for_rl(
 
 
 def predictive_idata_to_dataframe(
```
AlexanderFengler: consider function rename to `dt`

tests/test_sample_posterior_predictive.py
```diff
@@ -53,7 +50,8 @@ def test_sample_posterior_predictive(
             },
         ],
     )  # Doesn't matter what model or data we use here
-    delattr(cav_idata, "posterior_predictive")
+    if "posterior_predictive" in cav_idata:
+        del cav_idata["posterior_predictive"]
```
AlexanderFengler: might have to go for `cav_dt` as well then.

AlexanderFengler (commented):
minor points.

src/hssm/base.py
digicosmos86: Previously, `az.InferenceData.extend()` adds the fields in another inferenceData object to the original inference data object. With DataTree, it's not possible, so this code iterates through all groups in `prior_predictive` and assign the group to `_inference_obj`. In `DataTree`, there is a root group `/` which do not need to be re-assigned

PR #1008, opened 2026-06-23
[pymc6 migration] Updated tutorials to the best of LLM's ability

digicosmos86:


AlexanderFengler (approved):
lgtm. It's just a reasonable starting point. Thanks @digicosmos86

PR #1028, opened 2026-07-02
[pymc6 migration] add python 314 support

digicosmos86:
Updated `ssm-simulators` dependency to support Python 3.14. Also updated CI workflow to add tests for Python 3.14

### DIFF
--- a/tests/test_hssm.py
+++ b/tests/test_hssm.py
@@ -1,3 +1,5 @@
+import sys
+
 import bambi as bmb
 import numpy as np
 import pytest
@@ -148,6 +150,11 @@ def test_model_definition_outside_include(data_ddm):
         HSSM(data_ddm, include=[{"name": "a", "prior": 0.5}], a=0.5)
 
 
+@pytest.mark.xfail(
+    sys.version_info >= (3, 14),
+    reason="sample_posterior_predictive fails on 3.14 with cpickle issue",
+    strict=True,  # This will let us know in the future when this is fixed
+)
 @pytest.mark.slow
 def test_sample_prior_predictive(data_ddm_reg):
     data_ddm_reg = data_ddm_reg.iloc[:10, :]
@@ -416,6 +423,11 @@ class TestFixedVectorParams:
         assert "z" in idata.posterior.data_vars
 
 
+@pytest.mark.xfail(
+    sys.version_info >= (3, 14),
+    reason="sample_posterior_predictive fails on 3.14 with cpickle issue",
+    strict=True,  # This will let us know in the future when this is fixed
+)
 @pytest.mark.slow
 def test_sample_do(data_ddm):
     model = HSSM(data=data_ddm)
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -1,12 +1,13 @@
 """Test plotting module."""
 
-import pytest
+import sys
 
+import arviz as az
 import matplotlib.pyplot as plt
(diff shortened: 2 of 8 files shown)


tests/test_hssm.py
```diff
@@ -148,6 +150,11 @@ def test_model_definition_outside_include(data_ddm):
         HSSM(data_ddm, include=[{"name": "a", "prior": 0.5}], a=0.5)
 
 
+@pytest.mark.xfail(
+    sys.version_info >= (3, 14),
+    reason="sample_posterior_predictive fails on 3.14 with cpickle issue",
+    strict=True,  # This will let us know in the future when this is fixed
+)
```
cpaniaguam, line 157:
```suggestion
@pytest.mark.skipif(
    sys.version_info >= (3, 14),
    reason="sample_posterior_predictive fails on Python 3.14 due to cpickle issue",
)
```

tests/test_hssm.py
```diff
@@ -416,6 +423,11 @@ def test_fixed_vector_multiple_params(self, data_ddm):
         assert "z" in idata.posterior.data_vars
 
 
+@pytest.mark.xfail(
+    sys.version_info >= (3, 14),
+    reason="sample_posterior_predictive fails on 3.14 with cpickle issue",
+    strict=True,  # This will let us know in the future when this is fixed
+)
```
cpaniaguam, line 430:
```suggestion
@pytest.mark.skipif(
    sys.version_info >= (3, 14),
    reason="sample_posterior_predictive fails on Python 3.14 due to cpickle issue",
)
```

tests/test_plotting.py
```diff
@@ -165,6 +166,11 @@ def test__plot_predictive_2D(self, cav_dt, cavanagh_test):
         assert len(g2.figure.axes) == 5 * 2
         assert len(g2.figure.axes[0].get_lines()) == 1
 
+    @pytest.mark.xfail(
+        sys.version_info >= (3, 14),
+        reason="sample_posterior_predictive fails on 3.14 with cpickle issue",
+        strict=True,  # This will let us know in the future when this is fixed
+    )
```
cpaniaguam, line 173:
```suggestion
    @pytest.mark.skipif(
        sys.version_info >= (3, 14),
        reason="sample_posterior_predictive fails on 3.14 with cpickle issue",
    )
```

tests/test_plotting.py
```diff
@@ -322,6 +328,11 @@ def test__plot_quantile_probability_2D(
         )
         assert len(g.figure.axes) == 5 * 4
 
+    @pytest.mark.xfail(
+        sys.version_info >= (3, 14),
```
cpaniaguam, line 332: Consider defining a constant and use throughout.

cpaniaguam: And maybe use `skipif`!

cpaniaguam (changes requested):
Thanks for this! I think it might be better to use `skipif` as we know the test passes for versions <=3.13.

digicosmos86:
> Thanks for this! I think it might be better to use `skipif` as we know the test passes for versions <=3.13.

`skipif` was my first thought. The reason why I used `xfail` is that in strict mode (`strict=True`), when this issue is fixed (we wouldn't know when) and these tests `xpass`, pytest will fail and we will know the this is fixed in cloudpickle. `skipif` does not do this

cpaniaguam (approved):

PR #1029, opened 2026-07-02
Docker image, GHCR publish, and Codespaces devcontainer (dockerHSSM)

AlexanderFengler:
Adds a runnable HSSM Docker image, a workflow that publishes it to **GHCR** (`ghcr.io/lnccbrown/hssm`) on release, and a **Codespaces / dev container** so people can contribute to HSSM in the browser.

## Credit / origin
This revives **#439 (dockerHSSM)** by @panwanke and @hcp4715 — thank you. That PR (May 2024) is now stale against `main` (deprecated `jupyter/scipy-notebook` base, China-mirror pins, `pymc==5.14`, and a version read from the long-gone `tool.poetry.version`). Rather than rebase its ~4k lines of regenerated-notebook churn, this reworks the ~90 lines that matter for today's HSSM. Original authorship is preserved via `Co-authored-by` trailers.

## What's here
- **`Dockerfile`** — `python:3.13-slim` + `graphviz` + `uv`. Installs HSSM **from the checked-out source** (image matches the built ref exactly; no PyPI release-timing race) plus a minimal JupyterLab UI. Non-root user; tutorials baked in. `docker run --rm -p 8888:8888 ghcr.io/lnccbrown/hssm:latest` opens a notebook.
- **`.github/workflows/docker-image.yml`** — builds on relevant PRs (**validation only, no push**) and publishes on `release`/`workflow_dispatch`. Version read via stdlib `tomllib` from `[project].version`; auth via `GITHUB_TOKEN` (no secrets needed for GHCR).
- **`.devcontainer/devcontainer.json`** — Codespaces / VS Code dev container. Builds from the same `Dockerfile` (so run-image and dev-image share one base), then `uv sync` on create layers an editable dev env (project + `dev` group: pytest, ruff, mypy, pre-commit). `onCreateCommand` is baked into Codespaces prebuilds for fast startup.
- **`.dockerignore`** — lean build context.

## Deliberate scope
- **amd64-only for now.** `ssm-simulators` and `hddm-wfpt` publish no `linux/arm64` wheels, so a native arm64 image would have to compile them from source (GSL/OpenMP). Apple Silicon still runs this image under emulation, and Codespaces is amd64. Native multi-arch is a one-line `platforms:` change once those packages ship aarch64 Linux wheels — **ssm-simulators side is up in lnccbrown/ssm-simulators#293**; `hddm-wfpt` to follow.
- **Core + notebook UI only** in the image — heavy optional stacks (bayesflow, sbi, lanfactory, keras) are excluded to keep it slim; the dev container adds the `dev` group, and `uv sync --group notebook/docs` pulls the rest on demand.

## Follow-ups (not in this PR)
1. `hddm-wfpt` aarch64 wheels (mirror of ssm-simulators#293) → then flip the image `platforms:` to `linux/amd64,linux/arm64`.
2. Enable Codespaces prebuilds on the repo (Settings → Codespaces → Prebuilds) once merged.

Draft until CI image build passes and the direction is confirmed.

## Summary by CodeRabbit

* **New Features**
  * Added a dev container setup for a consistent local development environment.
  * Added CI automation to build and smoke-test the dev container.
  * Added Docker image build and publish workflows with multi-architecture support and versioned tags.
  * Added a runnable container image with JupyterLab and included tutorial content.
* **Chores**
  * Reduced Docker build context noise to speed up container builds.
  * Updated the container’s JupyterLab launch to use standard token-based access behavior.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,66 @@
+name: Docker image (GHCR)
+
+# Build the HSSM image on relevant PRs (validation only, no push) and publish it
+# to GHCR on release or manual dispatch.
+on:
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - Dockerfile
+      - .dockerignore
+      - .github/workflows/docker-image.yml
+      - pyproject.toml
+      - src/**
+  release:
+    types: [published]
+
+env:
+  IMAGE: ghcr.io/lnccbrown/hssm
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Read version from pyproject.toml
+        id: version
+        run: |
+          V=$(python3 -c "import tomllib; print(tomllib.load(open('pyproject.toml','rb'))['project']['version'])")
+          echo "value=$V" >> "$GITHUB_OUTPUT"
+
+      # Push on release/dispatch; PRs only build to validate the Dockerfile.
+      - name: Decide whether to push
+        id: push
+        run: |
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            echo "value=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "value=true" >> "$GITHUB_OUTPUT"
+          fi
+
+      - uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GHCR
+        if: steps.push.outputs.value == 'true'
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: docker/build-push-action@v6
+        with:
+          context: .
+          # ponytail: amd64 only until ssm-simulators + hddm-wfpt publish
+          # linux/arm64 wheels; then add ,linux/arm64 here for native multi-arch.
+          platforms: linux/amd64
+          push: ${{ steps.push.outputs.value }}
+          tags: |
+            ${{ env.IMAGE }}:${{ steps.version.outputs.value }}
+            ${{ env.IMAGE }}:latest
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,9 +1,15 @@
 name: Docker image (GHCR)
 
-# Build the HSSM image on relevant PRs (validation only, no push) and publish it
-# to GHCR on release or manual dispatch.
+# PRs: amd64 build only (fast sanity). Release: multi-arch build + push to GHCR.
+# Manual dispatch: multi-arch build, pushed only if `publish` is set (otherwise a
+# build-only validation of the arm64 leg).
 on:
   workflow_dispatch:
+    inputs:
+      publish:
+        description: "Push the image to GHCR (otherwise multi-arch build-only)"
+        type: boolean
+        default: false
   pull_request:
     paths:
       - Dockerfile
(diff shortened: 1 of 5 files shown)


AlexanderFengler:
Thanks @coderabbitai — went through the review, addressed in abd5045d:

- **Pin `uv`** → `ghcr.io/astral-sh/uv:0.11.28` for reproducible builds.
- **`.dockerignore`** → drop non-tutorial `docs/` from the build context (`docs/*` + `!docs/tutorials`); verified `COPY docs/tutorials` still resolves in a fresh (uncached) dev-container build.
- **`persist-credentials: false`** on the dev-container checkout (the workspace is mounted into the container).
- **PR trigger paths** → added `README.md`, `LICENSE`, `docs/tutorials/**` so changes to assets the Dockerfile copies are validated.
- **Tokenless JupyterLab** → now secure by default (dropped the empty-token override; Jupyter prints a tokenized URL). Tokenless is a documented opt-in for trusted local networks.

One I intentionally did **not** change:

- **`uv.lock` in the image** — kept as latest-compatible resolution. This is the runnable *demo* image; the reproducible/locked path is the **dev container** (`uv sync`). Building the demo image around the lockfile would mean restructuring it into a venv for little benefit. Documented the choice explicitly in the Dockerfile.

.github/workflows/docker-image.yml
```diff
@@ -0,0 +1,78 @@
...
+  IMAGE: ghcr.io/lnccbrown/hssm
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v4
```
digicosmos86: This is v7 now

Dockerfile
```diff
@@ -0,0 +1,50 @@
+# Runnable HSSM image: HSSM + JupyterLab, ready to explore the tutorials.
+#   docker run --rm -p 8888:8888 ghcr.io/lnccbrown/hssm:latest
+# Built and published from .github/workflows/docker-image.yml.
+#
+# Multi-arch (linux/amd64,linux/arm64) is driven by the workflow's `platforms`.
+# arm64 works because ssm-simulators (>=0.12.5) and hddm-wfpt (>=0.1.7) now ship
+# manylinux aarch64 wheels, so nothing compiles from source on either arch.
+
+FROM python:3.13-slim
```
digicosmos86: Maybe we can start with a uv base image?

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+LABEL org.opencontainers.image.description="Bayesian inference for hierarchical sequential sampling models (HSSM) with JupyterLab."
+LABEL org.opencontainers.image.licenses="MIT"
+
+# graphviz: the `dot` binary the python `graphviz` package shells out to for model graphs.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends graphviz \
+    && rm -rf /var/lib/apt/lists/*
+
+# uv for fast installs (ecosystem standard). Pinned for reproducible builds.
+COPY --from=ghcr.io/astral-sh/uv:0.11.28 /uv /usr/local/bin/uv
```
digicosmos86: This can directly be used as a base image above

Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+# Install HSSM from this checkout (not PyPI): the image matches the built ref
+# exactly and there is no release-timing race. Dependencies still arrive as wheels.
+# Deliberately resolves latest-compatible deps (not uv.lock) — this is the runnable
+# demo image; the reproducible/locked path is the dev container (`uv sync`).
+# ponytail: core + a minimal notebook UI only. The heavy optional stacks
+# (bayesflow, sbi, lanfactory, keras) are NOT included — add them at run time
+# (`uv pip install ...`) or in a follow-up "full" image if a tutorial needs them.
+COPY pyproject.toml README.md LICENSE ./
+COPY src ./src
+RUN uv pip install --system --no-cache . jupyterlab ipywidgets graphviz
```
digicosmos86: I don't think this will install all the deps in the notebook group, so some notebooks might not run properly

digicosmos86 (changes requested):
Generally looks good! A few minor comments, but the missing `notebook` deps might cause some notebooks to fail, but installing the group will make the image really huge with PyTorch, etc being installed. We might need some trade-offs

PR #1034, opened 2026-07-06
[pymc6 migration] Optimize dev workflows

digicosmos86:
This PR replaces `pre-commit` with `prek` for managing pre-commit hooks. `prek` is a modern `pre-commit` replacement written in rust that works better.

@AlexanderFengler @krishnbera @cpaniaguam if you are already using `pre-commit`, please run `uv run prek install -f` to re-install the pre-commit workflow after running `uv sync`. The rest should just work.

I also specified ruff to check notebooks as well, so that our tutorials are also formatted


cpaniaguam (commented):
This PR has commits touching notebooks. Is this intentional?

digicosmos86:
> This PR has commits touching notebooks. Is this intentional?

Yep. All the notebooks are now linted and reformatted with `ruff`

PR #1044, opened 2026-07-09
[pymc6 migration] Restore no-jitter behavior for JAX NUTS samplers (#999)

### LINKED ISSUE #999: [pymc6 migration] restore jitter behavior
According to [this post](https://discourse.pymc.io/t/how-to-pass-a-jitter-to-pm-sample-in-pymc6/17848), `jitter`  option seems to be gone. Jitters now need to be added to `initvals`, if we want to restore the same behavior

AlexanderFengler:
Closes #999.

## Problem
The pymc6 migration silently changed jitter behavior for the JAX samplers. `pymc.sampling.jax.sample_jax_nuts(jitter: bool = True)` still has the switch, but `pm.sample(nuts_sampler="numpyro"|"blackjax")` (which HSSM reaches via bambi) **never forwards a way to set it** — `_sample_external_nuts` calls `sample_jax_nuts(...)` without `jitter=`, so it is always `True`. Passing it via `nuts=`/`nuts_sampler_kwargs` routes it into the NumPyro/BlackJAX *kernel* → `TypeError: NUTS.__init__() got an unexpected keyword argument 'jitter'`.

HSSM historically disabled numpyro's jitter (it applies its own controlled `initval_jitter`, default 0.01); that code had been commented out with a `TODO: add back jitter after knowing what to do`.

## Fix
`pm.sample` resolves the sampler as a **call-time module attribute** (`import pymc.sampling.jax as pymc_jax; pymc_jax.sample_jax_nuts(...)`), so a context manager that temporarily patches that attribute to inject `jitter=False`, wrapped around HSSM's existing `self.model.fit(...)` call, is picked up by `pm.sample`. This keeps **all** of bambi's plumbing intact (only the one hardcoded `jitter` is overridden).

- `utils.py`: `_force_jax_nuts_no_jitter(active=...)` context manager. Guards a future pymc rename/removal (→ no-op); `active=False` is a no-op for the non-JAX samplers.
- `base.py`: wrap the `fit()` call with `active=sampler in ("numpyro", "blackjax")`; remove the dead commented block; one docstring sentence.
- `tests/test_jitter.py`: 3 fast unit tests (patch/inject/restore + missing-attr guard) and 1 fast full-stack test asserting `jitter=False` reaches `sample_jax_nuts` through the real HSSM→bambi→pm.sample dispatch (capture-and-raise sentinel, no MCMC). All 4 pass in <1s.

## Result
For `sampler in {"numpyro","blackjax"}`, chains again start from HSSM's controlled `initval_jitter` initvals (matching pre-migration `main` and the pymc sampler, which HSSM already runs with `init="adapt_diag"`), instead of the sampler's extra uniform[-1,1] jitter.

## Edge cases (validated)
- **Multiprocessing:** the JAX samplers run in the main process (chain parallelism via `pmap`/`vmap`, no Python fork), so the in-process patch applies — confirmed with `chains=2`. Forking only happens for `sampler="pymc"`, where the CM is inactive.
- **blackjax:** same `sample_jax_nuts` dispatch → covered.
- **Reentrancy/exceptions:** `try/finally` restores the captured attribute even if `fit()` raises.

## Upstream note
This is an interim workaround. `pm.sample` has no supported way to forward `jitter` to the external samplers; a pymc issue proposing to plumb it through (the parameter already exists on `sample_jax_nuts`) will be filed and linked here.

Verified: `ruff check`/`format` clean, `mypy` clean, `pytest tests/test_jitter.py` 4/4 pass.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,6 +17,10 @@ This version contains major breaking updates for HSSM. Please read the release n
 5. `model.summary()` and `model.plot_trace()` methods are now removed. Use `az.summary()` and `az.plot_trace_dist()` instead.
 6. HSSM can now be installed directly from PyPI via `pip` or `uv`. Conda support is no longer provided.
 
+#### Bug fixes:
+
+1. **Restore JAX-NUTS jitter control**: HSSM again disables the built-in initial-value jitter of the `numpyro`/`blackjax` samplers (PyMC 6 removed the public switch that made this possible), so sampling starts from HSSM's own controlled `initval_jitter` instead of an extra uniform jitter (#999).
+
 ### 0.3.1
 
 This version includes the following changes:
--- a/src/hssm/base.py
+++ b/src/hssm/base.py
@@ -43,6 +43,7 @@ from hssm.distribution_utils import (
 from hssm.missing_data_mixin import MissingDataMixin
 from hssm.utils import (
     _compute_log_likelihood,
+    _force_jax_nuts_no_jitter,
     _get_alias_dict,
     _print_prior,
     _split_array,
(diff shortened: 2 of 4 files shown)


digicosmos86 (approved):
LGTM!

PR #1059, opened 2026-07-13
Fix #1056: JAX conversion for LANLogpVJPOp, cotangent-correct CPN gradients, module-level LAN Ops

### LINKED ISSUE #1056: `HSSM.vi(..., backend="jax")` fails for differentiable LAN models because `LANLogpVJPOp` lacks a JAX conversion
## Summary

HSSM's built-in differentiable LAN models cannot run PyMC ADVI through the
PyTensor JAX linker. The forward likelihood Op (`LANLogpOp`) has a
`jax_funcify` registration, but the VJP Op emitted by its `grad()` method
(`LANLogpVJPOp`) does not. Consequently, compiling the ADVI gradient graph
raises `NotImplementedError` before optimization starts.

This prevents GPU-backed JAX VI for LAN-based HSSM models. It also affects
RLSSM models that reuse the same LAN likelihood machinery, but the reproducer
below uses only HSSM's built-in `angle` model.

## Minimal reproducer

```python
import hssm

data = hssm.load_data("cavanagh_theta").head(8)
model = hssm.HSSM(data=data, model="angle", p_outlier=0.0)
model.vi(method="advi", niter=1, draws=1, backend="jax", progressbar=False)
```

### Actual result

```text
NotImplementedError: No JAX conversion for the given `Op`: LANLogpVJPOp
```

### Expected result

The one-iteration ADVI run should compile and return an approximate posterior,
as it does through the C backend.

## Environment

- HSSM 0.4.0 (`main`, local checkout at `e08432f1`)
- PyMC 6.1.0
- PyTensor 3.1.3
- JAX 0.10.2
- Python 3.14.6

## Root cause

`hssm.distribution_utils.jax.make_jax_logp_ops()` defines both `LANLogpOp` and
`LANLogpVJPOp`. It registers only:

```python
@jax_funcify.register(LANLogpOp)
def logp_op_dispatch(op, **kwargs):
    return logp_nojit
```

When PyTensor differentiates the forward Op for ADVI, `LANLogpOp.grad()` emits
`LANLogpVJPOp`. The JAX linker then traverses that Op and cannot convert it.

## Proposed approach

1. Add a `jax_funcify` registration for `LANLogpVJPOp` in
   `src/hssm/distribution_utils/jax.py`.
2. Adapt the wrapped VJP function's positional inputs to the Op contract:
   - data-present case: `logp_vjp(data, *parameters, gz=upstream_gradient)`;
   - no-data vector case: `logp_vjp(*parameters, gz=upstream_gradient)`;
   - preserve the scalar/no-data special case already implemented in `perform`.
3. Add a regression test that calls `model.vi(..., backend="jax")` on a tiny
   built-in `angle` dataset, rather than testing only `eval()` and normal
   PyTensor differentiation.
4. Test both scalar and regression/vector parameter paths, because the VJP Op
   returns one output per differentiable parameter and treats optional extra
   fields specially.

The dispatch should return JAX-native arrays and avoid closing over objects
that PyTensor/JAX cannot serialize. Once this works, JAX-backed ADVI can use a
GPU in the ordinary PyMC manner.

## Workaround

Use `backend="c"` for VI. It executes the LAN Op's Python/PyTensor `perform`
and VJP implementation successfully, but it does not provide GPU acceleration.

AlexanderFengler:
Fixes the HSSM layer of #1056: `HSSM.vi(..., backend="jax")` crashed with `NotImplementedError: No JAX conversion for the given Op: LANLogpVJPOp`.

## Mechanism (three stacked failures)

`backend=` on `pm.fit` is new PyMC 6 surface (pymc-devs/pymc@679ff1235), forwarded blindly through `vi(**vi_kwargs)`. Unlike the NUTS-via-numpyro path (which funcifies only the forward logp graph and lets JAX autodiff through it), VI differentiates **symbolically** (`pytensor.grad` in `pymc/variational/operators.py`), so the `LANLogpVJPOp` emitted by the Op's gradient lands in the JAX-compiled step graph — and it had no `jax_funcify` registration.

Two further failures are **upstream in PyMC** and reproduce with a 5-line plain-PyMC model (no HSSM):
1. `pm.fit(backend="jax")` fails for *every* model — VI approximation params (`mu`/`rho`) are created without static shapes → runtime-shape `Alloc` → JAX tracing `TypeError`. Filed as pymc-devs/pymc#8359.
2. After a JAX-backend fit, shared variables hold raw `jax.Array`s, breaking the subsequently numba-compiled `approx.sample()`. Filed as pymc-devs/pymc#8360.

Upstream issues filed: pymc-devs/pymc#8359, pymc-devs/pymc#8360. **As of 1f604b5f this PR no longer waits for them**: `hssm._vi_compat` ships two scoped, self-disabling workarounds applied inside `HSSM.vi()` only when `backend="jax"` is requested — (1) a context manager that post-processes PyMC's VI parameter creation to add static shapes (no PyMC logic duplicated; no-op once upstream fixes land), and (2) a numpy coercion of approximation storage before `approx.sample()` (unconditionally safe). `vi(backend="jax")` therefore **works end-to-end now**; the previously-xfail reproducer is a passing regression test (advi + fullrank_advi). Known ceiling: models with a single free parameter dimension may still fail to compile (needs the scan-graph part of the upstream fix that cannot be applied from outside PyMC); `backend="c"` remains the fallback there. The shim module is deletable once the pinned PyMC includes both upstream fixes.

## What this PR changes

- **`jax_funcify` registration for `LANLogpVJPOp`** — gradient graphs now compile under the JAX linker; gradients verified identical to the default (`perform`) execution path across all input configurations (data + mixed params, scalars-only CPN, no-data + regression params, single-output, `n_params`/extra-fields).
- **Correctness fix surfaced by review:** the scalars-only params-only gradient path used plain `jax.grad`, silently dropping the upstream cotangent — wrong gradients for e.g. scalars-only choice-only models with `p_outlier > 0`. Now a true VJP; a chain-rule ratio test (`grad((2·logp).sum()) == 2·grad(logp.sum())`) guards it on both backends.
- **`LANLogpOp`/`LANLogpVJPOp` are module-level classes** registered once at import. Previously every `make_jax_logp_ops` call defined fresh classes and registered them on PyTensor's global singledispatch — an unbounded registry growth pinning each model's jitted likelihood (ONNX weights included) for the process lifetime. `perform`/dispatch are now branchless and no longer read mutable `make_node`-set attributes.
- **`grad` → `pullback` migration** (PyTensor 3 API): removes the `FutureWarning` on every symbolic differentiation (pytensor>=3 is guaranteed via bambi>=0.19).
- **Tests:** five graph-level JAX-vs-default equivalence tests + cotangent ratio test + FutureWarning guard + explicit `vi(backend="c")` coverage (the parametrized VI tests only exercised pymc's default compile mode; note their `backend` param is the *likelihood* backend, not the `pm.fit` compile backend).

## Commands run

- `uv run pytest tests/distribution_utils/ tests/test_make_jax_single_trial_logp_from_networks.py` (46 passed)
- `uv run pytest tests/test_likelihoods_choice_only.py tests/test_missing_data_mixin.py tests/test_config.py tests/test_missing_and_deadline.py` (33 passed)
- `uv run pytest tests/slow/test_vi.py -m slow` (19 passed, 1 xfailed)
- `uv run pytest tests/rl/ tests/addm/` (265 passed, 4 skipped)
- Registry-growth check: 0 growth across 5 factory calls (was +2/call)
- NUTS-numpyro smoke + end-to-end `vi(backend="jax")` with the two upstream fixes monkeypatched → returns a posterior `DataTree`
- `uv run prek run` on changed files: clean (repo-wide run has pre-existing failures on main in `docs/tutorials/choice_only_rlssm.ipynb`, `tests/rl/test_rlssm.py`, `tests/addm/oracle/validation.py` — untouched here)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved JAX backend compatibility for LAN likelihoods, including reverse-mode/gradient correctness and successful compilation.
  - Fixed gradients for scalar-only choice-probability likelihoods using correct upstream cotangent handling.
  - Reduced/avoided differentiation warnings and improved repeated model compilation robustness.
  - Ensured variational inference outputs from JAX remain compatible with later standard-backend sampling.

- **Tests**
  - Added coverage for JAX gradient accuracy (including chain-rule with non-unit cotangents), VJP output-shape behavior, and gradient-null handling.
  - Added slow end-to-end compilation tests for both JAX and C backends in variational inference.


docs/changelog.md
```diff
@@ -14,6 +14,9 @@
...
+2. **LAN gradient Op is now JAX-convertible.** The `LANLogpVJPOp` emitted when PyTensor differentiates a LAN likelihood symbolically (e.g. ADVI via `model.vi(..., backend="jax")`) had no `jax_funcify` registration, so compiling the gradient graph with the JAX linker raised `NotImplementedError` (#1056). The VJP Op now has a JAX conversion, with gradients verified identical to the default (`perform`) execution path across all input configurations. Note: end-to-end `vi(backend="jax")` additionally requires two upstream PyMC fixes (VI approximation parameters lack static shapes; JAX-backend fits leave `jax.Array`s in shared-variable storage) — until those land, use `backend="c"` for VI (~1.4x slower per ADVI iteration on CPU).
```
digicosmos86: I don't think I understand this and what follows

pyproject.toml
```diff
@@ -33,6 +33,8 @@ dependencies = [
     "jaxonnxruntime>=0.3.0",
     "numpyro>=0.19",
     "onnx>=1.16.0",
+    # Imported directly; >=3.0.0 required for the Op.pullback gradient API.
```
digicosmos86: Unnecessary. Pytensor 3+ comes with Bambi 0.18+

digicosmos86 (changes requested):
LGTM! Just some comments on AI slop

src/hssm/distribution_utils/jax.py
```diff
@@ -14,6 +14,232 @@
...
+            which is required for choice-probability networks with binary choices.
+        dist_params
+            A list of parameters used in the likelihood computation. The parameters
+            can be a mix of scalars and arrays.
+        """
+        inputs = [pt.as_tensor_variable(dist_param) for dist_param in dist_params]
+        self.has_data = data is not None
+
+        if self.has_data:
+            inputs = [pt.as_tensor_variable(data)] + inputs
```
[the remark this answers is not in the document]
AlexanderFengler: Addressed in 929baa81. `has_data` cannot be derived from the symbolic inputs inside `pullback` (a prepended data matrix is structurally indistinguishable from a matrix-valued parameter), and mixed with/without-data reuse of one instance can never be supported anyway — the wrapped jitted `logp_vjp` has a single calling convention baked in. `make_node` now locks the configuration on first application and raises a descriptive `ValueError` on conflicting reuse, so the latent silent-wrong-gradient scenario is an explicit error. Regression test added (`test_lan_logp_op_locks_data_configuration`).

src/hssm/distribution_utils/jax.py
```diff
@@ -14,6 +14,232 @@
...
+            output = [
+                pytensor.gradient.grad_not_implemented(self, 0, inputs[0]),
+            ] + output
+
+        if self.n_params is not None:
+            start_idx = self.n_params + (1 if self.has_data else 0)
+            output[start_idx:] = [
+                pytensor.gradient.grad_undefined(self, i, inputs[i])
+                for i in range(start_idx, len(inputs))
+            ]
```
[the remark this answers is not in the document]
AlexanderFengler: Same resolution as the comment above (929baa81): the configuration is locked at first `make_node` call, so by the time `pullback` runs, `self.has_data` is guaranteed consistent with every node this instance ever created; conflicting reuse raises immediately at graph-construction time.

docs/changelog.md
AlexanderFengler: Fair — rewrote entries 2–4 in 25bffbfa to lead with the user-visible effect. Short version: (2) `vi(backend="jax")` no longer crashes in HSSM code, but still fails inside PyMC itself for *any* model (two upstream bugs tracked in #1056) — so nothing changes for users until a PyMC release lands; (3) choice-only models with all-scalar parameters + `p_outlier` were fitting with incorrect gradients (chain-rule factor dropped) — now exact; (4) internal: Op classes defined once at module level (fixes unbounded memory growth when building many models in one process) + migration off a deprecated PyTensor API.

pyproject.toml
AlexanderFengler: Removed in 25bffbfa — relying on the bambi>=0.19 → pytensor>=3 chain as you suggest.

digicosmos86 (dismissed):
LGTM. We might want to check if this breaks any tutorials

AlexanderFengler:
Heads-up for reviewers: pushed 1f604b5f, which changes the scope in a user-visible way — `vi(backend="jax")` now works **end-to-end** instead of being blocked on the two upstream PyMC bugs. The workarounds live in a new `hssm._vi_compat` module, are applied only when `backend="jax"` is requested, self-disable once a fixed PyMC is installed (they only add missing static-shape metadata / coerce storage to numpy — no PyMC logic is duplicated), and the whole module is deletable after the upstream fixes ship. The e2e reproducer flipped from xfail to a passing regression test (advi + fullrank_advi). Changelog rewritten accordingly.

AlexanderFengler:
> LGTM. We might want to check if this breaks any tutorials

yep will do that in a follow up as well.

krishnbera (approved):
lgtm!

PR #1060, opened 2026-07-13
Infer VI backend from model_config (#1058)

### LINKED ISSUE #1058: `backend="jax"` in model construction does not gaurantee jax backend for VI
**Describe the bug**
As a general note, if users specify backend="jax" during model construction, that does not guarantee jax backend for VI (atleast). If we run the code below, the VI does not utilize jax. 

```
import hssm

data = hssm.load_data("cavanagh_theta").head(8)
model = hssm.HSSM(data=data, model="angle", p_outlier=0.0, model_config={backend="jax"})
model.vi(method="advi", niter=1, draws=1, progressbar=False)
```

digicosmos86:
- Add `backend` argument to `HSSMBase.vi()`, defaulting to `None`
- When `None`, infer from `model_config.backend`: `jax` → `jax`, `pytensor` → `c`, `None` → `numba`
- Forward the resolved backend to `pm.fit` so a `jax` model config actually uses jax for VI
- Add test asserting `vi()` forwards the resolved backend to `pm.fit`
- Bump version to 0.4.1

Fixes #1058

## Summary by CodeRabbit

- **New Features**
  - Variational inference now allows selecting a computation backend directly.
  - If not provided, the backend is auto-selected from the model configuration (mapping: `jax`→`jax`, `pytensor`→`c`, otherwise→`numba`).

- **Tests**
  - Added a slow test to verify the backend selection is correctly forwarded during variational inference.

- **Chores**
  - Updated the package version to **0.4.1**.

### DIFF
--- a/tests/test_hssm.py
+++ b/tests/test_hssm.py
@@ -571,6 +571,36 @@ def test_deprecated_inference_helpers_raise_documented_error(
     assert str(error.value) == expected_message
 
 
+@pytest.mark.parametrize(
+    "config_backend, backend_arg, expected_backend",
+    [
+        (None, None, "numba"),
+        ("jax", None, "jax"),
+        ("pytensor", None, "c"),
+        ("jax", "numba", "numba"),
+    ],
+)
+def test_vi_passes_backend_to_pm_fit(
+    data_ddm, monkeypatch, config_backend, backend_arg, expected_backend
+):
+    """The resolved backend is forwarded to `pm.fit`."""
+    model = HSSM(data=data_ddm, model="ddm")
+    model.model_config.backend = config_backend
+
+    captured = {}
+
+    def fake_fit(*args, **kwargs):
+        captured["backend"] = kwargs.get("backend")
+
+    monkeypatch.setattr("hssm.base.pm.fit", fake_fit)
+    # Skip post-processing since fake_fit returns no approximation object.
+    monkeypatch.setattr(model, "_clean_posterior_group", lambda dt=None: None)
+
+    model.vi(niter=1, draws=1, backend=backend_arg)
+
+    assert captured["backend"] == expected_backend
+
+
 def test_vi_idata_rejects_attached_approximation(data_ddm):
     """VI approximations must be sampled before they can be exposed as DataTrees."""
     model = HSSM(data=data_ddm)
--- a/tests/test_hssm.py
+++ b/tests/test_hssm.py
@@ -578,6 +578,8 @@ def test_deprecated_inference_helpers_raise_documented_error(
         ("jax", None, "jax"),
         ("pytensor", None, "c"),
         ("jax", "numba", "numba"),
+        ("pytensor", "numba", "numba"),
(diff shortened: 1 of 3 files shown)


tests/test_hssm.py
```diff
@@ -571,6 +571,36 @@ def test_deprecated_inference_helpers_raise_documented_error(
     assert str(error.value) == expected_message
 
 
+@pytest.mark.parametrize(
+    "config_backend, backend_arg, expected_backend",
+    [
+        (None, None, "numba"),
+        ("jax", None, "jax"),
+        ("pytensor", None, "c"),
+        ("jax", "numba", "numba"),
```
krishnbera, line 580: can add a test for these cases ?
`("pytensor", "numba", "numba")`
`("pytensor", "jax", "jax")`

krishnbera (dismissed):
looks good -- except for expanding the test cases?

tests/test_hssm.py
digicosmos86: Added both cases to the parametrized test — they verify that an explicit `backend` argument overrides the config-inferred backend (`("pytensor", "numba", "numba")` and `("pytensor", "jax", "jax")`).

digicosmos86:
@codex review

AlexanderFengler:
Cross-checked this against #1059 (the #1056 fix) — no duplicated work, the two are complementary: this PR is the API surface (explicit `backend` + model-config inference), #1059 is the machinery that makes the jax backend actually work. Three interactions to coordinate though:

1. **Merge order: #1059 should go first.** This PR makes `backend="jax"` the *default* for jax-config models (angle, weibull, all LANs), but on main today `pm.fit(backend="jax")` crashes for those models (#1056) — and the test here mocks `pm.fit`, so CI won't see it. Merged alone, this PR turns an opt-in crash into the default `vi()` behavior for LAN models.
2. **Both PRs edit the same lines of `vi()`** — whoever merges second gets a conflict (trivial, but see 3).
3. **The conflict resolution must move #1059's shim gate onto the resolved backend.** #1059 gates its PyMC compatibility shims on `vi_kwargs.get("backend") == "jax"`; with this PR's explicit `backend` parameter that key never appears in `vi_kwargs`, so a naive resolution would silently skip the shims and re-break jax VI. Post-merge it should read:

```python
if backend == "jax":   # the resolved value, after model-config inference
    with _vi_compat.static_shape_vi_params():
        self._vi_approx = pm.fit(n=niter, method=method, backend=backend, **vi_kwargs)
else:
    self._vi_approx = pm.fit(n=niter, method=method, backend=backend, **vi_kwargs)
```

Happy to rebase this branch on top of #1059 with that resolution once #1059 lands, if useful. One more thought: with the model-config inference in place, the end-to-end jax VI test in #1059 (`tests/slow/test_vi.py::test_vi_jax_compile_backend`) will also cover the inferred-default path for free if it drops the explicit `backend="jax"` argument after both PRs are in.

krishnbera (approved):
lgtm!

PR #1097, opened 2026-07-27
refactor: remove legacy RLDM module

cpaniaguam:
Remove legacy likelihood module for rldm.


krishnbera (approved):
this is just removing the older file that is not longer required. Approving but need to address the linting/type checking.

cpaniaguam:
@krishnbera Linting is flagging some files under docs/design/, not code. I guess some changes in the config make the linter check files there which is probably something we don't want to do.

PR #1108, opened 2026-07-27
Cache angle onnx

cpaniaguam:


digicosmos86 (commented):
I think the test suite share the same issue. Theoretically, we use a huggingface function to download the model, which should cache the file automatically. However, in reality the file is downloaded every time a test is run, especially in slow tests. I think it's worth digging in a bit more

.github/workflows/run_tests.yml
```diff
@@ -36,6 +36,14 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Cache Hugging Face hub
```
digicosmos86, line 39: I think this is a different caching mechanism here than what we want to achieve. This gets us caching across different runs, so that we don't have to download the onnx files again. However, since the onnx models are so small, I am not sure this is necessary. Can we spend some time looking into why the hf_download call is not cached during testing?

cpaniaguam: Sure, we can address this later.

PR #1001, opened 2026-06-17
[pymc6 migration] Fixed model cartoon plots

digicosmos86:
## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved posterior predictive compatibility issues in model cartoon plotting.
  * Fixed quantile probability plotting with various parameter combinations.

* **Breaking Changes**
  * `plot_model_cartoon()` now accepts DataTree objects instead of InferenceData.

* **Documentation**
  * Clarified return type descriptions for `plot_quantile_probability()` function.

PR #1121, opened 2026-08-02
Fix HSGP prior handling: hsgp() terms are usable through HSSM

### LINKED ISSUE #624: HSSM does not properly send priors to Bambi for an HSGP
**Describe the bug**
When setting up a regression for a model parameter with an HSGP term, HSSM is unable to construct a model without raising the following error:

`ValueError: The priors for an HSGP term must be passed within a dictionary. Keys must the names of the parameters of the covariance function and values are instances of bambi.Prior or numeric constants.`

**HSSM version**
0.2.5

**To Reproduce**
Create an HSSM model as normal and provide a regression formula for any parameter like below:
```
 "name": "a",
 "formula": "a ~ 0 + hsgp(stim, m=10, c=2, centered=True)",
 "prior": {
     "hsgp(stim, m=10, c=2, centered=True)": {
          "name": "HSGP",
          "sigma": bmb.Prior("Exponential", lam=2), # amplitude
          "ell": bmb.Prior("InverseGamma", mu=10, sigma=1) # lengthscale
    },
    "sigma": Prior("HalfNormal", sigma=10)
}
```

**Additional context**
It seems like this issue occurs because HSSM automatically creates a bmb.Prior object while the HSGP expects to receive a dictionary.

AlexanderFengler:
Closes #624.

**Severity was higher than reported**: not only did user-specified HSGP priors fail — `hsgp()` terms could not be used through HSSM *at all*, because the safe-prior machinery injected a scalar default that bambi rejects even when no prior was given.

### What changed

- `make_safe_priors` no longer injects scalar defaults onto HSGP terms (detected with bambi's own `is_hsgp_term`); bambi's automatic HSGP priors apply when none are given
- `_make_prior_dict` passes an `hsgp(...)` key's dict through as the covariance-prior dictionary bambi requires, instead of collapsing it into a single `bmb.Prior`; inner `{"name": ...}` specs still convert, and a top-level `"name": "HSGP"` (the shape from #624's report) is accepted and dropped
- All production changes in `src/hssm/param/regression_param.py` (+54 lines); docstring for the prior shapes in `hssm.py`
- New tutorial `docs/tutorials/hsgp_regression.ipynb` (nav-wired, outputs committed, execution skipped in docs build and notebook CI like the other sampling tutorials): recovers a smooth drift effect `v = A*sin(stim)` end to end

### Tests

- 12 new tests: prior shapes (incl. the #624 repro), `by=` variant, two HSGP terms, mixed formulas keeping safe defaults on linear terms, both `prior_settings` modes, non-centering inertness, and a slow end-to-end fit asserting the full HSGP posterior (`sigma`, `ell`, `weights`, GP contribution)

### Verification

- `uv run pytest -m "not slow"` → 549 passed
- `uv run pytest tests/slow/test_hsgp_mcmc.py -m slow` → passed (19s)
- `uv run prek run --all-files` (ruff, ruff-format, pyrefly, mypy) → green per commit
- Cross-repo gate: full fast suite against `bambi @ address-815` (bambinos/bambi#1001) with `UNUSED_PRIORS="error"` → 549 passed — HSSM has no reliance on bambi's silent prior leniency, and correct HSGP prior keys validate under #1001

### Notes

- No dependency or pin changes; `bambi.utils.is_hsgp_term` exists at the current floor (0.19.0)
- Known deferral: a typo'd `hsgp(...)` term name still falls to bambi's silent-drop behavior; bambinos/bambi#1001 (open) turns that into a warning/error

## Summary by CodeRabbit

* **New Features**
  * Added an HSGP regression tutorial demonstrating smooth, stimulus-dependent drift effects, model fitting, and posterior visualization.
  * Added HSGP regression support with configurable covariance-function priors and Bambi-compatible defaults.
  * Added the tutorial to the documentation navigation.

* **Documentation**
  * Expanded regression prior documentation, including HSGP-specific configuration guidance.

* **Bug Fixes**
  * Improved handling of HSGP terms so their structured priors are preserved correctly.


AlexanderFengler:
CI note: the initial lint failure was **not** from this PR's changes. `run_tests (3.13)` / pyrefly flagged `plotting/utils.py:59` — code untouched here. Root cause: `uv.lock` is gitignored (`.gitignore:98`), so CI resolves dependencies fresh on every run; pyrefly 1.2.0 (released since the last green run, with pandas 3.0.5 stubs) types `.loc` column-list selection as `DataFrame | Series` and rejects the declared return. Any PR opened right now would hit this.

Fixed in eddc896a with a one-line `cast`, verified against a lockfile-free fresh-resolve environment (Python 3.13, pyrefly 1.2.0 → 0 errors).

Worth a separate discussion: re-tracking `uv.lock` (the repo docs still call it "the source of truth") or pinning pyrefly, so the type gate doesn't float with upstream releases.

PR #1143, opened 2026-08-06
Scheduled drift detection: weekly gates + monthly notebooks re-run against fresh resolves

AlexanderFengler:
Part of the ecosystem self-healing rollout (spine PR lnccbrown/HSSMSpine#35 is the aggregation layer).

- new `drift.yml`: weekly (lint + fast + slow) and monthly (notebooks) scheduled re-runs of the existing gates via `workflow_call`, against a fresh PyPI resolve of unchanged main — with no committed lockfile this is the detector for upstream/toolchain releases breaking us (ruff 0.16 and pyrefly both did, silently, last month)
- failures create/update ONE deduped `drift`-labeled issue; green runs close it; `force_fail` dispatch input rehearses the path
- `run_slow_tests.yml` / `check_notebooks.yml` gain `workflow_call:`
- fixes riding along: SKIP_NOTEBOOKS bare-filename entries never matched (workshop notebooks ran while "skipped"); notebook run sharded 4-way (single job could exceed the 6 h limit); drift runs clear the `--exitfirst`/`--reruns` addopts; dead `cache-hit` guards removed from the setup-env composite actions

## Summary by CodeRabbit

* **New Features**
  * Added scheduled weekly quality checks and monthly notebook validation.
  * Added automated reporting for failed checks, including issue tracking and resolution updates.
  * Enabled notebook and slow-test workflows to be reused by other automation workflows.
  * Added manual controls for running notebooks and handling check failures.

* **Improvements**
  * Notebook checks now run across four parallel jobs for faster feedback.
  * Improved notebook selection and workshop exclusions for more reliable validation.
  * Installation steps now run consistently across environment setup workflows.

### DIFF
--- a/.github/setup-env-notebooks/action.yml
+++ b/.github/setup-env-notebooks/action.yml
@@ -26,7 +26,8 @@ runs:
         version: "latest"
         enable-cache: true
 
+    # (a stale `if: steps.cache.outputs.cache-hit` guard used to sit here —
+    # no step with id `cache` exists, so it always evaluated true)
     - name: Install hssm
-      if: steps.cache.outputs.cache-hit != 'true'
       run: uv sync --group notebook
       shell: bash
--- a/.github/setup-env-notebooks/action.yml
+++ b/.github/setup-env-notebooks/action.yml
@@ -18,6 +18,8 @@ runs:
         python-version: ${{ inputs.python-version }}
         enable-cache: true
 
+    # (a stale `if: steps.cache.outputs.cache-hit` guard used to sit here —
+    # no step with id `cache` exists, so it always evaluated true)
     - name: Install hssm
       run: uv sync --group notebook
       shell: bash
--- a/.github/setup-env/action.yml
+++ b/.github/setup-env/action.yml
@@ -26,6 +26,8 @@ runs:
         python-version: ${{ inputs.python-version }}
         enable-cache: true
 
+    # (a stale `if: steps.cache.outputs.cache-hit` guard used to sit here —
+    # no step with id `cache` exists, so it always evaluated true)
     - name: Install hssm
       run: uv sync
       shell: bash
--- a/.github/setup-env/action.yml
+++ b/.github/setup-env/action.yml
@@ -26,7 +26,8 @@ runs:
         version: "latest"
         enable-cache: true
 
+    # (a stale `if: steps.cache.outputs.cache-hit` guard used to sit here —
+    # no step with id `cache` exists, so it always evaluated true)
     - name: Install hssm
-      if: steps.cache.outputs.cache-hit != 'true'
       run: uv sync
       shell: bash
--- a/.github/workflows/check_notebooks.yml
+++ b/.github/workflows/check_notebooks.yml
@@ -2,6 +2,7 @@ name: Check notebooks
 
 on:
   workflow_dispatch:
+  workflow_call:  # Called by drift.yml (monthly notebook drift run)
 
 jobs:
   check_notebooks:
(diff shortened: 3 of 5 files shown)


AlexanderFengler:
**Status and merge-order note.**

The three red `run_tests` legs here are **not caused by this PR** — they're the numpy/numba resolution break tracked in #1144, which lands on any fresh resolve of unchanged main. #1137 is the fix; once that merges, a rerun here should go green. (My original attribution in #1144 was wrong, corrected [there](https://github.com/lnccbrown/HSSM/issues/1144#issuecomment-5228620549): it's numba, not pytensor, reached via a pre-release whose metadata omitted numpy's cap.)

**Pushed since opening**, from the review of the sibling drift workflows in ssm-simulators#318 and LANfactory#107 — the same report-job logic is shared across all five repos, so a bug found in one applied here too:

- **A cancelled run no longer counts as green.** `report` runs under `if: always()`, which includes cancellation, and the old logic only treated an exact `failure` as bad — so a cancelled weekly run would have closed an open drift issue with "green again" on no evidence. Closing now requires at least one genuine success and nothing cancelled. `skipped` stays neutral, which matters here specifically: the monthly `notebooks` job is skipped in every weekly run and must not block the close. Verified against all seven realistic result combinations.
- **`issues: write` scoped to the `report` job**, so the token that runs freshly-resolved third-party test code stays read-only.
- **Serialized issue reconciliation** via a job-level concurrency group — the `gh issue list` read and the create/close write aren't atomic.

**This PR must not merge before #1142** without an update: that PR removes the `test_args` input this workflow passes, and gates `run_slow_tests.yml` behind a commit-message opt-in a scheduled run can't satisfy. Details and options are in [my comment there](https://github.com/lnccbrown/HSSM/pull/1142). Plan: let #1137 and #1142 land first, then rebase this and adapt the `fast` and `slow` jobs in one commit.

AlexanderFengler:
**Correcting my earlier comment: this does not need to wait for #1142.** I overstated it there, so to be precise about what the interaction actually is.

**Today this PR is correct and self-consistent.** Rebased onto main at `11195b39`, all three `run_tests` legs are green, and every input it passes exists on main (`run_tests.yml` still declares both `run_slow_tests` and `test_args`). Nothing about merging this is blocked.

**The #1142 interaction is a one-way, time-shifted obligation, not a merge dependency:**

- Merging **this** first does not harm #1142 in any way — that PR merges just as cleanly either way.
- Merging **#1142** later invalidates `drift.yml`, because it removes the `test_args` input the `fast` job passes and adds a commit-message gate to `run_slow_tests.yml` that a `schedule` event can never satisfy.
- So the requirement is simply: **whoever merges #1142 should expect a one-commit follow-up here.** I'll push it as soon as #1142 lands.

I looked at whether I could make this workflow immune to the ordering so no follow-up is needed at all. The clean version would be to drop `test_args` and call `run_tests.yml` with `run_slow_tests: true` instead of calling `run_slow_tests.yml` — both inputs exist before and after #1142. I decided against it for now, for two concrete reasons:

1. **It would lose the batching.** `run_slow_tests.yml` splits the suite into three parallel batches (Missing Data / Core MCMC / Remaining Slow) and the slowest takes ~47 min. Unbatched, against `run_tests.yml`'s 90-minute timeout, that's a real timeout risk — and on main today `run_slow_tests: true` means *all* tests, not just slow ones, so it would be even longer.
2. **`--exitfirst`.** Dropping `test_args` gives up the `-o addopts=` override, so a drift run would stop at the first failure. For PRs that's right; for drift triage the useful signal is the full blast radius — which suites broke and how many tests — since that's what distinguishes "one upstream API changed" from "the whole stack moved".

Both of those objections disappear once #1142 lands, since it removes the batching and the `test_args` mechanism itself. So the follow-up commit is genuinely better done *after* that PR, not worked around now.

**Worth knowing about the failure mode if the follow-up is ever missed:** the weekly run would fail as a workflow configuration error, which files no issue — silent. The freshness budgets are the backstop: `hssm-toolchain-lint` and `hssm-fast-tests` go red after 10 days without a successful job, `hssm-slow-suite` after 17. So it surfaces, just late. That's the layered escalation working as designed, but the follow-up commit is obviously the better path.

PR #1145, opened 2026-08-06
Declare directly-imported dependencies; cap numpy<2.5

### LINKED ISSUE #1144: drift: numpy 2.5 broke the fast suite on fresh resolves (row_stack removed)
**Every fresh dependency resolve of unchanged main now fails** `tests/addm/test_addm_builder.py::test_logp_op_matches_func` on all Python versions with:

```
AttributeError: module 'numpy' has no attribute 'row_stack'
```

**Root cause (verified by diffing installed versions between the last green run and today's red runs):** the only delta is `numpy 2.4.6 → 2.5.1`. numpy 2.5 removed the long-deprecated `row_stack` alias; the freshly-resolved Linux test stack (pytensor 3.2.4 execution path) still calls it. Local environments that include numba stay pinned to numpy ≤ 2.4 (numba's own cap) and therefore don't reproduce — CI's test env has no numba, so it resolves 2.5.1 freely.

**Why this bit us silently:** HSSM imports numpy (38 files), pytensor, pymc, jax, arviz, xarray, and pandas **without declaring them** — their versions arrive transitively with no floor and no cap, and no committed lockfile. This incident surfaced on an unrelated PR (#1143) only because CI resolves fresh per run.

**Healing (mechanical, per the ecosystem's audit-drift playbook):** declare the undeclared runtime dependencies with honest floors, including a temporary `numpy<2.5` cap until the upstream `row_stack` usage is fixed — healing PR to follow, linked here. Lift the cap when pytensor/numpy compatibility is restored upstream.

Classification: root failure (external release), detected pre-merge of the scheduled drift workflow (#1143) — exactly the failure class that workflow exists to catch on Mondays instead of on someone's PR.

AlexanderFengler:
Closes #1144

- Declares the seven packages `src/hssm` imports directly but never declared (`arviz`, `jax`, `numpy`, `pandas`, `pymc`, `pytensor`, `xarray`) — until now their versions arrived transitively via bambi/numpyro with no floor and no cap, which is exactly how the numpy 2.5 breakage landed unbounded.
- Adds a temporary `numpy>=2.0,<2.5` cap: numpy 2.5 removed `np.row_stack`, which the freshly-resolved Linux CI stack (pytensor 3.2.4 path) still calls, failing `tests/addm/test_addm_builder.py` on every fresh resolve of unchanged main (#1144). The cap comment in `pyproject.toml` links the issue; once this merges, the spine's package radar tracks the cap as a standing `ceiling-crossed` event, so lifting it when upstream is fixed cannot be forgotten.
- Floors chosen at the majors the codebase already assumes (pymc≥6 / arviz≥1 from the 0.4.0 migration, jax≥0.7.0 matching the cuda extras, numpy≥2.0 matching ssm-simulators); all sit well below today's resolutions.

Commands run: `uv lock` resolves cleanly (288 packages; numpy → 2.4.6, pymc 6.2.0 / pytensor 3.2.4 unchanged from the last green run — lockfile not committed, HSSM resolves fresh by design).

First mechanical healing PR of the drift-homeostasis rollout (see #1143); rebasing/rerunning #1143 after this merges should turn its `run_tests` legs green.

## Summary by CodeRabbit

* **Chores**
  * Updated runtime dependency requirements to specify supported minimum versions.
  * Added a compatibility constraint for NumPy versions.
  * Documented dependency sources and compatibility guidance.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -25,19 +25,32 @@ classifiers = [
 
 dependencies = [
     "absl-py>=2.3.1",
+    # arviz, jax, numpy, pandas, pymc, pytensor, and xarray are imported
+    # directly throughout src/hssm but were previously only transitive
+    # dependencies (via bambi/numpyro), so upstream releases reached us
+    # unbounded — see https://github.com/lnccbrown/HSSM/issues/1144.
+    "arviz>=1.0",
     "bambi>=0.19.0",
     "h5netcdf>=1.8.1",
     "h5py>=3.16.0",
     "hddm-wfpt>=0.1.6",
     "huggingface-hub>=1.17.0",
+    "jax>=0.7.0",
     "jaxonnxruntime>=0.3.0",
+    # numpy 2.5 removed np.row_stack, which the resolved pytensor stack still
+    # calls; lift the cap when upstream compatibility is restored (#1144).
+    "numpy>=2.0,<2.5",
     "numpyro>=0.19",
     "onnx>=1.16.0",
+    "pandas>=2.0",
+    "pymc>=6.0",
+    "pytensor>=3.0",
     # Imported directly by hssm.plotting (gaussian_kde, chi2); previously only
     # a transitive dependency via pymc.
     "scipy>=1.10",
     # 0.13.1 ships the aDDM engine + fixation_continuation (used by aDDM PPC).
     "ssm-simulators>=0.13.1",
+    "xarray>=2024.10",
 ]
 
 [project.urls]


digicosmos86:
This is mostly a duplicate of https://github.com/lnccbrown/HSSM/pull/1137. Also declaring versions of transitive dependencies transfers the burden to us for managing the versions of these dependencies

AlexanderFengler:
You're right on both counts — closing this in favor of #1137.

**On the duplicate half:** confirmed. The two PRs share exactly one line of intent (capping numpy below 2.5); everything else is disjoint. Since #1137 predates this and is the maintainer's own, it should own the fix. I also re-verified the root cause and **my attribution in #1144 was wrong** — it is numba, as you said, not pytensor. Full traceback and the resolver bisect are in [#1144](https://github.com/lnccbrown/HSSM/issues/1144#issuecomment-5228620549); the short version is that bayesflow 2.0.13 (notebook group, never installed in CI) raised its numpy floor to `>=2.5.1`, and uv 0.12's new pre-release fallback then reached for `numba 0.63.0b1` — the one release in pytensor's `numba<=0.66.0` window whose metadata omits the `numpy<2.5` cap.

**On the transitive-burden half:** that's a fair call and it's yours to make. I'd only note the distinction I was actually reaching for — these six aren't transitive in the "we don't use them" sense; they're imported directly by our own source (numpy in 38 files, pytensor 17, jax 17, pandas 15, pymc 12, xarray 7, arviz 1). But your point about the *cost* stands regardless: every floor we declare is one we then own, and bambi already pins pymc/pytensor for us. Delegating to bambi is a defensible policy, and I'm not going to relitigate it in a PR. If it's ever worth revisiting, the narrow version would be declaring only numpy — the one with no upstream guardian and the heaviest direct use — and leaving the rest to bambi.

Two things I found while verifying that are worth carrying over to #1137, so they don't get lost when this closes. I've left the detail there:

1. #1137 deletes `"scipy>=1.10"` and its comment. I checked whether that was a three-dot-diff artifact and it isn't — the branch is 0 commits behind main, and scipy is imported directly in two places (`gaussian_kde` in `plotting/predictive.py:13`, `chi2` in `plotting/quantile_probability.py:14`). Looks like an unintended side effect of regenerating the dependencies array.
2. There's a fix that closes the actual hole without publishing any numpy cap to downstream users — worth a look before settling on the pin.

PR #1153, opened 2026-08-10
Cartoon: honour keep_boundary in auto height; make two tests able to fail

AlexanderFengler:
Follow-up to #1134. Two reviewer findings were never acted on before merge, and a third was marked "✅ Addressed in commit ce965aa" — that commit only refactored the code into a closure without making the fix. So all three are live in main.

### 1. `keep_boundary=False` silently shrinks the histograms

`_visible_ribbon_max` never checked `keep_boundary`, so with `keep_boundary=False` no ribbon is drawn but `hist_height="auto"` still reserves headroom for one. Measured on a `ddm` fit: **16% of histogram height given away** to something that isn't on screen. One-line gate, plus a test asserting the histograms are strictly taller without the ribbon.

### 2 & 3. Two tests that passed while the thing they tested was broken

Not suspicion — each was demonstrated by mutation, then re-demonstrated as fixed:

- **Drift-band width.** The assertion was `np.ptp(vertices[:, 1]) > 1e-6`, which measures the *cone's* rise, not the band's width. `fill_between` traces the upper curve and returns along the lower one, so a zero-width band still spans ~1.0 in y. Replaced with polygon area (shoelace), which is exactly 0 in the degenerate case. Side-by-side on bands forced degenerate (no per-draw variation at all): **new assertion fails, old assertion passes.**
- **`hist_height="auto"` on the n-choice path.** The assertion was "every artist sits below `ylim_high`", which is near-vacuous there: the drawn top is `bottom + margin * headroom`, so it is under the limit *by construction*. Dropping `HIST_HEIGHT_MARGIN` to 0.0001 — histograms flat and invisible — still passed. Now measures the height actually used and pins it to the margin (halve the margin, halve the height), so an `auto` that ignores headroom or collapses to zero cannot pass.

### Mutation results

| mutation | test |
|---|---|
| revert the `keep_boundary` gate | fails ✓ |
| n-choice `hist_height_ = 1.0` (ignore headroom) | fails ✓ |
| remove all per-draw variation → zero-width bands | new: fails ✓ / old: **passes** ✗ |

My first mutation pass reported all three as "passing under mutation" — because two of them targeted the wrong class and the wrong code path (2-choice `plot_func_model` at line 1476 rather than the n-choice one at 2923). Worth recording: without re-running against the right targets I would have shipped this believing it verified.

### Not fixed, deliberately

CodeRabbit's fourth point — that `_visible_ribbon_max` slices raw decision-time-indexed boundary arrays rather than the arrays `_geometry_arrays` rolls by non-decision time. Real but inert: the error direction is conservative (over-estimate, never overflow), and it is unreachable with every shipped boundary function. All the 2-choice bounds are monotone non-increasing, and even `conflict_gamma` — the one genuinely rising bound — peaks at t≈0.01, so raw and rolled agree exactly. Fixing it is churn against a case nobody can trigger.

### Verification

`97 passed` (full `tests/unit/plotting/`) · `ruff check src/hssm` clean · `ruff format --check .` clean, 242 files. The 56 pydocstyle warnings in the test file are pre-existing (58 on main) and outside what CI lints (`ruff check src/hssm`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic histogram sizing when boundary ribbons are disabled.
  * Prevented unused space from being reserved for hidden boundary elements.
  * Improved visualization of uncertainty bands and drift cones by ensuring their plotted areas are correctly represented.
  * Kept plot limits aligned with automatically calculated heights for clearer, more consistent chart layouts.

* **Tests**
  * Expanded coverage for automatic height scaling, proportional sizing, and plot boundary containment.

### DIFF
--- a/src/hssm/plotting/model_cartoon.py
+++ b/src/hssm/plotting/model_cartoon.py
@@ -1441,6 +1441,10 @@ def plot_func_model(
     def _visible_ribbon_max() -> float:
         # Highest point any (no-noise) boundary reaches inside the visible
         # x-range — expanding bounds rise above their t=0 anchor.
+        if not keep_boundary:
+            # No ribbon is drawn, so reserving headroom for one just shrinks
+            # the histogram for nothing.
+            return 0.0
         n_vis = int(xlim_high / delta_t_model) + 1
         if theta_samples is not None:
             return max(
--- a/tests/unit/plotting/test_model_cartoon.py
+++ b/tests/unit/plotting/test_model_cartoon.py
@@ -894,12 +894,16 @@ class TestPlotFuncModelN:
         assert len(ax.patches) == len(INTERVALS)  # ndt spans
         colors = {ln.get_color() for ln in ax.get_lines() if ln.get_zorder() == 3}
         assert colors == {"black", "green", "blue"}  # per-choice mean curves
-        # Drift cones carry real width. Measure enclosed area, not the y-span
-        # of the vertices: a cone rises, so `ptp` is ~1.0 even when every draw
-        # shares one drift rate and the band is a zero-width line traced out
-        # and back. Area is 0 in exactly that degenerate case, which is what
+        # Every drift cone carries real width. Measure enclosed area, not the
+        # y-span of the vertices: a cone rises, so `ptp` is ~1.0 even when every
+        # draw shares one drift rate and the band is a zero-width line traced
+        # out and back. Area is 0 in exactly that degenerate case, which is what
         # the noise injected into v0/v1/v2 above exists to avoid.
-        assert any(_polygon_area(c.get_paths()[0].vertices) > 1e-6 for c in drift_bands)
+        # `all`, not `any`: measured areas span 2.3e-2 .. 1.1e-1, so the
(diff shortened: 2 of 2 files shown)


AlexanderFengler:
Valid — applied.

`any(...)` would have passed with one healthy band and five degenerate ones, which is a weaker guarantee than the test's own comment claims. Now `all(...)`.

I measured before tightening rather than assuming it was safe, since `all` over six bands could plausibly be flaky if the innermost HDI band were near-degenerate on only six draws (2 chains × 3). It isn't — the areas are:

```
2.30e-02  2.39e-02  4.28e-02  5.29e-02  9.60e-02  1.06e-01
```

The narrowest clears the `1e-6` threshold by about four orders of magnitude, so there's no realistic flakiness risk. Those numbers are now in the comment so the next person doesn't have to re-derive them before touching the threshold. The assertion also carries `areas` as its message, so a future failure reports which band collapsed instead of just `False`.

Re-verified after the change: `97 passed` across `tests/unit/plotting/`, formatting clean, and the mutation guard still holds — forcing every draw identical (no per-draw variation at all) still fails this test, while the original `np.ptp` assertion passes on that same input.

PR #1192, opened 2026-08-15
Point estimation: trustworthy find_MAP, new find_MLE

### LINKED ISSUE #1102: add MAP /MLE functions
Would be good to provide users the opportunity to fit MAP() or MLE to datasets using HSSM -- based on PyMC's findMAP(). This will allow for fast debugging and cheap estimation if they don't want to be fully bayesian (or to compare).

fmuia:
`find_MAP` delegated to `pm.find_MAP` with no starting point, so PyMC used its own (`t=2.0`, `a=2.0`) and ignored HSSM's processed initial values. For many models — including a plain hierarchical DDM on `cavanagh_theta` at float64 — the *gradient* is non-finite there while the logp is finite, so PyMC's start check passed, L-BFGS-B aborted at `nit=0`, and the untouched start was cached as `model.map` with no warning.

- `find_MAP` now starts from `model.initvals`, audits the optimizer result, and never caches a failed run (warns and returns `None`, or raises under `strict=True`). `sample(initvals="map")` raises instead of silently falling back to PyMC's default point.
- New `find_MLE()` maximizes the observed log-likelihood only. It refuses hierarchical models (without the priors the group-level scale is unidentified — a flat ridge, not a point) and `pm.Potential` models, with `allow_unidentified=True` as an escape hatch.
- Both return a `PointEstimate` — a `dict` subclass, so existing `initvals=` code keeps working — carrying `.params`, `.logp`, `.se`, `.to_dataframe()` and `.to_datatree()`. New `n_starts`, `strict`, `se`, `seed`, `method` and `progressbar` parameters; the optimizer that *actually* ran is recorded.
- Multi-start jitter is applied in transformed space, so a start near a bound cannot be pushed out of support. Gradients go through `rewrite_pregrad`, as `pymc.Model.dlogp` does.
- `sample(sampler="laplace")` now warns: bambi's laplace path offers no initval plumbing, so it hits the same bad start.
- New "Find MAP and MLE point estimates" tutorial and `hssm.PointEstimate` reference page.

Verification: 699 fast tests, 6 slow tests, ruff/pyrefly/mypy and `mkdocs build --strict` all pass.

Closes #1102


fmuia:
Superseded by #1218, which carries the same work on a clean linear history rebased onto current `main`.

Rewriting the history here wasn't possible (the branch ruleset blocks force-pushes) so this went to a new branch instead.

PR #1197, opened 2026-08-16
Add the JEAM circular simulator adapter

AlexanderFengler:
## Dependency

HSSM #1196 is merged into `dev` at `8d0a86c0`. This branch starts from that integration commit.

## Summary

- adapt HSSM's `[v_x, v_y, a, t]` simulator protocol to JEAM's fixed circular diffusion model
- pass HSSM's RNG-derived integer seed through unchanged and preserve trial-major replica ordering
- enforce the two-column `[rt, response]` producer contract, including positive RTs and angles on `[-pi, pi)`
- verify exact equality with direct seeded JEAM plus deterministic scalar and trial-wise HSSM random-variable draws
- run the real-producer simulator checks in the dedicated optional-dependency workflow

## Scope boundary

This PR adds only the fixed-CDM simulator handshake. It does not register a public model, change the generic HSSM simulator contract, or claim cross-backend RNG equality. JEAM remains the simulator implementation and mathematical source of truth.

## Verification

- `git range-diff` marks all four pre-restack and post-restack commits patch-identical
- focused direct, adapter, compiled-likelihood, and predictive suite: 30 passed
- affected Ruff lint and formatting checks pass
- `git diff --check origin/dev...HEAD`

The optional CI workflow installs the immutable JEAM prototype revision and reruns the real-producer checks on GitHub.

### DIFF
--- a/.github/workflows/jeam_prototype.yml
+++ b/.github/workflows/jeam_prototype.yml
@@ -7,6 +7,7 @@ on:
       - ".github/workflows/jeam_prototype.yml"
       - "src/hssm/integrations/jeam.py"
       - "tests/integration/test_jeam.py"
+      - "tests/integration/test_jeam_simulator.py"
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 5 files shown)


AlexanderFengler:
Refreshing checks after the patch-identical restack and base change to dev.

PR #1198, opened 2026-08-16
Register the experimental JEAM circular model

AlexanderFengler:
## Dependency

- HSSM #1197 is merged into `dev` at `9897c9d`; this branch starts from that merge commit.

## Summary

- register `circular_diffusion` as a lazy built-in config with `[v_x, v_y, a, t]`, circular responses on `[-pi, pi)`, the pinned JEAM blackbox likelihood, and its seeded simulator
- use the ordinary `hssm.HSSM` lifecycle; no circular subclass is introduced
- preserve categorical registration behavior and reject unsupported non-categorical `p_outlier` mixtures explicitly
- make categorical-only model cartoons fail clearly for non-categorical configs

## Verification

- all six commits are patch-identical to the previously reviewed #1198 stack under `git range-diff`
- Python 3.12 focused config, registration, cartoon, and real JEAM construction slice: 103 passed
- verified the worktree source import and exact JEAM pin `a9f547b3630ae8ff31ccec1b904e0c02fdba6d99`
- affected Ruff lint and format checks pass
- `git diff --check`

## Deferred

- predictive and serialization coverage remains in H08
- optimizer parity, Bayesian recovery, and the repeated recovery benchmark remain split across the H09 stack

### DIFF
--- a/.github/workflows/jeam_prototype.yml
+++ b/.github/workflows/jeam_prototype.yml
@@ -6,7 +6,9 @@ on:
       - "pyproject.toml"
       - ".github/workflows/jeam_prototype.yml"
       - "src/hssm/integrations/jeam.py"
+      - "src/hssm/modelconfig/circular_diffusion_config.py"
       - "tests/integration/test_jeam.py"
+      - "tests/integration/test_jeam_model.py"
       - "tests/integration/test_jeam_simulator.py"
   workflow_dispatch:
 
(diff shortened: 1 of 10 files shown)


AlexanderFengler:
Refreshing CI after the patch-identical restack onto the merged dev tip; no source changes beyond the existing six-commit slice.

PR #1199, opened 2026-08-16
Verify seeded JEAM predictive sampling

AlexanderFengler:
# Dependency

HSSM #1198 is merged into `dev` at `a626c178`. This branch starts directly from that integration commit.

## Scope

- add a keyword-only `random_seed` argument to posterior predictive sampling
- create one NumPy `Generator` per predictive request and advance it across safe-mode chunks
- preserve the established positional `aDDM` continuation arguments while forwarding the seed by keyword
- verify JEAM circular predictive shape, support, precision, seeded repeatability, and save/load behavior
- run the predictive tests in the optional JEAM workflow

## Why one generator

Safe mode splits posterior draws into chunks. Reusing an integer seed for every chunk would restart the stream and duplicate simulated chunks. One request-scoped generator makes the complete request deterministic while each chunk consumes the next part of the stream.

The seed is keyword-only because it is new API. This preserves the positional meaning of `aDDM`'s existing `continuation_mode` and `continuation_params` arguments and keeps the subclass signature compatible with `HSSMBase`.

## Regression coverage

- dependency-free safe-mode regression proves both chunks receive the same generator object
- fixed-parameter JEAM regression proves equal seeds reproduce, unequal seeds differ, and opposite sides of a chunk boundary are not duplicates
- `aDDM` regression calls the existing continuation arguments positionally, passes the seed by keyword, and verifies the effective continuation override and seed forwarding
- signature assertions keep the new seed keyword-only in both the base and subclass APIs
- integration checks cover `(chain, draw, __obs__, rt,response_dim)`, `[rt, response]` order, positive RTs, `[-pi, pi)` support, active precision, and model-state round-tripping

## Verification after restacking

- `git range-diff`: the original six commits are patch-identical to the reviewed stack; `fb668014` and `33e0059c` are focused compatibility follow-ups
- Python 3.12 non-slow suite before the compatibility follow-ups: 703 passed, 371 deselected
- focused predictive, JEAM, persistence, and `aDDM` slice before the compatibility follow-ups: 29 passed
- focused continuation and predictive-RNG regression after the final follow-up: 3 passed, 1 slow deselected
- focused Pyrefly on `base.py` and `addm.py`: 0 errors
- Ruff check and format checks on changed Python files
- `mypy src/hssm`: 85 source files checked before the compatibility follow-ups
- wheel and source distribution builds before the compatibility follow-ups
- `git diff --check`

Hosted checks are rerunning for the final two compatibility commits.

## Deferred

This PR establishes reproducible generative behavior. Objective parity, recovery, and the analytical likelihood remain in later review units.

### DIFF
--- a/.github/workflows/jeam_prototype.yml
+++ b/.github/workflows/jeam_prototype.yml
@@ -9,6 +9,7 @@ on:
       - "src/hssm/modelconfig/circular_diffusion_config.py"
       - "tests/integration/test_jeam.py"
       - "tests/integration/test_jeam_model.py"
+      - "tests/integration/test_jeam_predictive.py"
       - "tests/integration/test_jeam_simulator.py"
   workflow_dispatch:
 
@@ -36,4 +37,5 @@ jobs:
           uv run --group jeam-prototype pytest
           tests/integration/test_jeam.py
           tests/integration/test_jeam_model.py
+          tests/integration/test_jeam_predictive.py
           tests/integration/test_jeam_simulator.py
--- a/src/hssm/addm/addm.py
+++ b/src/hssm/addm/addm.py
@@ -205,6 +205,7 @@ class aDDM(HSSMBase):
         kind: Literal["response", "response_params"] = "response",
         draws: int | float | list[int] | np.ndarray | None = None,
         safe_mode: bool = True,
+        random_seed: int | np.random.Generator | None = None,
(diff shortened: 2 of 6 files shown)


AlexanderFengler:
Temporarily closing and reopening after retargeting this patch-identical restack to dev so the dev-targeted pull-request workflows evaluate the new head.

Issue #1243, opened 2026-08-24
Tutorial notebooks: all 4 drift shards fail, 3 distinct root causes


AlexanderFengler:
Found during the weekly `audit-drift` triage on 2026-08-24. Filing rather than
auto-healing: every one of these needs a code or API decision, none is
mechanical.

The spine's `hssm-notebooks` freshness budget is **red** — there is no
successful `notebooks` job in the last 7 scheduled/dispatched drift runs.

Latest notebook evidence: run
[32095001542](https://github.com/lnccbrown/HSSM/actions/runs/32095001542)
(2026-08-18), where **all four shards failed**. (The 08-24 run is the
lint+fast+slow slot and skips notebooks, so 08-18 is the newest signal.)

Note these failures were never filed as a drift issue — they were appended as
"Still failing" comments to #1220 instead. That queue bug is fixed by #1242.

## Cause 1 — stale PyMC sampler kwarg (shard 0)

`docs/tutorials/sbi_nre_integration.ipynb`

```
TypeError: NUTS.__init__() got an unexpected keyword argument 'nuts_kwargs'
```

Raised from HSSM's own sampling wrapper, not the notebook. A PyMC release
dropped `nuts_kwargs`. This is a dependency **code** migration — deliberately
out of scope for the unattended drift healer. Worth running
`dev-pymc6-migration-check` over the sampling path, since anything else
forwarding `nuts_kwargs` is equally broken and only the notebooks exercise it.

## Cause 2 — RLSSM parameter contract (shard 1)

`docs/tutorials/rlssm_hssm_custom_models.ipynb`

```
ValueError: theta is missing required params: ['v'].
Expected all of: ['rl_alpha', 'rl_alpha_neg', 'v', 'a', 'z', 't', 'theta']
```

The notebook builds a custom RLSSM whose parameter list no longer satisfies the
validator. Either the tutorial or the contract moved; deciding which is the
authority is a modelling call, not a mechanical one.

## Cause 3 — non-picklable native objects under multiprocess sampling (shards 2 and 3)

Two notebooks, one shape of failure — a native handle held by the model cannot
be pickled to the worker processes PyMC spawns per chain:

- `docs/tutorials/bayesflow_nle_onnx_integration.ipynb`
  → `TypeError: cannot pickle 'google._upb._message.EnumDescriptor' object`
- `docs/tutorials/blackbox_contribution_onnx_example.ipynb`
[...]

PR #1258, opened 2026-08-27
fix(log-likelihood): compile JAX models with JAX

AlexanderFengler:
## Summary

- compile post-sampling log-likelihood graphs with the JAX linker when the resolved HSSM model backend is `jax`
- keep the existing compiler behavior for `None` and `pytensor` backends, with focused regression coverage
- replace the BayesFlow mixed-observation continuous-density example with the scientifically appropriate BayesFlow `RatioApproximator` NRE path
- export that NRE to ONNX, load it through HSSM, and validate it against HSSM's analytical DDM posterior
- retain the strict tutorial gates: zero divergences, R-hat <= 1.01, bulk ESS >= 400, and posterior-mean fidelity <= 1.5 pooled posterior SD

## Why

The original tutorial exposed two independent problems.

First, NumPyro sampling succeeded but automatic post-sampling log-likelihood compilation fell back to the Numba linker. That path object-mode lifted `LANLogpOp` and tried to pickle its ONNX closure, failing on protobuf descriptors. JAX-backed HSSM likelihood Ops already provide `jax_funcify` registrations, so post-sampling compilation now uses that supported backend too.

Second, the tutorial treated joint `[reaction time, choice]` observations as a continuous-density BayesFlow NLE target. The revised tutorial uses neural ratio estimation, estimates only drift `v` while fixing nuisance parameters at their simulation values, and compares the result against the analytical DDM on the same data, prior, and inference domain.

Refs #1243.

## Hosted notebook evidence

Targeted workflow run: https://github.com/lnccbrown/HSSM/actions/runs/33123202717

- exact tested source commit: `a61399a8`
- job passed in 50m05s; all 21 cells executed
- retained artifact: `executed-notebook-33123202717-1` (artifact ID `9668483375`)
- executed notebook SHA-256: `613f6db3cb787c1a562735b090a1e8adad9d1acd450e88ec11539e07f5db2377`
- zero execution errors, three inspected plots, and no machine-local paths
- BayesFlow NRE: 0 divergences, max R-hat `1.009`, minimum bulk ESS `454`
- analytical DDM: 0 divergences, max R-hat `1.003`, minimum bulk ESS `700`
- posterior means: NRE `0.332` versus analytical `0.445`; standardized gap `0.83`
- the committed notebook at `7380dfa7` is byte-identical to the retained executed artifact; source cells are unchanged from the tested source commit

The page explicitly presents this as an integration check, not production calibration or simulation-based calibration.

## Additional verification

- focused log-likelihood compiler regression tests: 4 passed
- targeted notebook validator and privacy guard passed in Actions
- `notebook-inspect`: 21 cells, 0 errors, 3 plots, 0 path leaks, 0 unexecuted cells
- Ruff check/format on source changes, Pyrefly on changed Python, Python byte-compilation, and `git diff --check`
- ordinary PR checks at `7380dfa7`: strict docs passed; lint/type passed; package tests passed on Python 3.12, 3.13, and 3.14; PR deployment skipped as designed

No dependency pins are changed. The PR is ready for human review; it is not authorized for agent-side merge.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved post-sampling likelihood computation for JAX-backed models by using the appropriate JAX compilation mode.
  * Preserved standard compilation behavior for non-JAX backends.

* **Documentation**
  * Corrected BayesFlow integration references from NLE to NRE across tutorials and navigation.
  * Updated guidance to recommend the ratio-estimation route for mixed discrete and continuous observations.
  * Clarified related limitations for NSF-based flows.


mkdocs.yml
```diff
@@ -70,7 +70,7 @@ nav:
           - Custom models from JAX callables: tutorials/jax_callable_contribution_onnx_example.ipynb
           - Custom models from ONNX files (blackbox): tutorials/blackbox_contribution_onnx_example.ipynb
           - Integrate an sbi NRE (ONNX): tutorials/sbi_nre_integration.ipynb
-          - Integrate a BayesFlow NLE (ONNX): tutorials/bayesflow_nle_onnx_integration.ipynb
+          - Integrate a BayesFlow NRE (ONNX): tutorials/bayesflow_nle_onnx_integration.ipynb
```
AlexanderFengler: we change the title to `BayesFlow NRE` from `BayesFlow NLE`, but the filename still save `bayesflow_nle_onnx_integration.iypnb`.

Should we rename that too?

src/hssm/base.py
```diff
@@ -942,8 +942,22 @@ def log_likelihood(
             else:
                 dt = self._inference_obj
 
+        # The response logp for JAX-backed models contains custom Ops whose
+        # ``jax_funcify`` registrations are the supported compilation path.
+        # PyMC's default compiler currently selects the Numba linker, which
+        # object-mode-lifts those Ops and cloudpickles their ONNX closures.
+        # Some valid ONNX graphs contain protobuf descriptors that cannot be
+        # pickled, so preserve the model backend for this post-sampling step.
+        compile_mode = "JAX" if self.model_config.backend == "jax" else None
```
AlexanderFengler, line 951: Can you please explain what motivates adding the compile mode here now, and why this wasn't necessary before?

AlexanderFengler (commented):
Need to still address my comments.

src/hssm/base.py
AlexanderFengler: The new mode is for HSSM's **separate post-sampling log-likelihood compilation**, not for NumPyro sampling itself.

`sample()` first finishes `model.fit(...)` and then, by default, calls `self.log_likelihood()` to attach the `log_likelihood` group. Before this PR, that helper always called `pm.compile(mode=None)`. Our [PyMC 6 migration](https://github.com/lnccbrown/HSSM/commit/c6e88c94) moved this path from PyMC 5 / PyTensor 2 to PyMC 6 / PyTensor 3; [PyMC 6 changed the default linker from CVM to Numba](https://github.com/pymc-devs/pymc/releases/tag/v6.0.0).

In the failing notebook run, NumPyro sampling completed successfully. The later log-likelihood step then entered the Numba linker, object-mode-lifted `LANLogpOp.perform`, tried to cloudpickle the BayesFlow ONNX/JAX closure, and failed on its protobuf `EnumDescriptor`. Existing built-in LAN closures happened to serialize through the same fallback, so the backend mismatch remained latent; this BayesFlow graph exposed it. The notebook itself previously completed under PyMC 5.27.1 / PyTensor 2.37.0, before the default-linker change.

`LANLogpOp` already has a supported `jax_funcify` registration. Selecting `"JAX"` from `model_config.backend` makes only JAX-backed likelihoods use that path for post-sampling evaluation; `None` and `pytensor` backends retain the existing default. That is why the switch follows the likelihood backend rather than the sampler, and the three backend cases are covered by the new regression test.

mkdocs.yml
AlexanderFengler: Yes — addressed in `2ff189b0`.

The notebook is now `docs/tutorials/bayesflow_nre_onnx_integration.ipynb`. I also updated the MkDocs nav, the heavy-notebook `execute_ignore` entry, all reader-facing links, and the Docker tutorial list.

Because the old NLE-named route has already been public since #965, I added an anchor-preserving redirect from `/tutorials/bayesflow_nle_onnx_integration/` to `/tutorials/bayesflow_nre_onnx_integration/`. A strict build produced both the new page and the old-route redirect.

This is a 100% file rename: the notebook still has SHA-256 `613f6db3cb787c1a562735b090a1e8adad9d1acd450e88ec11539e07f5db2377`, byte-identical to the validated hosted artifact.

AlexanderFengler:
Review follow-up is complete at `817e5702`:

- both human review threads have evidence-backed replies and are resolved;
- the BayesFlow notebook now uses the NRE filename everywhere, with an anchor-preserving redirect from the published NLE route;
- CodeRabbit produced no actionable inline comments; its sole prior warning was 70% touched-function docstring coverage (7/10), and the exact three undocumented test helpers now have docstrings (10/10);
- strict MkDocs build, the generated redirect, notebook/privacy checks, seven focused compiler tests, and Ruff all passed locally;
- hosted docs, lint/type, and package tests on Python 3.12, 3.13, and 3.14 are all green; PR deployment is skipped as designed.

CodeRabbit's post-push delta review was rate-limited, so it added no fresh comments; its status remains green.

PR #1292, opened 2026-08-31
Fix ensure_positive_ndt for models with trial-to-trial ndt variability

EItanm1999:
ensure_positive_ndt floors the log-likelihood wherever rt - t <= 1e-15. That is the correct support edge for a fixed non-decision time, but not for any model carrying st: there t is drawn per trial from Uniform(t - st, t + st), so the fastest admissible response time is t - st and the band [t - st, t] carries real density. The guard replaces that entire band with LOGP_LB regardless of what the likelihood returned.

The comment above the call site states the assumption ("Assuming that the non-decision time parameter is always named t") but nothing checks whether the model also has st. Affects every *_st model, applied unconditionally at dist.py:607 and :617.

Measured on ddm_st: compiling HSSM's observed-RV logp for a model whose likelihood IS the exact ddm_st quadrature and comparing against the same quadrature called directly gave differences of -5.3 to -12586.8 nats, varying with theta. With the guard skipped the two agree to exactly 0.0000 on 7 of 8 parameter vectors (the 8th had a subject z outside its bound). So this guard accounts for the entire discrepancy while parameters stay in bounds.

p_outlier partially masks it: the floored value is wrapped in the lapse mixture, so affected trials emerge at log(0.05/20) = -5.99 rather than at LOGP_LB, which is why this presented as bad geometry rather than an obvious -inf.

Only st moves the response-time support edge. sz (starting point) and sv (drift) do not, and are deliberately not consulted; a regression test pins that down.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved response-time boundary handling for models with variable non-decision time.
  * Preserved valid likelihood values while correctly replacing invalid values below the model’s support boundary.
  * Ensured other variability parameters do not incorrectly shift this boundary.
  * Kept full-model boundary behavior unchanged.

* **Tests**
  * Added coverage for variable and fixed non-decision-time scenarios.

* **Documentation**
  * Documented the corrected support-boundary behavior and its effect on likelihood calculations.


EItanm1999:
@AlexanderFengler

EItanm1999:
@coderabbitai review

PR #1293, opened 2026-08-31
Clamp default initial values into the model's declared bounds

EItanm1999:
INITVAL_SETTINGS hard-codes natural-scale starting values (t = 0.025, a = 1.5, ...) that predate models whose bounds exclude them: any model with trial-to-trial non-decision-time variability declares t >= 0.25, so the default start sits at -inf log-probability and sampling freezes - 100% divergences, rhat = nan, ESS pinned at the draw count, and no error message pointing at the cause. Found by a fresh-user validation: a stock hssm.HSSM(data, model=<st model>).sample() cannot converge on defaults.

_postprocess_initvals_deterministic now clamps DEFAULT values into the parameter's declared bounds (5% margin) with a warning naming the override. Scope is deliberately narrow: only the hard-coded defaults on the natural-scale (None-link) branch; log_logit defaults are link-space and unbounded; user-supplied initial values were already skipped before this point and are never touched. In-bounds defaults are byte-identical.

## Summary by CodeRabbit

* **Bug Fixes**
  * Default parameter starting values outside declared bounds are now adjusted safely within the allowed range.
  * Adjustments use a 5% margin from the boundary and are logged for visibility.
  * Explicitly provided starting values, link-space defaults, and already-valid values remain unchanged.
  * Bounds from model configuration and included specifications are respected.
  * Default values now follow each parameter’s configured link function.

* **Documentation**
  * Clarified automatic initial-value processing, parameter bounds, and support for additional model types.

* **Tests**
  * Added coverage for bounded, unbounded, valid, link-specific, and intercept parameter scenarios.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -18,6 +18,8 @@
 
 8. **`hssm.load_data` now returns a `pd.DataFrame` unconditionally** (#1146). Its `dataset` argument is required, and the return type is no longer `pd.DataFrame | str`, which forced type-checkers (and users) to narrow away a `str` branch that existed only to print the dataset listing. Use the new **`hssm.list_data()`** to get the names of the built-in datasets as a `tuple[str, ...]` (mirroring `hssm.list_models()`). Breaking: `hssm.load_data()` with no argument now raises `TypeError` instead of returning a listing string.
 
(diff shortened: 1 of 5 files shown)


EItanm1999:
@coderabbitai review

EItanm1999:
@AlexanderFengler

EItanm1999:
@coderabbitai review

PR #1320, opened 2026-09-10
feat: add opt-in MLflow tracking for inference runs

cpaniaguam:
- Introduced `hssm.track()` to record inference workflows as MLflow runs.
- Added tracking documentation in `tracking.md` and `how_to/track_with_mlflow.md`.
- Updated `changelog.md` to reflect new tracking feature.
- Integrated tracking into the HSSM model fitting process.
- Implemented network provenance tracking for models downloaded from HuggingFace.
- Added tests for tracking functionality and network provenance.

## Summary by CodeRabbit

- **New Features**
  - Added opt-in MLflow tracking for inference runs, including model details, sampling metrics, artifacts, and network lineage.
  - Tracking failures do not interrupt inference, and failed runs are marked accordingly.
  - Added an optional installation extra for tracking support.

- **Documentation**
  - Added API reference and guidance for installing, configuring, and using MLflow tracking.
  - Documented lineage queries, recorded metadata, artifacts, and metrics.
  - Reorganized documentation navigation and added release announcement support.


src/hssm/distribution_utils/onnx_utils/model.py
```diff
@@ -68,7 +72,9 @@ def load_onnx_model(
         )
 
     if isinstance(model, str):
-        return onnx.load(hf_hub_download(repo_id=REPO_ID, filename=model))
+        local_path = hf_hub_download(repo_id=REPO_ID, filename=model)
+        tracking.record_network(model, local_path)
```
AlexanderFengler, line 76: this is a no-op if no MlFlow installed?

src/hssm/base.py
```diff
@@ -768,6 +768,10 @@ def sample(
         # values and expose no supported switch; HSSM already applies its own
         # controlled `initval_jitter`, so disable the sampler's jitter for them
         # (issue #999; see also the upstream pymc gap this works around).
+        tracker = tracking.active()
```
AlexanderFengler, line 771: returns None if not `active()`?

docs/how_to/track_with_mlflow.md
```diff
@@ -0,0 +1,83 @@
+# Track fits with MLflow
+
+HSSM can record each fit to an [MLflow](https://mlflow.org) tracking server: the
```
AlexanderFengler, line 3: I'll bring it up in the dev meeting today, just leaving a note:

I think MLflow integration at the HSSM level can make sense, but we probably have to reflect on the actual use-case a little there. 

 1. Linking HSSM runs to the network origin lineage, makes sense for initial proofing of bringing a network into production.
2. One can imagine someone downstream (independent researcher) wanting to track their own numerical experiments, or even their literal overall data analysis workflows. In that case, the linkage to our oscar level file lineage might not be too useful. 

I am concerning with making things useful for the latter case.

AlexanderFengler (commented):
coderabbit still left a ton of comments, I left a few reflections and small code-level questions.

PR #1331, opened 2026-09-14
[bambi rewrite migration] fix(param): Raise ValueError for unknown formula variables

### LINKED ISSUE #1317: [Migration] Reconcile two drifted test assertions (formula error type, noncentered message)
Part of #1305. Parent: #1306.

Two small assertion-level failures. Neither is a bambi API break; grouped because both are one-line judgment calls.

**R7 — exception type drift.** `tests/test_hssm.py::test_transform_params_general[include4-IndexError]` expects `IndexError` for a malformed formula. [bambi#1002](https://github.com/bambinos/bambi/pull/1002) bumps `formulae` (along with `pymc`, `pytensor`, `arviz`, `matplotlib`), and under the new stack the bare name is looked up as a data-frame column, so pandas raises first:

```
KeyError: 'invalid_formula'   (pandas Index.get_loc, via formulae)
```

Worth deciding whether HSSM should normalise this into its own `ValueError` rather than chase whatever type the upstream stack happens to raise.

**R9 — message wording drift.** `tests/test_noncentered.py::test_unknown_dict_key_raises_at_construction` expects `[Uu]nknown component`, but HSSM raises:

```
Unknown parameter name(s) in `noncentered`: ['nonexistent']. Valid parameter names for this model: ['a', 't', 'v', 'z'].
```

Still a `ValueError`; only the wording moved. **Nothing in PR #1002 touches this** — it is a stale assertion against HSSM's own code and is probably independent of #1305. Flagged here so it does not get lost among the migration xfails.

### Tasks
- [ ] R7 — decide intended exception type, then update assertion or code
- [ ] R9 — decide intended message, then update assertion or code
- [ ] Remove the R7/R9 xfail marks

Details: `design/bambi-dev-upgrade-test-failures.md` (R7, R9), `design/bambi-migration-fix-plan.md` (F8).

digicosmos86:
Reconciles the two assertion-level xfails (R7, R9) left over from the bambi 0.20 test triage, and removes both marks.

**Unknown formula variables now raise `ValueError`**

A formula naming a column absent from `data` used to leak whatever bare lookup error formulae/pandas raised (`IndexError` on the old stack, `KeyError` on the new one). `RegressionParam._get_design_matrices` now catches `LookupError` and re-raises a `ValueError` naming the parameter and formula, so the type no longer tracks the upstream stack:

```
Could not build the design matrix for parameter 'v' from formula 'v ~ 1 + not_a_column': 'not_a_column'. Every variable in the formula must be a column of `data`.
```

**Noncentered unknown-key test asserts behaviour, not wording**

The `[Uu]nknown component` string was never HSSM's own — the `noncentered` dict is validated inside `bmb.Model`. The test now checks that a `ValueError` reports the bad key and lists the valid parameter names, independent of bambi's message text.

Closes #1317

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of invalid formula definitions that reference missing data columns.
  - These errors now consistently appear as clear validation errors identifying the affected parameter, formula, and missing variable.
  - Improved error messages for unknown parameter keys by reporting the invalid key alongside the valid parameter names.
  - Invalid configurations that were previously tolerated in some cases are now rejected with more actionable feedback.


digicosmos86:
@coderabbitai review

PR #1345, opened 2026-09-20
Feat/register ddm normalt

EItanm1999:
Register ddm_normal_st as a supported model

Adds the registry entry for the DDM with Normal trial-to-trial
variability in non-decision time (t_trial ~ Normal(t, st); st is the
kernel SD, support unbounded). approx_differentiable only, LAN via
ddm_normal_st.onnx; bounds are the network's training box.

The entry ships ndt_edge_width = 3.0 as its DEFAULT: the Normal kernel
is unbounded, so the admissibility floor belongs at the practical
3-sigma edge t - 3*st (measured: floor at t - st -> rhat 3.39 / ESS 4;
at t - 3*st -> rhat 1.010 / ESS 697, 0% divergences). With this key in
the registry, switching kernels is just the model string - ddm_uniform_st gets
its exact t - st floor, ddm_normal_st gets t - 3*st, and neither requires
the user to remember anything. Stacked on feat/ndt-edge-width-config,
which plumbs the field.

GATED - do not merge before:
1. ssm-simulators ships the ddm_normal_st model config
   (feat/ddm-normal-st-model-config) so the simulator side resolves.
2. The LAN is retrained with exact-likelihood labels and published. The
   current network fabricates ~20 nats of spurious density at the
   st -> 0 training-box corner with a sign-flipped st-gradient there
   (KDE labels cannot resolve structure below their own bandwidth);
   publishing it would bake that defect into the ecosystem.

- Records that the bounds are the network's training box rather than modelling
  choices, and that st means a uniform half-width in ddm_uniform_st but a Normal SD in
  ddm_normal_st, so equal st values are not equal dispersions.
- Lists the model where users look for it, and adds a changelog entry.

---
**Stacked on #1344** (which is itself stacked on #1292; the first two commits are theirs — review the top commit). Model string is `ddm_normal_st`. Its ssm-simulators config is in a separate PR there (branch `feat/ddm-normal-st-model-config`); until it lands, ssm-simulators does not know `ddm_normal_st`, so the simulator side does not resolve. The network `ddm_normal_st.onnx` is not yet on franklab/HSSM; until it is, constructing the model needs `loglik=<local path>`.

Tests: the branch's own test files pass on current `main` (fefed57d) in an environment with ssm-simulators 0.14.0 and bambi 0.21; see the commit for the added cases. Fork CI has not been approved for this fork, so no check-runs appear here.

## Summary by CodeRabbit

* **New Features**
  * Added support for the `ddm_normal_st` model with normally distributed non-decision-time variability.
  * Added the optional `ndt_edge_width` setting to control response-time admissibility boundaries; it defaults to prior behavior and supports model-specific configuration.
  * Invalid edge-width values now produce clear validation errors.

* **Documentation**
  * Updated likelihood documentation and tutorials with `ddm_normal_st` and `ndt_edge_width` guidance.


EItanm1999:
Correct as described: ddm_normal_st.onnx is not yet in franklab/HSSM, and the bare-name lookup is pinned to that repository, so until it is uploaded the model constructs only with an explicit loglik=<path>. The docstring and the PR body both say so. Both networks are public at hf.co/Eitanm/ddm-st-lans for anyone who wants to test now. Uploading to franklab/HSSM is a maintainer action, so the registration and the upload are meant to land together at merge rather than the registration being deferred. The simulator side is the same: ssm-simulators learns ddm_normal_st in lnccbrown/ssm-simulators#361, which this PR depends on for simulate_data.

Issue #359, opened 2024-02-21
import hssm error (with import wfpt)


suwangcn:
ModuleNotFoundError                       Traceback (most recent call last)
Cell In[1], [line 12](vscode-notebook-cell:?execution_count=1&line=12)
      [9](vscode-notebook-cell:?execution_count=1&line=9) import pymc as pm
     [10](vscode-notebook-cell:?execution_count=1&line=10) import jax
---> [12](vscode-notebook-cell:?execution_count=1&line=12) import hssm
     [13](vscode-notebook-cell:?execution_count=1&line=13) hssm.set_floatX("float32")

File [d:\Anaconda_python\envs\hssm_env\lib\site-packages\hssm\__init__.py:14](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/__init__.py:14)
     [11](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/__init__.py:11) import logging
     [12](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/__init__.py:12) import sys
---> [14](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/__init__.py:14) from .config import ModelConfig
     [15](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/__init__.py:15) from .datasets import load_data
     [16](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/__init__.py:16) from .defaults import show_defaults

File [d:\Anaconda_python\envs\hssm_env\lib\site-packages\hssm\config.py:10](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/config.py:10)
      [7](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/config.py:7) from dataclasses import dataclass, field
      [8](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/config.py:8) from typing import TYPE_CHECKING, Any, Literal, cast
---> [10](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/config.py:10) from .defaults import LogLik, LoglikKind, SupportedModels, default_model_config
     [12](file:///D:/Anaconda_python/envs/hssm_env/lib/site-packages/hssm/config.py:12) if TYPE_CHECKING:
[...]

PR #1346, opened 2026-09-20
Feat/register ddm st

EItanm1999:
Register ddm_uniform_st as a supported model

Adds the registry entry for the DDM with uniform trial-to-trial
variability in non-decision time (t_trial ~ Uniform(t - st, t + st); st
is the half-width, SD = st/sqrt(3)). approx_differentiable only, LAN via
ddm_uniform_st.onnx; bounds are the network's training box - in particular
t >= 0.25 and st in [1e-3, 0.25].

Requires the ddm_uniform_st LAN to be published to the franklab/HSSM hub as
ddm_uniform_st.onnx (90 KB; conversion and parity check scripted separately).
The simulator side is the same config that released ssm-simulators
(0.14.0) ships as ddm_st; the explicit name ddm_uniform_st arrives with
our ssm-simulators PR "Add ddm_normal_st and register ddm_uniform_st as
the explicit name of ddm_st" (branch feat/ddm-normal-st-model-config,
stacked on #359, not yet opened). Until that alias ships,
hssm.simulate_data("ddm_uniform_st") needs the ssms name ddm_st.

Correct inference for this model also needs the st-aware admissibility
floor (fix/ensure-positive-ndt-st-aware): without it the stock guard
floors real density in [t - st, t] on ~0.14% of trials and biases t.

- Records that the bounds are the network's training box rather than modelling
  choices, and that st means a uniform half-width in ddm_uniform_st but a Normal SD in
  ddm_normal_st, so equal st values are not equal dispersions.
- Lists the model where users look for it, and adds a changelog entry.

---
**Stacked on #1292** (first commit is #1292's; review the top commit). Model string is `ddm_uniform_st`; the same config exists upstream in ssm-simulators as `ddm_st`, and an ssm-simulators PR registers `ddm_uniform_st` as its explicit alias so `hssm.simulate_data` accepts the new name. The network `ddm_uniform_st.onnx` is not yet on franklab/HSSM; until it is, constructing the model needs `loglik=<local path>`.

Tests: the branch's own test files pass on current `main` (fefed57d) in an environment with ssm-simulators 0.14.0 and bambi 0.21; see the commit for the added cases. Fork CI has not been approved for this fork, so no check-runs appear here.

## Summary by CodeRabbit

- **New Features**
  - Added support for the `ddm_uniform_st` model with uniformly distributed non-decision-time variability.
  - Enabled its default approximate-differentiable likelihood, standard parameter bounds, and predictive sampling support.
  - Added trial-varying non-decision-time constraints using `t - st` where applicable, while preserving existing behavior for other variability parameters.

- **Documentation**
  - Updated likelihood tutorials and changelog documentation to describe `ddm_uniform_st`, its likelihood, and parameter semantics.


EItanm1999:
Both networks are public at hf.co/Eitanm/ddm-st-lans for anyone who wants to test now.

